PR #2, opened 2018-06-03
Add missing module

fabaff:
Looks like that Sphinx is not enough.

```bash
(esphomedocs) [fab@localhost esphomedocs]$ make html
sphinx-build -M html "." "_build"  
Running Sphinx v1.7.5

Extension error:
Could not import extension breathe (exception: No module named 'breathe')
make: *** [Makefile:13: html] Error 2
```

### DIFF
--- a/esphomeyaml/guides/contributing.rst
+++ b/esphomeyaml/guides/contributing.rst
@@ -176,7 +176,7 @@ To check your documentation changes locally, you first need install sphinx (**wi
 
 .. code:: bash
 
-    pip3 install sphinx
+    pip3 install sphinx breathe
 
 Then, use the provided Makefile to build the changes and start a simple web server:
 


OttoWinter:
Ah yes, completely forgot that. Thanks 😺

(I think even with `breathe` installed, it will generate tons of warnings because `doxygen` isn't set up - I'll put a note about this in the docs)

fabaff:
Yes, there are warnings and the process fails.

```bash
$ make html
sphinx-build -M html "." "_build"  
Running Sphinx v1.7.5
making output directory...
[...]
Exception occurred:
  File "/home/fab/Documents/repos/esphomedocs/lib64/python3.6/site-packages/breathe/directives.py", line 696, in get_mtime
    raise MTimerError('Cannot find file: %s' % os.path.realpath(filename))
breathe.directives.MTimerError: Cannot find file: /home/fab/Documents/repos/esphomedocs/_doxyxml/index.xml
```

PR #3, opened 2018-06-03
Details about Sonoff basic

fabaff:
This is a short guide for the Sonoff Basic. For me it looks like that those devices are very similar to the S20 thus the guide is cut short.


OttoWinter:
Amazing! Thanks 😺

PR #6, opened 2018-06-08
Make the example configuration entry copy&paste-friendly

fabaff:
aka make it match the other example configuration entries.


OttoWinter:
Thanks 😀

PR #7, opened 2018-06-08
Passing `inverted` to output should be valid, but does not work, use …

aequitas:
…working invocation in docs until bug is fixed.


OttoWinter:
Good catch !

PR #12, opened 2018-06-09
Add workaround for out of space esp8266 ota update.

aequitas:
I found out this trick to still do ota updates on low flash size devices with a lot of features enabled. Maybe I work this into esphomeyaml one day, but for now I think others might benefit from it as well.

### DIFF
--- a/esphomeyaml/guides/faq.rst
+++ b/esphomeyaml/guides/faq.rst
@@ -197,6 +197,52 @@ Devices that are technically already supported, but for which guides will be cre
 Anything missing? I'd be happy to chat about more integrations over on the `discord channel
 <https://discord.gg/KhAMKrd>`__ - no guarantees that everything will be supported though!
 
+I can't update using OTA because of to little space, now what?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using ESP8266/Sonoff devices and you have many components enabled you will probably encounter this error during OTA update:
+
+.. code::
+
+  ERROR [esphomeyaml.espota] Bad Answer: ERR: ERROR[4]: Not Enough Space
+
+This is because of the limited amount of flash memory available on these devices (often just 1M). The size of the firmware data that is created by esphomeyaml depends on the number of components enabled (eg: webserver, sensors, etc). Especially the webserver component is very large.
+
+During an OTA update the new firmware data needs to be stored on the flash chip so it can be used to replace the old firmware. However it is possible the old firmware is taking up to much space so the new firmware won't fit next to it. This makes a normal OTA update impossible. Forcing you to choose between easy updates or components.
+
+A possible solution is to disable (large) components like webserver so the size of the firmware data stays below a certain size.
+
+If even this doesn't work or you like to have a lot of components enabled there is a workaround that might help you out so you can have your cake and eat it too. Using a two stage OTA update.
+
+First we temporary 'remove' (comment out) all components from the ``yaml`` file, leaving only: ``esphomeyaml``, ``ota`` and ``wifi``, example:
+
+.. code:: yaml
+
+    esphomeyaml:
+      name: sonoff_basic
+      platform: espressif8266
+      board: esp01_1m
+      board_flash_mode: dout
+
+    wifi:
+      ssid: '***'
+      password: '***'
+
+    ota:
+
+    # mqtt:
+    #   broker: 'mqtt'
+    #   username: ''
+    #   password: ''
+    #
+    #
+    # logger:
+    #
+    # switch:
+    # ...
+
+This will result in really small firmware data which has a high chance of fitting the remaining space on your device. After this OTA update has succeeded you are left with a device with no functionality except OTA. Now you can re-enable all components previously commented out and perform a 'normal' OTA update again.
+
 See Also
 ~~~~~~~~
 


esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
 Anything missing? I'd be happy to chat about more integrations over on the `discord channel
 <https://discord.gg/KhAMKrd>`__ - no guarantees that everything will be supported though!
 
+I can't update using OTA because of to little space, now what?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using ESP8266/Sonoff devices you will probably encounter this error during OTA update:
```
OttoWinter: I don't want to scare off users 😅 Maybe add `if you have many features enabled`.

esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
...
+
+`ERROR [esphomeyaml.espota] Bad Answer: ERR: ERROR[4]: Not Enough Space`
```
OttoWinter: Still not markdown 😉

```rst
.. code::

    ERROR [esphomeyaml.espota] Bad Answer: ERR: ERROR[4]: Not Enough Space
```

esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
...
+
+This is because of the limited amount of flash memory available on these devices (often just 1M).
+
+During an OTA update the new firmware data needs to be stored on the flash chip so it can be used to replace the old firmware. However it is possible the old firmware is taking up to much space so the new firmware won't fit next to it. Because the more feature enabled (eg: webserver, sensors, etc) the bigger the firmware data. This makes a normal OTA update impossible. Forcing you to choose between easy updates or features.
+
+There is a workaround however that might help you out so you can have your cake and eat it too. Using a two stage OTA update.
+
+First we temporary 'remove' (comment out) all features from the `yaml` file, leaving only: `esphomeyaml`, `ota` and `wifi`, example:
```
OttoWinter: Literals are written with double back ticks in rst - I know, I hate it too :(

```rst
First we temporary 'remove' (comment out) all features from the ``yaml`` file, leaving only: ``esphomeyaml``, ``ota`` and ``wifi``, for example:
```

esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
...
+
+```yaml
```
OttoWinter: Same:

```rst
.. code:: yaml

    esphomeyaml:
      name: sonoff_basic

    # ...
```

esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
...
+#   password: ''
+#
+#
+# logger:
+#
+# switch:
+# ...
+````
+
+This will result in really small firmware data which has a high chance of fitting the remaining space on your device. After this OTA update has succeeded you are left with a device with no functionality except OTA. Now you can re-enable all features previously commented out and perform a 'normal' OTA update again.
```
OttoWinter: That's actually really smart 😎 - Would never have thought of doing that.

esphomeyaml/guides/faq.rst
```diff
@@ -197,6 +197,48 @@ Devices that are technically already supported, but for which guides will be cre
...
 <https://discord.gg/KhAMKrd>`__ - no guarantees that everything will be supported though!
 
+I can't update using OTA because of to little space, now what?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using ESP8266/Sonoff devices you will probably encounter this error during OTA update:
+
+`ERROR [esphomeyaml.espota] Bad Answer: ERR: ERROR[4]: Not Enough Space`
+
+This is because of the limited amount of flash memory available on these devices (often just 1M).
```
OttoWinter: So it might be good to give indications what usually takes up a whole lot of memory. Especially the web server often uses up a whole lot of space.

Maybe also add a comment here or further down that disabling the web server can help (I mean the web interface isn't that great anyway and I *think* few people are actually using it)

OttoWinter (commented):
Oh noes :( I thought esphomelib would be really efficient with this. Certainly all my ESP8266s only have binary sizes of max. 400kB.

Stil good to add this in the FAQ. Especially your trick is genius 👍

esphomeyaml/guides/faq.rst
aequitas: Sorry, i will probably be making that mistake a lot, time for a rst linter in my editor i guess :)

esphomeyaml/guides/faq.rst
aequitas: Disabling the webserver was my first approach as well, but then I ran into one config that even then was to big, so I found this solution :) I'll add some more extra explaining as well.

aequitas: I've also been thinking on how to make the size of each feature visible, don't know if there currently is some smartness in platformio for this, otherwise something I think might work is a script that compiles a config with one feature enabled at a time and records all the different sizes so we can put that in a table or something.

aequitas:
I pushed some updates, let me know what you think.

OttoWinter (approved):
LGTM 👍 Thanks!

PR #13, opened 2018-06-14
Fix file name

fabaff:


OttoWinter:
Thanks 😺- Switched to a semi-automated build recently. And the build server has, in contrast to my laptop, a case-sensitive file system :)

PR #14, opened 2018-06-20
Note added about resetting the ESP after serial upload before OTA upload

exxamalte:
The OTA update wasn't working for me initially (after a serial upload) and the error message I received (`ERR: ERROR[11]: Invalid bootstrapping`) wasn't much help either. Apparently the ESP module requires a power cycle once after a serial upload before the OTA upload can work. (Source: https://arduino-esp8266.readthedocs.io/en/latest/ota_updates/readme.html, "Note: ESP module should be reset after serial upload. ...")
I added a note to the OTA section on this. Please let me know if you prefer this to be added to the FAQ section instead (or both?).

### DIFF
--- a/esphomeyaml/components/ota.rst
+++ b/esphomeyaml/components/ota.rst
@@ -7,7 +7,7 @@ uploads. esphomeyaml natively supports this through its ``run`` and
 ``upload`` helper scripts.
 
 .. note::
-  Please be aware that the ESP module must be reset after a serial 
+  Please be aware that ESP8266 modules must be reset after a serial 
   upload before OTA can work.
   When you are trying to conduct an OTA update and receive an error message
   ``Bad Answer: ERR: ERROR[11]: Invalid bootstrapping`` the reason is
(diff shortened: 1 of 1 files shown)


esphomeyaml/components/ota.rst
```diff
@@ -6,6 +6,15 @@ firmware binaries to your node without having to use an USB cable for
 uploads. esphomeyaml natively supports this through its ``run`` and
 ``upload`` helper scripts.
 
+.. note::
+  Please be aware that the ESP module must be reset after a serial
```
OttoWinter: Only required on an ESP8266 AFAIK, please update to something like `that ESP8266 modules`

OttoWinter (approved):
Thanks, just a quick comment on ESP8266s

esphomeyaml/components/ota.rst
exxamalte: No worries, I just changed that part.

OttoWinter:
Thanks, and sorry for the long delay ❤️

PR #17, opened 2018-07-22
Correct manual ip example: no indentation

apeeters:


OttoWinter:
Are you sure? For me esphomeyaml throws an *error* if I don't do it with indentation, so it only works with it... (as it should be)

apeeters:
Indeed, no idea what happened.

PR #20, opened 2018-08-28
Update deep_sleep.rst

NearlCrews:
Please see here for reference:
https://esphomelib.com/esphomeyaml/components/mqtt.html

You get a syntax error placing "availability" under mqtt.

### DIFF
--- a/esphomeyaml/components/deep_sleep.rst
+++ b/esphomeyaml/components/deep_sleep.rst
@@ -51,7 +51,24 @@ Configuration variables:
 
         mqtt:
           # ...
-          availability:
+          birth_message:
+            topic: sensor_name/status
+            payload: online
+          will_message:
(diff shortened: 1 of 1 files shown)


esphomeyaml/components/deep_sleep.rst
```diff
@@ -51,7 +51,24 @@ Configuration variables:
 
         mqtt:
           # ...
-          availability:
+          birth_message:
+            topic: sensor_name/status
+            payload: online
+          will_message:
+            topic: sensor_name/status
+            payload: online
```
OttoWinter, line 59: Instead of setting them to `online`, you should be able to set them both to ` ` (empty). Also, the note above should be updated too ("set `availability` to an empty value:").

Anyway, I already fixed this on the next branch for the next release.

PR #22, opened 2018-09-25
1.54in display works

fabaff:
Details:
https://fabian-affolter.ch/blog/wetter-station-mit-e-ink-display-in-5-min/


OttoWinter (approved):
Great, thanks 😺

PR #29, opened 2018-10-01
Add missing (required) name to example Remote Receiver Binary Sensor.

corbanmailloux:
I just noticed that the documented example was missing a required field.


OttoWinter:
Thanks!

PR #31, opened 2018-10-01
Added docker local USB port mapping

jladefoged:
Added docker local USB device port mapping when using Linux


OttoWinter:
Thanks!

PR #98, opened 2018-12-06
Updated docker command to what worked for me.

sherbang:
The docker command in the documenation leaves the docker container with an empty /data/esphomedocs directory.  I assume this is because of the "VOLUME" line in the Dockerfile, but specifying that volume in the command-line fixed it for me.

## Checklist:
  - [✓] The documentation change has been tested and compiles correctly.
  - [✓] Branch: `next` is for changes and new documentation that will go public with the next esphomelib release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [✓] Check this box if you have read, understand, comply, and agree with the [Code of Conduct](https://github.com/OttoWinter/esphomedocs/blob/master/CODE_OF_CONDUCT.md).

### DIFF
--- a/esphomeyaml/guides/contributing.rst
+++ b/esphomeyaml/guides/contributing.rst
@@ -192,7 +192,7 @@ To check your documentation changes locally, you first need install sphinx (**wi
 
     .. code-block:: bash
 
-        docker run --rm -v "$PWD/..":/data -p 8000:8000 -it ottowinter/esphomedocs
+        docker run --rm -v "$PWD/..":/data -v "$PWD":/data/esphomedocs-p 8000:8000 -it ottowinter/esphomedocs
 
     And then go to ``<CONTAINER_IP>:8000`` in your browser.
 


OttoWinter:
If you have the folder structure described in the lines after that this should not be necessary... Under what folder name did you copy the docs repo?

sherbang:
I cloned both `esphomelib` and `esphomedocs` to my `home_automation` folder, then I run docker from within the `home_automation/esphomedocs` folder (this is on Ubuntu 18.10 with docker 18.06.1-ce).

If I use `docker run --rm -v "$PWD/..":/data -p 8000:8000 -it ottowinter/esphomedocs /bin/bash`, I can list the files in `/data/esphomelib` and everything's there, but `/data/esphomedocs` is empty.  Is the VOLUME line in the Dockerfile creating a blank volume on top of the volume mounted on the comand-line with -v?

screenshot 2018-12-07 at 10 58 48 am

sherbang:
BTW: Thanks for these docker images for each of these projects.  It made it so much easier to get started!

OttoWinter:
Hmm, you might be right about that with the volume! I guess I'll just remove the `VOLUME` line. It's not strictly necessary after all.

OttoWinter:
Should work now

PR #104, opened 2018-12-12
Update remote_transmitter.rst

jjcrimmins:
Carrier Duty Percentage should be 47% as this will set the Carrier Frequency to 38kHz.  With 50% it will set to 40kHz, which still works but may cause issues with some IR receivers.

## Description:

**Pull request in [esphomeyaml](https://github.com/OttoWinter/esphomeyaml) with YAML changes (if applicable):** OttoWinter/esphomeyaml#<esphomeyaml PR number goes here>
**Pull request in [esphomelib](https://github.com/OttoWinter/esphomelib) with C++ framework changes (if applicable):** OttoWinter/esphomelib#<esphomelib PR number goes here>

## Checklist:
  - [x] The documentation change has been tested and compiles correctly.
  - [x] Check this box if you have read, understand, comply, and agree with the [Code of Conduct](https://github.com/OttoWinter/esphomedocs/blob/master/CODE_OF_CONDUCT.md).

### DIFF
--- a/esphomeyaml/components/remote_transmitter.rst
+++ b/esphomeyaml/components/remote_transmitter.rst
@@ -23,7 +23,7 @@ Use-cases are for example infrared remotes or 433MHz signals.
     # Example configuration entry
     remote_transmitter:
       pin: GPIO32
-      carrier_duty_percent: 50%
+      carrier_duty_percent: 47%
 
     # Individual switches
     switch:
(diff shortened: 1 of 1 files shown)


OttoWinter:
How did you determine this and with which protocol?

esphomelib dynamically changes the carrier frequency to match the IR protocol. For example for sony remotes it choses 40kHz and for most other ones 38kHz. Anyway, this will probably not cause any issues with remotes anyway since the oscillators in the remotes I've seen have terrible consistency - so the receivers probably accept a wide range of frequencies too.

This is definitely not a "fix" - if this is a problem the fix would be to write a note somewhere on that page telling users that they might want to try another duty percentage.

OttoWinter:
Closing due to inactivity.

PR #124, opened 2019-01-16
Added explanation how to change the password

wolph:
Added an explanation as to how to change/add the password


OttoWinter (commented):
Unfortunately, this won't work for dashboard users (which make up the majority now) because there's no `upload` function there.

I have told people in the past though to use this trick:
```yaml
esphomeyaml:
  on_boot:
    - lambda: |-
        id(ota).set_auth_password("New password");
ota:
  password: "Old password"
  id: ota
```

Could you change it to that?

wolph:
Done :)

OttoWinter (approved):
Thanks!

PR #138, opened 2019-01-27
Extra example in cookbook / flashing DOIT ESP32

DavidDeSloovere:
## Description:
Added extra example
Fixed deadlink
Added info for flashing DOIT ESP32
Improve contributing section

## Checklist:


DavidDeSloovere:
@OttoWinter  No CI build on PRs yet?
I'll try to build myself. Have very little experience with Python and it's doc tools.

esphomeyaml/cookbook/garage-door.rst
```diff
@@ -41,6 +41,26 @@ for a short period of time, the close/open action begins.
...
+        pin: D6
+        id: relay
+      - platform: template
+        name: "Gate Remote"
+        icon: "mdi:gate"
+        optimistic: no
+        turn_on_action:
+          - switch.turn_on: relay
+          - delay: 500ms
+          - switch.turn_off: relay
```
OttoWinter: There's already an example for exactly this here: https://esphomelib.com/esphomeyaml/components/switch/gpio.html

esphomeyaml/guides/contributing.rst
```diff
@@ -178,39 +181,49 @@ RST primer:
...
-    Alternatively, you can use the `esphomedocs docker image <https://hub.docker.com/r/ottowinter/esphomedocs/>`__:
+    The easiest way is to use the `esphomedocs docker image <https://hub.docker.com/r/ottowinter/esphomedocs/>`__:
 
     .. code-block:: bash
 
         docker run --rm -v "$PWD/..":/data -p 8000:8000 -it ottowinter/esphomedocs
 
+    .. code-block:: powershell
+
+        docker run --rm -v "${PWD}/..":/data -p 8000:8000 -it ottowinter/esphomedocs
```
OttoWinter: Let's just modify the line above to use `${PWD}` - I think for bash both will work.

esphomeyaml/guides/faq.rst
```diff
@@ -81,9 +81,10 @@ Select the firmware binary and finally press "Flash ESP".
 
 .. note::
 
-    If you're just seeing ``Connecting....____....`` on the screen and the flashing fails, that's
-    usually a sign that the ESP is defect or cannot be programmed. Please double check the UART wires
-    are connected correctly if flashing using a USB to UART bridge.
+    If you're just seeing ``Connecting....____....`` on the screen and the flashing fails, that might
+    be a sign that the ESP is defect or cannot be programmed. Please double check the UART wires
+    are connected correctly if flashing using a USB to UART bridge. For some devices you need to
+    keep pressing the BOOT button until flashing has begun (ie. Geekcreit DOIT ESP32 DEVKIT V1).
```
OttoWinter, line 87: Is this really the case? I understand why you would need to press it down while plugging power *in*, but do you really need to keep it pressed down?

esphomeyaml/cookbook/garage-door.rst
DavidDeSloovere: I know,that's were I got it :-)
but it wasn't easy to find, and here I add some more info about it

esphomeyaml/guides/contributing.rst
DavidDeSloovere: does it work without quotes in bash? just tried it in powershell and it can't have quotes

esphomeyaml/guides/faq.rst
DavidDeSloovere: yes, worked for me. flashed 2 devices this way just yesterday. found this solution after some googling.

esphomeyaml/cookbook/garage-door.rst
OttoWinter: I do not like duplication in code, and especially so because the garage door example here has nothing to do with this additional example.

Also, I think here it is even harder to find than in the `switch.gpio` docs - you can add more infos in the example there if you want.

DavidDeSloovere: Ok, I'll have a look at moving this

DavidDeSloovere:
Better like this?

OttoWinter (approved):
Yes!

PR #143, opened 2019-01-29
Updating Repo URLs in Contribution Guide

badbadc0ffee:
## Checklist:


OttoWinter (approved):
Thanks!

PR #144, opened 2019-01-30
Add support for DS1307 real time clock

badbadc0ffee:
## Description:

**Pull request in [ESPHome](https://github.com/esphome/ESPHome) with YAML changes (if applicable):** esphome/ESPHome#399
**Pull request in [ESPHome-Core](https://github.com/esphome/ESPHome-Core) with C++ framework changes (if applicable):** esphome/ESPHome-Core#457

## Checklist:
  - [ ] Branch: `next` is for changes and new documentation that will go public with the next esphomelib release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/esphomeyaml/components/time.rst
+++ b/esphomeyaml/components/time.rst
@@ -59,6 +59,32 @@ Configuration variables:
 - **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run at specific intervals using
   a cron-like syntax. See :ref:`time-on_time`.
 
+DS1307 Time Source
+------------------
+
+With the ``DS1307`` time platform, the battery powered clock chip connected via
+:doc:`I2C </esphomeyaml/components/i2c>` bus is used to synchronize the system
+time at start-up and upon request.
+
+.. code-block:: yaml
+
+    # Example configuration entry
+    time:
+      - platform: sntp
+      - platform: ds1307
+        id: ds1307_time
+
+    interval:
+      - interval: 15min
+        then:
+          - lambda: id(ds1307_time).write_data();
+
+Configuration variables:
+
+- **id** (*Optional*, :ref:`config-id`): Specify the ID of the time for use in lambdas.
+- **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run
+  at specific intervals using a cron-like syntax. See :ref:`time-on_time`.
+
 Use In Lambdas
 --------------
 
--- a/esphomeyaml/components/time.rst
+++ b/esphomeyaml/components/time.rst
@@ -62,28 +62,31 @@ Configuration variables:
 DS1307 Time Source
 ------------------
 
-With the ``DS1307`` time platform, the battery powered clock chip connected via
-:doc:`I2C </esphomeyaml/components/i2c>` bus is used to synchronize the system
-time at start-up and upon request.
+With the ``ds1307`` time platform, the battery powered :doc:`DS1307 <https://datasheets.maximintegrated.com/en/ds/DS1307.pdf>`
+clock chip connected via :doc:`I2C </esphomeyaml/components/i2c>` bus is used to synchronize the system time.
 
 .. code-block:: yaml
 
     # Example configuration entry
     time:
-      - platform: sntp
       - platform: ds1307
-        id: ds1307_time
-
-    interval:
-      - interval: 15min
-        then:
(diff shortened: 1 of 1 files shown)


esphomeyaml/components/time.rst
```diff
@@ -59,6 +59,32 @@ Configuration variables:
 - **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run at specific intervals using
   a cron-like syntax. See :ref:`time-on_time`.
 
+DS1307 Time Source
+------------------
+
+With the ``DS1307`` time platform, the battery powered clock chip connected via
```
OttoWinter: Please add a reference to a datasheet and link to a seller (like adafruit, sparkfun; not ebay or so because those links expire)

esphomeyaml/components/time.rst
```diff
@@ -59,6 +59,32 @@ Configuration variables:
...
+:doc:`I2C </esphomeyaml/components/i2c>` bus is used to synchronize the system
+time at start-up and upon request.
+
+.. code-block:: yaml
+
+    # Example configuration entry
+    time:
+      - platform: sntp
```
OttoWinter: This is going to need more documentation. Users will be confused why there is both sntp and ds1307 here.

esphomeyaml/components/time.rst
```diff
@@ -59,6 +59,32 @@ Configuration variables:
...
+      - platform: ds1307
+        id: ds1307_time
+
+    interval:
+      - interval: 15min
+        then:
+          - lambda: id(ds1307_time).write_data();
```
OttoWinter: The user ideally shouldn't need to write an automation for a basic task like synchronizing time.

The component should call `set_interval` in setup() and register a function to write data to the sensor every 15 minutes or so (could even be a config option, where `never` would mean the time is never written, see `cv.update_interval`). In there, the call to `now()` should also be checked with `is_valid` and only then written.

esphomeyaml/components/time.rst
badbadc0ffee: Sure! I'll link to a datasheet. But do you really want a link to a specific seller? Basically you can get those chips everywhere - just look for DS1307 on Amazon, Aliexpress, ebay, whichever is your preferred dealer.

OttoWinter: Well, I don't care which seller it is specifically. It's just that I want links that don't go dead after a while (a mess to maintain)

badbadc0ffee:
the sync intervals have not yet been implemented...

esphomeyaml/components/time.rst
badbadc0ffee: Sure - and that is why I would avoid putting in a link such as https://www.amazon.de/dp/B00Y7ZBZW6: that one will definitely expire sooner or later. But you will always get the chip if you search for the chip name...

esphomeyaml/components/time.rst
```diff
@@ -59,6 +59,35 @@ Configuration variables:
...
+        get_interval: 10min
+      # optionally add a time platform to sync the hardware clock to
+      - platform: sntp
+
+Configuration variables:
+
+- **id** (*Optional*, :ref:`config-id`): Specify the ID of the time for use in lambdas.
+- **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run
+  at specific intervals using a cron-like syntax. See :ref:`time-on_time`.
+- **set_interval** (*Optional*): Specify the interval at which the system time
```
OttoWinter, line 83:
```suggestion
- **set_interval** (*Optional*, :ref:`config-time`): Specify the interval at which the system time
```

esphomeyaml/components/time.rst
```diff
@@ -59,6 +59,35 @@ Configuration variables:
...
+  is written to the hardware clock. By default, the hardware clock is only
+  written to once if the clock chip was not synchronized before and if the system
+  time is valid (i.e. synchronized from another time source).
```
OttoWinter, line 86: Potential problem with this "once" behavior will be that the "get" can interfere with it. For example on boot the DS1307 will be the first to get the current time - so then you would only be writing the time you got from the sensor back to it.

I would make it an interval with the magic value `never`. So for first setup users sets something like `1m` and after that can disable it.

OttoWinter:
Closing due to inactivity :(

PR #161, opened 2019-02-11
Enable use of alternate hardware UARTs for logging 

brandond:
## Description:
Add documentation for use of alternate hardware UARTs for logging and HardwareSerial interop with Serial UART component.

**Pull request in [esphomeyaml](https://github.com/OttoWinter/esphomeyaml) with YAML changes (if applicable):** OttoWinter/esphomeyaml#427
**Pull request in [esphomelib](https://github.com/OttoWinter/esphomelib) with C++ framework changes (if applicable):** OttoWinter/esphomelib#483

## Checklist:
  - [X] The documentation change has been tested and compiles correctly.
  - [X] Branch: `next` is for changes and new documentation that will go public with the next esphomelib release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/components/logger.rst
+++ b/components/logger.rst
@@ -21,15 +21,36 @@ Configuration variables:
 
 -  **baud_rate** (*Optional*, int): The baud rate to use for the serial
    UART port. Defaults to ``115200``. Set to ``0`` to disable logging via UART.
--  **tx_buffer_size** (*Optional*, string): The size of the buffer used
+-  **tx_buffer_size** (*Optional*, int): The size of the buffer used
    for log messages. Decrease this if you’re having memory problems.
-   Defaults to 512.
+   Defaults to ``512``.
+-  **hw_uart** (*Optional*, string): The Hardware UART to use for logging.
+   Defaults to ``UART0``.
 -  **level** (*Optional*, string): The global log level. Any log message
-   with a lower severity will not be shown. Defaults to DEBUG.
+   with a lower severity will not be shown. Defaults to ``DEBUG``.
 -  **logs** (*Optional*, mapping): Manually set the log level for a
    specific component or tag. See :ref:`Manual Log Levels for more
    information <logger-manual_tag_specific_levels>`.
 -  **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
+
+.. _logger-hw_uarts:
+
+Hardware UARTs
+--------------
+
+The logger component makes use of platform-specific hardware UARTs.
+By default, the logger will occupy ``UART0``. The ESP32 has three hardware UARTs, all of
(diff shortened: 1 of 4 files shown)


components/uart.rst
```diff
@@ -27,10 +27,20 @@ In some cases only **TX** or **RX** exists as the device at the other end only a
...
+    ``UART0`` is (by default) used by the :doc:`logger component </components/logger>`, connected to ``GPIO1`` and
+    ``GPIO3``. If you configure a UART that overlaps with these pins, you can share the hardware with the logger
+    and leave others available. If you have configured the logger to use a different hardware UART, the pins used
+    for hardware sharing change accordingly.
+
+    The ESP32 has three UARTs. Any pair of GPIO pins can be used, as long as they support the proper output/input modes.
+
+    The ESP8266 has two UARTs; the second of which is TX-only. Only a limited set of pins can be used. ``UART0`` may
+    use either ``GPIO1`` and ``GPIO3``, or ``GPIO15`` and ``GPIO13``. ``UART1`` must use ``GPIO2``. Any other
+    combination of pins will result in use of a software UART.
```
OttoWinter: This is very long for a note now (and it's above the example code, so it might push the example code off screen on some devices).

Could this be both:
1. moved below the configuration variables section
2. Transformed into a section - notes are harder to read because of less space (plus seo doesn't seem to like it, google has been complaining about these in search console for some time already)

brandond:
Does this look better?

components/logger.rst
```diff
@@ -21,16 +21,43 @@ Configuration variables:
...
+- ``UART0_SWAP`` - TX: GPIO15, RX: GPIO13  (Only on ESP8266)
+- ``UART1`` - TX: GPIO2, RX: None  (Only on ESP8266)
+- ``UART1`` - TX: GPIO9, RX: GPIO10  (Only on ESP832)
+- ``UART2`` - TX: GPIO16, RX: GPIO17  (Only on ESP832)
+
+.. note::
+
+   The ESP8266 Arduino framework has some logging that is hardcoded to use ``UART0`` by
+   default. If you want to send all messages via ``UART1`` on ESP8266, you will also need
+   to add ``build_flags: -DDEBUG_ESP_PORT=Serial1`` to your :ref:`esphomeyaml-platformio_options`.
```
OttoWinter: We can do better than this. In esphome[yaml] we have full control of the build flags (see https://github.com/esphome/esphome/pull/427/files#diff-ff8b4d8fa7ad3e06cc7e1dbdbc931debR101). We should set that build flag automatically.

(also because adding `build_flags:` to your platformio option will _replace_ all auto-generated ones, which will probably cause build issues)

brandond: Oh, perfect. Will do.

OttoWinter (approved):
🐬

PR #167, opened 2019-02-22
Update output documentation for min_power configuration

mtl010957:
## Description:
Documentation changes to support adding a min_power configuration to the output component, complementary to the max_power configuration already present. This setting will let the user clamp the output of any floating point variable component like [ESP8266 PWM](https://esphome.io/components/output/esp8266_pwm) to make the full range between min and max available to the frontend in HA. Servos especially will benefit from this, since many standard servos operate in a range between 10%-20% duty cycle.

**Related issue (if applicable):** fixes OttoWinter/esphome/feature-requests#64

**Pull request in [esphomeyaml](https://github.com/OttoWinter/esphomeyaml) with YAML changes (if applicable):** OttoWinter/esphomeyaml#448
**Pull request in [esphomelib](https://github.com/OttoWinter/esphomelib) with C++ framework changes (if applicable):** OttoWinter/esphomelib#516

## Checklist:
  - [X] The documentation change has been tested and compiles correctly.
  - [X] Branch: `next` is for changes and new documentation that will go public with the next esphomelib release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/components/output/index.rst
+++ b/components/output/index.rst
@@ -27,6 +27,7 @@ Each output platform extends this configuration schema.
         id: my_output_id
         power_supply: power_supply_id
         inverted: False
+        min_power: 0.01
         max_power: 0.75
 
 Configuration variables:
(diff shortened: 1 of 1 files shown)


components/output/index.rst
```diff
@@ -38,9 +39,14 @@ Configuration variables:
   automatically be switched on too.
 - **inverted** (*Optional*, boolean): If the output should be treated
   as inverted. Defaults to ``False``.
+- **min_power** (*Optional*, float): Only for float outputs. Sets the
+  minimum output value of this output platform. Each value will be
```
OttoWinter: I think we can drop the comment about how the value will be calculated. I think it's pretty clear from the names what it does and the comment just makes things harder to understand.

mtl010957: Sure, I agree that's a bit redundant, I've cleaned it up. Thanks.

OttoWinter (approved):

PR #172, opened 2019-02-25
Added the proper way to make an HMI file for ESPhome

AalianKhan:
I had great difficulties using this component and added how to properly make a display file so it can communicate to the microcontroller properly. Love your work 👍

## Description:

## Checklist:


AalianKhan:
I have the images but can't upload them because of permissions

OttoWinter (approved):
Good 👍 

As for the bkcmd, I will change it so that is set directly on boot by esphome itself - shouldn't interfere with anything else (https://github.com/esphome/esphome-core/pull/523/files#diff-854204f562df2daa82ef0efcf94d12d2R18)

AalianKhan:
Can you also change the baud rate to 115200 so the communication is faster

AalianKhan:
And no one will have to play with it to make work

components/display/nextion.rst
```diff
@@ -47,14 +47,33 @@ Configuration variables:
...
 
+Creating a HMI file
+-------------------
+
+First, you need to use the `Nextion Editor <https://nextion.itead.cc/resources/download/nextion-editor/>`__ to
+create a display file. When you open up the editor create new project and set your device type (Device model is listed at the back of the display). 
+
+Now in ``Preinitialize Event`` You have to type ``bkcmd=3``
+
+.. figure:: images/nextion-HMI1.jpg
```
OttoWinter: This image does not exist?

OttoWinter (commented):
No baud_rate won't be changed to something higher. The reason is that the UART for ESP8266 is a software based one, which can't handle very high baud rates like 115kbaud.

components/display/nextion.rst
AalianKhan: it says you don't have permissions to pulish images. since we are making ``bkcmd=3`` publish at start we won't need that.

AalianKhan: here is the image that I want to display
nextion_editor_2019-02-26_16-36-36

PR #178, opened 2019-02-26
change "inverted" to "filters: invert"

Snipercaine:
## Description:

## Checklist:


OttoWinter (commented):
Both should work... And actually I'd rather keep it the old way with pin inversion (slightly more efficient memory-wise, and a bit cleaner from the generated code side)

PR #185, opened 2019-03-01
Document SDS011 component

puuu:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#467
**Pull request in [esphome-core](https://github.com/esphome/esphome-core) with C++ framework changes (if applicable):** esphome/esphome-core#538

## Checklist:


components/sensor/sds011.rst
```diff
@@ -0,0 +1,85 @@
...
+Configuration variables:
+------------------------
+
+- **pm_2_5** (*Optional*): Use the concentration of particulates of size less than 2.5µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **pm_10_0** (*Optional*): Use the concentration of particulates of size less than 10.0µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **query_mode** (*Optional*): Whether the sensor should be actively polled or actively report. Defaults to ``false``.
```
OttoWinter:
```suggestion
- **query_mode** (*Optional*, boolean): Whether the sensor should be actively polled or actively report. Defaults to ``false``.
```

components/sensor/sds011.rst
```diff
@@ -0,0 +1,85 @@
...
+
+- **rx_only** (*Optional*): Whether the sensor is connected only by RX pin to the UART bus. This connection is not
+  possible in query mode. Since the SDS011 can not be configured in RX-only mode, `update_interval` has no
```
OttoWinter:
```suggestion
  possible in query mode. Since the SDS011 can not be configured in RX-only mode, ``update_interval`` has no
```

components/sensor/sds011.rst
```diff
@@ -0,0 +1,85 @@
...
+
+- **pm_2_5** (*Optional*): Use the concentration of particulates of size less than 2.5µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **pm_10_0** (*Optional*): Use the concentration of particulates of size less than 10.0µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **query_mode** (*Optional*): Whether the sensor should be actively polled or actively report. Defaults to ``false``.
+
+- **rx_only** (*Optional*): Whether the sensor is connected only by RX pin to the UART bus. This connection is not
```
OttoWinter:
```suggestion
- **rx_only** (*Optional*, boolean): Whether the sensor is connected only by RX pin to the UART bus. This connection is not
```

index.rst
```diff
@@ -132,6 +132,7 @@ Sensor Components
     Xiaomi MiFlora, components/sensor/xiaomi_miflora, xiaomi_miflora.jpg
     Xiaomi MiJia, components/sensor/xiaomi_mijia, xiaomi_mijia.jpg
     Custom Sensor, components/sensor/custom, language-cpp.svg
+    SDS011 Sensor, components/sensor/sds011, sds011.jpg
```
OttoWinter: 🔤 This list is ordered alphabetically (except for custom)

components/sensor/sds011.rst
```diff
@@ -0,0 +1,86 @@
...
+- **pm_2_5** (*Optional*): Use the concentration of particulates of size less than 2.5µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **pm_10_0** (*Optional*): Use the concentration of particulates of size less than 10.0µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **update_interval** (*Optional*, :ref:`config-time`): The interval to check the sensor. This affects the working
+  period of the SDS011 sensor as well as the query interval if query mode is enabled. Defaults to ``1s``.
+
+- **query_mode** (*Optional*, boolean): Whether the sensor should be actively polled or actively report.
```
OttoWinter: Reading through the code again, what benefit does ``query_mode`` have over non-query mode? If it's not something a user would need to care about I think we should either 
a) drop this option
b) put this option in an `**Advanced Options**` block below these main configuration variables

puuu: Since I am personally no longer interested in the query-mode, I would prefer to remove this option. This simplifies the code and we do not have to switch to `PollingComponent`.

OttoWinter (approved):
:+1:

components/sensor/sds011.rst
```diff
@@ -0,0 +1,81 @@
...
+------------------------
+
+- **pm_2_5** (*Optional*): Use the concentration of particulates of size less than 2.5µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **pm_10_0** (*Optional*): Use the concentration of particulates of size less than 10.0µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **update_interval** (*Optional*, :ref:`config-time`): The interval to check the sensor. This affects the working
+  period of the SDS011 sensor. Defaults to ``1s``.
```
puuu: 1s is not really right. The default value is 0 (in terms of configuration value), but the behavior is 1s.

OttoWinter: I see. What about changing the default update_interval to something like 5min?

The thought behind that is: If times >1min are preferred because of sensor lifetime, let's make that the default and users can still change it later.

PR #190, opened 2019-03-03
Add text-based sensor example to OLED cookbook entry

tubalainen:
## Description:

Update to the display cookbook


OttoWinter (approved):
Thanks! 🐦

PR #196, opened 2019-03-11
Update gpio.rst

john-arvid:
The inverted binary_sensor property has been replaced by the new 'invert' binary  sensor filter. Please see https://esphome.io/components/binary_sensor/index.html.

## Description:

Just a doc change regarding inverting binary sensors.

## Checklist:


OttoWinter:
Please check your indentation. The inverted part must be indented into the pin block. That is also the preffered way of doing it.

PR #202, opened 2019-03-20
Add to example code how to use the value retrieved from MQTT

kwdavidson:
Add to example code how to use the value retrieved from MQTT.

## Description:

## Checklist:


OttoWinter (changes requested):
- This example won't compile, a semicolon is missing.
- The example config shown at that part should always be bare minimum. More info like how to use the data should be shown in a new section on that page.

PR #203, opened 2019-03-20
Extend the documentation of esp8266_restore_from_flash to include pla…

placidorevilla:
…tforms that use the underlying component

## Checklist:

### DIFF
--- a/components/esphome.rst
+++ b/components/esphome.rst
@@ -187,7 +187,9 @@ To still have these components restore their state upon power loss the state can
 saved in *flash* memory by setting this option to ``true``.
 
 Beware: The flash has a limited number of write cycles (usually around 100 000), after that
-the flash section will fail.
+the flash section will fail. So do not use this option when you have components that update rapidly.
+These include GPIO switches that are used internally (disable restoring with the ``restore_mode`` option),
+certain light effects like ``random`` and the ``on_value_range`` trigger.
 
 .. _esphome-on_boot:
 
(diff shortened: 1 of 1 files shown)


OttoWinter (commented):
This is getting a bit long. Can this be moved to a new section below with a reference to it here?
see the other examples on this page (arduino_core_version, esphome_core_version etc)

```rst
# At end of esp8266_restore_from_flash block: 
See :ref:`esphome-esp8266_restore_from_flash` for more info.

# Below in a new section
.. _esphome-esp8266_restore_from_flash

``esp8266_restore_from_flash``
------------------------------

<Copy text from before>

placidorevilla:
Done

OttoWinter (commented):
I rewrote this section slighty, for these reasons:

- If we have a special section for this, we should also copy the block in the top `esp8266_restore_from_flash` over there.

- I added `like` to the component list that restore state. It is impossible to maintain this list accurately, as components that add restore functionality will be added in the future (and surely the contributors will miss updating this). Wrong info will result in users being confused.

- As I said, ESP32 have done it this way for a long time already and OTA has not been a problem. I dropped the section because the user shouldn't need to care about this and it will only confuse them.

- I took another look at `on_value_range`. It does only store if it's in the range or outside. For all uses I can think of (for example (de)humidifiers) this will only happen once every few hours or so). That will never be enough to wear the flash out too quickly.

PR #208, opened 2019-03-27
cookbook: bruh.rst: Add feels like calculation to dht sensor

efratomer:
## Description:
This PR adds a sensor template which calculates feels like according to dht's temprature & humidity readings.

## Checklist:

### DIFF
--- a/cookbook/bruh.rst
+++ b/cookbook/bruh.rst
@@ -39,8 +39,18 @@ configuration file 😀
         pin: D7
         temperature:
           name: "Multisensor Temperature"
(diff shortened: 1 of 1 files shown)


OttoWinter (changes requested):
This formula is way too long to be included in a sample file for a quick project - this config should be minimal and give basic operation to the device. This calculation should go somewhere else, for example the BME280 cookbook entry.

Please also state what this formula is based on, from all I see it could just be some arbitrary numbers

OttoWinter:
Closing due to inactivity.

PR #218, opened 2019-04-04
ifan02 cookbook

quazzie:
## Description:
Sonoff ifan02 cookbook.
It is not pretty but the info is correct.

## Checklist:


cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
+Ifan02
```
OttoWinter: Longer title please (include sonoff and fix capitalization)

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+======
+
+.. seo::
+    :description: Instructions for using Sonoff ifan02 in ESPHome.
+    :keywords: Fan, Sonoff, ifan02
```
OttoWinter, line 6: This needs an `:image:` too.

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+
+
+Configuration for `Sonoff ifan02 <https://www.itead.cc/sonoff-ifan02-wifi-smart-ceiling-fan-with-light.html>`__.
```
OttoWinter: Longer description please. Explain what the iFan02 is, what it does and what functionality the user will get in the end.

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+
+First you need a :doc:`components/output/custom.html` to control the ifan02.
```
OttoWinter:
```suggestion
First you need a :doc:`components/output/custom` to control the iFan02.
```

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Create a ifan02.h file:
+
+.. code-block:: c++
+
+    #include "esphome.h"
+    using namespace esphome;
+
+    class IFan02Output : public Component, public output::FloatOutput {
```
OttoWinter:
```suggestion
    class IFan02Output : public output::FloatOutput {
```

this component has no setup/loop function, so it doesn't need any component functionality.

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+              digitalWrite(5, HIGH);
+              digitalWrite(4, HIGH);
+              digitalWrite(15, LOW);
+            }
+            if (state >= 0.9) {
+              digitalWrite(5, HIGH);
+              digitalWrite(4, LOW);
+              digitalWrite(15, HIGH);
+            }
+        }
```
OttoWinter, line 48: rewrite this to be
```cpp
if (state < 0.3) {
  // OFF
} else if (state < 0.6) {
  // low speed
} else if (state < 0.9) {
  // medium speed
} else {
  // high speed
}
```

There are some values that this component does not handle like 0.31 (the if/else if fixes that kinda)

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+        pin: GPIO5
+        id: fan_relay1
+
+      - platform: gpio
+        pin: GPIO4
+        id: fan_relay2
+
+      - platform: gpio
+        pin: GPIO15
+        id: fan_relay3
```
OttoWinter, line 194: What all this gpio switches here when the custom output already handles those? This could potentially cause issues in the future if: for example the output platform writes the state more often and other stuff.

The custom output should be the only controller of these GPIOs. Please remove this logic and use `fan.turn_on` / `fan.turn_off` in the button triggers.

cookbook/ifan02.rst
```diff
@@ -0,0 +1,181 @@
...
+
+    fan:
+      - platform: speed
+        output: fanoutput
+        id: ifan02
+        name: ifan02_fan
```
OttoWinter: A see also section is missing here. It should point to some of the components used here (fan, custom output, etc)

OttoWinter (commented):
- If you don't add this file to index.rst, it won't show up on the front page. It needs to be added there in the cookbook section.

cookbook/ifan02.rst
quazzie: Hmm where to find a good free non copyrighted image ? Mine are installed so i can't take a picture.

cookbook/ifan02.rst
quazzie: This is because when you use the remote you get virtual button presses (vbutton_*) and you should toggle those relays. Depending on what speed you are on and what speed you want to go to diffrent vbutton fire. Pressing 1 on the remote will not fire the same vbuttons all the time. 
For example from pressing 1 on remote when fan is off the remote fires vbutton1, but pressing 1 when fan is on medium will fire vbutton2. 
And the vbuttons need to toggle corresponding relays.

cookbook/ifan02.rst
quazzie: i just used the fan.svg

CallMeAreks:
Hi. Is this example correct? I have the ifan02 already mounted so I wanted to confirm if it's working as intended before flashing via OTA.

workingmanrob:
Hey @CallMeAreks 

I've been using Quazzie's cookbook for a couple months now without issue.

I was running Tasmota firmware on my sonoff fan02 so I downloaded the esphome firmware after I built it then used the tasmota web UI to flash and everything is working as expected.

CallMeAreks:
@workingmanrob Thank you for the confirmation. I tried it as well like a week ago, (leap of faith because I have 4 already installed ifan02), but work flawlessly. Forgot to leave a comment, though.

pstechsolutions:
Not sure what made my case different.  I had an issue during compile where I would get an error about only being allowed to register sub-components.  The solution: In the  ifan02.h file the line 

class IFan02Output : public output::FloatOutput {

needed to be

class IFan02Output : public Component, public output::FloatOutput {

@OttoWinter and @quazzie Thank you both for your efforts and making them available for those of us, like myself, who are just learning!

CallMeAreks:
@pstechsolutions Yeah, now that you mention this I noticed the code in this commit history is slightly different from the one I copied: https://gist.github.com/quazzie/09ee3ef2c419ecbcf979a7410062481b

In this gist it reflects the change you made. Leaving the link here for future reference.

workingmanrob:
Updated to 1.13.2 am now getting a valication error:

Unknown value 'MEDIUM', did you mean 'MEDIuM'?

Pretty sure I mean MEDIUM and trying to use MEDIuM instead gave the same error.

Any ideas?

workingmanrob:
Looks like 1.13.3 fixed it.. thanks Otto!

mjoshd:
All is good now!

@quazzie Thanks for letting me assist!

@OttoWinter I have successfully compiled and installed this on my iFan02 with ESPHome 1.13.6 and it works 100% perfectly. Please merge this into the docs so others may benefit from it as well. 

Have a great day everyone! 😃

quazzie:
It's only stale because @OttoWinter has not responded.

quazzie:
Still not stale @OttoWinter

quazzie:
I guess it must be stale as nothing is happening...

mjoshd:
It appears that a link has been added in the DIY Examples
<https://esphome.io/guides/diy.html#sample-configurations> section but it's
pointing to your original gist and not to the code we worked on in the PR.

We should try to get that link re-pointed to the code from the PR at some
point.

PR #242, opened 2019-05-13
Improved the docs for Nextion Display

AalianKhan:
will add the examples soon

## Description:

## Checklist:

### DIFF
--- a/components/display/nextion.rst
+++ b/components/display/nextion.rst
@@ -51,7 +51,7 @@ Making a Display File
 ---------------------
 
 First, you need to use the `Nextion Editor <https://nextion.itead.cc/resources/download/nextion-editor/>`__ to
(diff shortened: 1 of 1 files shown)


AalianKhan:
@OttoWinter Examples will be hard in the table format because of the 160 character limit. Any other way to display information, or should I just scratch the table idea?

OttoWinter:
Yeah that's true - however even if the limit would be lifted for this file a table of such width also won't look good; especially on small screens.

So either:

 1. Add all this to the doxygen generated documentation and link to that better (so add examples etc to the .h file using doxygen format).

 2. Make this a list

Personally I would prefer 1) because doxygen is much better suited for API documentation than RST.

AalianKhan:
https://github.com/esphome/esphome/pull/548

PR #251, opened 2019-05-24
Add 'internal' to GPIO switch

Joshfindit:
In testing, ESPHome does honour `internal: true` for a GPIO switch.

Didn't add the following as I'm not sure if it's accurate:

```
Only specifying an ``id`` without  a ``name`` will implicitly set this to true.
```

## Description:

**No related issue**
**No related PR for [esphome](https://github.com/esphome/esphome) or [esphome-core](https://github.com/esphome/esphome-core)**

## Checklist:


OttoWinter (commented):
Yes internal is an option for GPIO Switches.

But, it's already documented if you follow the "All other options ..." link.

OttoWinter:
Closing due to inactivity.

PR #252, opened 2019-05-25
changed pca9685_id: to id:

alva-seal:
## Description:

Since ESPhome gave me an error that 'id: ' is required I think it should be only id: instead of pca9685_id:  and the Configuration variables say also id:.
So I wanted to help the next one to not have the same problem and changed it.

## Checklist:

### DIFF
--- a/components/output/pca9685.rst
+++ b/components/output/pca9685.rst
@@ -29,7 +29,7 @@ global ``pca9685`` hub and give it an id, and then define the
     # Individual outputs
     output:
       - platform: pca9685
-        pca9685_id: 'pca9685_hub1'
+        id: 'pca9685_hub1'
         channel: 0
 
 Configuration variables:


components/output/pca9685.rst
```diff
@@ -29,7 +29,7 @@ global ``pca9685`` hub and give it an id, and then define the
     # Individual outputs
     output:
       - platform: pca9685
-        pca9685_id: 'pca9685_hub1'
+        id: 'pca9685_hub1'
```
OttoWinter, line 32:
```suggestion
        id: 'pca9685_output1'
```

OttoWinter (commented):
Yeah the option `pca9685` does exist (for using multiple pca9685s with one ESP).

But in most cases it won't be required by the user.

The name of the ID is based on the `pca9685_id` option - if we're defining an ID for the *output* another name would be more accurate.

components/output/pca9685.rst
alva-seal: Yes, that's a better naming.

OttoWinter:
Closing due to inactivity.

PR #256, opened 2019-05-26
Add docs for CT Clamp

jesserockz:
## Description:
Adds documentation for CT (Current Transformer) Clamps.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#559

## Checklist:


OttoWinter:
Ok, I've updated this branch too with some smaller changes.

The biggest one is that the page now has a full calibration section outlining how to calibrate the device. It's using the calibrate_linear filter which is a bit more user-friendly than a direct multiply. Therefore the note about 3.3V adc is also not really required anymore because that's part of the calibration steps.

If you're ok with these changes, I will merge.

jesserockz:
I am happy with those changes, they make it look so much nicer.

Thanks

PR #257, opened 2019-05-29
Add SIM800L docs

glmnet:
## Description:

Adding docs for new SIM800L component

## Checklist:


components/sim800l.rst
```diff
@@ -0,0 +1,112 @@
...
+The ``SIM800L`` Component provides the ability to send and receive SMS text messages. The device must be
+connected via a :doc:`UART bus </components/uart>` supporting both receiving and transmitting line.
+
+The uart bus must be configured at the same speed of the module which is by default 9600bps. Either hardware
+or software uarts are supported. If you are using the hardware uart in the ESP8266, make sure you disable
+the  :doc:`logger`.
+
+.. note::
+
+    This module requires a custom power supply between 3.8v and 4.2v that can handle  current pikes up
```
OttoWinter:
```suggestion
    This module requires a custom power supply between 3.8v and 4.2v that can handle current spikes up
```

components/sim800l.rst
```diff
@@ -0,0 +1,112 @@
...
+.. seo::
+    :description: Instructions for setting up the SIM800L gsm module to send and receive SMS in ESPHome.
+    :image: sim800l.png
+    :keywords: SMS SIM800L GSM
+
+The ``SIM800L`` Component provides the ability to send and receive SMS text messages. The device must be
+connected via a :doc:`UART bus </components/uart>` supporting both receiving and transmitting line.
+
+The uart bus must be configured at the same speed of the module which is by default 9600bps. Either hardware
+or software uarts are supported. If you are using the hardware uart in the ESP8266, make sure you disable
```
OttoWinter: `Either harware or software uarts are supported.`

Can we remove this? Ideally esphome makes this as transparent as possible to the user.

Or alternatively, say here that if the user uses pins X and Y, they need to disable logging in logger with `baud_rate: 0`.

Maybe also add a quick paragraph about what connections need to be made: `VCC`, `GND`, `TX` and `RX`

components/sim800l.rst
```diff
@@ -0,0 +1,112 @@
...
+    This action can also be written in :ref:`lambdas <config-lambda>`:
+
+    .. code-block:: cpp
+
+        id(sim800l1).send_sms("+15551234567", "The message content");
+
+See Also
+--------
+
+- :apiref:`sim800l.h`
```
OttoWinter:
```suggestion
- :apiref:`sim800l/sim800l.h`
```

components/sim800l.rst
```diff
@@ -0,0 +1,112 @@
+Sim800L Component
+=================
+
+.. seo::
+    :description: Instructions for setting up the SIM800L gsm module to send and receive SMS in ESPHome.
+    :image: sim800l.png
```
OttoWinter:
```suggestion
    :image: sim800l.jpg
```

components/sim800l.rst
glmnet: I just want to add all most relevant information that might cause trouble. By saying this I am also saying that the TX / RX lines can be used for Sim800L (some user may believe it can only be used for logging)
May be adding another `.. note::` / `.. warning::` about using the `sim800l` and `logger` component at the same time.
I did not test it, but I guess it will fail if you don't explicitly disable the logger via serial. 
About the connections: looks like a good idea, there are some other pins which don't need to be connected and putting it clear will answer questions before they are made, like reset, it also has pins for speaker, mic, etc.
I'll push more later...

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+
+.. seo::
+    :description: Instructions for setting up the SIM800L gsm module to send and receive SMS in ESPHome.
+    :image: sim800l.jpg
+    :keywords: SMS SIM800L GSM
+
+The ``SIM800L`` Component provides the ability to send and receive SMS text messages. The device must be
+connected via a :doc:`UART bus </components/uart>` supporting both receiving and transmitting line. 
+The uart bus must be configured at the same speed of the module which is by default 9600bps. 
+The required connection wires are `+VCC`, `GND`, `RX` and `TX`.
```
OttoWinter:
```suggestion
The required connection wires are ``+VCC``, ``GND``, ``RX`` and ``TX``.
```

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+---------------------------
+
+With this configuration option you can write complex automations whenever a sms message
+is received. To use the message content, use a :ref:`lambda <config-lambda>`
+template, the message content and the sender phone number are available inside that lambda
+under the variables named ``message`` and ``sender`` respectively.
+
+.. code-block:: yaml
+
+  on_sms_received:
```
OttoWinter: Indent this so that it's more clear it's part of the sm800l block:

```rst
.. code-block:: yaml

  sim8000l:
    # ...
    on_sms_received:
```

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+            id: sim800l1
+            recipient: !lambda |-
+              if (id(reed_switch).state) return "+15551234567";
+              else return "15551234568";
+            message: !lambda |-
+              return id(reed_switch).state ? "Door is now OPEN" : "Hey door just CLOSED";
+
+Configuration options:
+
+- **id** (*Required*, string): The ID of the SIM800L
```
OttoWinter: See https://github.com/esphome/esphome/pull/522#discussion_r288912605 - this should not be a required option. Make it optional and move it to the bottom of the configuration options.

```suggestion
- **id** (*Optional*, :ref:`config-id`): The ID of the SIM800L module to send the SMS with,
  defaults to the first one in the config.
```

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+- **recipient** (*Required*, string, :ref:`templatable <config-templatable>`): The message recipient.
```
OttoWinter:
```suggestion
- **recipient** (**Required**, string, :ref:`templatable <config-templatable>`): The message recipient.
```

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+  number. Make sure to use quotes if the number starts with a `+` or the parser will remove it.
```
OttoWinter: Let's solve that in validation, string_strict will the validation print an error if the value is interpreted as a number and tells the user to use quotes. See https://github.com/esphome/esphome/pull/522/files#r288913659

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+- **message** (*Required*, string, :ref:`templatable <config-templatable>`): The message content.
```
OttoWinter:
```suggestion
- **message** (**Required**, string, :ref:`templatable <config-templatable>`): The message content.
```

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+    .. code-block:: cpp
+
+        id(sim800l1).send_sms("+15551234567", "The message content");
+
+
+Getting started with Home Assistant
+-----------------------------------
+
+The following code will get you up and running with a configuration updating received messages 
+on Home Assistant and will also setup a service so you can send messages with your Sim800L.
```
OttoWinter, line 112: Add a paragraph with an example of how this would be triggered in HA from an automation. For example

```yaml
automation:
  # ...
  action:
  - service: esphome.livingroom_send_sms
    data:
      recipient: "+15551234567"
      message: "Hello World!"
```

Also add a quick note what the text sensors do: Displaying the last seen SMS.

components/sim800l.rst
```diff
@@ -0,0 +1,156 @@
...
+
+.. code-block:: yaml
+
+    api:
+      services:
+      - service: send_sms
+        variables:
+          param_recipient: string
```
OttoWinter:
```suggestion
          recipient: string
```

same for param_message - param_ is not necessary (unless that leads to a compile error).

components/sim800l.rst
```diff
@@ -0,0 +1,165 @@
...
+.. warning::
+
+    If you are using the :doc:`logger` make sure you are not using the same pins for ``TX`` and ``RX`` or
+    otherwise disable the uart logging with the ``baud_rate: 0`` option.
+
+.. note::
+
+    This module requires a power supply between 3.8V and 4.2V that can handle current spikes up
+    to 2 amps, it will not work by powering from the same 3.3v power source of the ESP. However you can
+    connect `TX` and `RX` lines directly without any level shifter.
```
OttoWinter:
```suggestion
    connect ``TX`` and ``RX`` lines directly without any level shifter.
```

glmnet:
I’ll recheck all pin names have double back ticks and update here. I’m away from pc now.

glmnet:
Ok I believe I'm done, travis error seems to be something else

OttoWinter (approved):
🎉🎉🎉

PR #258, opened 2019-05-31
Add missing filter name for ct_clamp

jesserockz:
## Description:
Missing filter name in CT Clamp docs

## Checklist:


OttoWinter (approved):
Oh noes. Thanks!

PR #261, opened 2019-05-31
Document the backlight control for lcd_pcf8574

adarazs:
Add documentation about controlling the backlight for lcd_pcf8574
through the the lamda function.

## Description:

**Related issue (if applicable):** https://github.com/esphome/feature-requests/issues/160

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#573

## Checklist:

### DIFF
--- a/components/display/lcd_display.rst
+++ b/components/display/lcd_display.rst
@@ -112,7 +112,7 @@ Rendering Lambda
 
 The LCD displays has a similar API to the fully fledged :ref:`display-engine`, but it's only a subset as LCD displays
 don't have a concept of individual pixels. In the lambda you're passed a variable called ``it``
-as with all other displays. In this case however, ``it`` is an ``LCDDisplay`` instance.
+as with all other displays. In this case however, ``it`` is an instance of either ``GPIOLCDDisplay`` or ``PCF8574LCDDisplay``.
 
 The most basic operation with LCD Displays is writing static text to the screen as in the configuration example
 at the top of this page.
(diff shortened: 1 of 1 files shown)


components/display/lcd_display.rst
```diff
@@ -112,7 +112,11 @@ Rendering Lambda
 
 The LCD displays has a similar API to the fully fledged :ref:`display-engine`, but it's only a subset as LCD displays
 don't have a concept of individual pixels. In the lambda you're passed a variable called ``it``
-as with all other displays. In this case however, ``it`` is an ``LCDDisplay`` instance.
+as with all other displays. In this case however, ``it`` is an instance of either ``GPIOLCDDisplay`` or ``PCF8574LCDDisplay``.
+
+In case of ``lcd_pcf8574`` the backlight can be turned on by ``it.backlight()`` and off by ``it.no_backlight()`` when
+the left jumper is closed on the PCF8574 board. For ``lcd_gpio``, the backlight pin of the display needs
+to be controlled separately.
```
OttoWinter: Can we move this paragraph down to below the code example? I'd like to keep the most important part (drawing text) at the beginning - backlight can be after that for advanced use.

adarazs: Sure.

Do you think that in that advanced part would be a good place (or in general useful) if I would add an example with a motion sensor and timeout for the backlight? I assume that's a quite common thing to do. I would add this code to it which I posted to https://github.com/esphome/feature-requests/issues/160#issuecomment-497962487 -- also let me know if this is an ugly solution and there's a nicer one if you think I should add anything regarding it. :)

OttoWinter: Sure yeah that could give users some more examples.

For the yaml code from the fr: It can be done that way yes. If you're going to do it with a template binary sensor though use the binary_sensor.template.publish instead so that the user doesn't need to leave YAML land.

Another option would be to start a script in on_press that has a delay and then turns the backlight off.

In on_press, turn on the backlight, call script.stop, then script.execute. Could be more messy though, you can try that solution and ise the template binary sensor if it turns out to be more complex

adarazs:
I added a new revision with a new section for backlight control. I went with the binary_sensor.template based example instead of the script (sounded nice though), mostly because I did not use any scripts yet and I don't have more time to dedicate to this doc change (and afraid to mess it up with not well tested code). This example code works well for me in the last few days.

OttoWinter (approved):

PR #266, opened 2019-06-05
Add timezone mapping to Docker compose.

jeff-h:
## Description:
As per https://github.com/esphome/issues/issues/398

### DIFF
--- a/guides/faq.rst
+++ b/guides/faq.rst
@@ -259,6 +259,7 @@ And a docker compose file looks like this:
         image: esphome/esphome
         volumes:
           - ./:/config:rw
+          # Use local time for logging timestamps
           - /etc/localtime:/etc/localtime:ro
         network_mode: host
         restart: always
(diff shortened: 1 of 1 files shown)


guides/faq.rst
```diff
@@ -259,9 +259,12 @@ And a docker compose file looks like this:
         image: esphome/esphome
         volumes:
           - ./:/config:rw
+          - /etc/localtime:/etc/localtime:ro
```
OttoWinter, line 263:
```suggestion
          # Use local time for logging timestamps
          - /etc/localtime:/etc/localtime:ro
```

I think the explanation below can be moved into a simple comment in the docker compose YAML.

guides/faq.rst
```diff
@@ -259,9 +259,12 @@ And a docker compose file looks like this:
...
         network_mode: host
         restart: always
 
+To ensure the timezone of the ESPHome container is correct, only map `/etc/localtime`; ESPHome has no need for `/etc/timezone` nor a `TZ` environment variable.
```
OttoWinter: Remove this line - the user really shouldn't care *what* file are included in general docker, the docker compose example already shows how it's done for ESPHome.

jeff-h:
Good ideas. I've made the changes.

OttoWinter (approved):

PR #269, opened 2019-06-07
Lambas won't just work with the short name

ryannazaretian:
lambdas won't accept IDLE, OPENING, or CLOSING. Need to specify COVER_OPERATION_IDLE, COVER_OPERATION_OPENING, or COVER_OPERATION_CLOSING.

## Description:

## Checklist:


OttoWinter (commented):
Well with this change, non-lambda users will be confused.

Please change this back to before and add a quick sentence about what this looks like in lambdas.

ryannazaretian (commented):
Updated current_operation based on feedback. Updated the state to be consistent, also fixed the position range lower bound to floating point.

OttoWinter (approved):
Thanks!

PR #274, opened 2019-06-12
added D1,2 numbering on ESP8266 to avoid confusion

bazza2000:
Spent about 2 hours debugging why the I2C bus wasn't seeing any components until I realised that GPIO4 = D2 and GPIO5 = D1.
Will hopefully help other who are using the same board.

## Description:

## Checklist:

### DIFF
--- a/components/i2c.rst
+++ b/components/i2c.rst
@@ -25,10 +25,10 @@ Configuration variables:
 ------------------------
 
 - **sda** (*Optional*, :ref:`config-pin`): The pin for the data line of the I²C bus.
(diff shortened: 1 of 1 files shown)


OttoWinter (changes requested):
No,

Each ESP8266 board has different pin alias names (see also https://github.com/esphome/esphome/blob/dev/esphome/pins.py) - D1 is different for each board. That's why esphome consistently uses the internal pin names GPIOx for these things.

If you click on the link net to the sda/scl option, you will also see a page where this is explained.

bazza2000:
Understood,

Thanks for the clarification.

PR #282, opened 2019-07-01
Add Shelly RGBW2 pinout to sonoff.rst

erdii:
## Description:

## Checklist:

### DIFF
--- a/devices/sonoff.rst
+++ b/devices/sonoff.rst
@@ -311,7 +311,7 @@ Shelly 2
     GPIO14, SW Input #2,
     
 Shelly RGBW2
---------
+------------
 
 .. pintable::
 
--- a/devices/sonoff.rst
+++ b/devices/sonoff.rst
@@ -309,6 +309,18 @@ Shelly 2
     GPIO5, Relay #2,
     GPIO12, SW Input #1,
     GPIO14, SW Input #2,
+    
+Shelly RGBW2
+--------
+
+.. pintable::
+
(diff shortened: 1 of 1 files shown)


devices/sonoff.rst
```diff
@@ -309,6 +309,18 @@ Shelly 2
     GPIO5, Relay #2,
     GPIO12, SW Input #1,
     GPIO14, SW Input #2,
+    
+Shelly RGBW2
+--------
```
OttoWinter:
```suggestion
------------
```

devices/sonoff.rst
```diff
@@ -309,6 +309,18 @@ Shelly 2
...
+
+.. pintable::
+
+    GPIO12, LED Red,
+    GPIO15, LED Green,
+    GPIO14, LED Blue,
+    GPIO04, LED White,
+    GPIO??, SW Input,
+    GPIO??, Power Meter,
```
OttoWinter, line 323: Do not add information here if we don't know the actual pin.

Also, the power meter chip on the Shelly is not supported - add a line below saying that power metering is not supported at this point.

erdii: So do I either remove the unknown pins or add information about the unsupported power meter?

OttoWinter: Both would be best :)  (so remove the unknown pins and add a quick comment below as for the power meter chip not being supported)

PR #288, opened 2019-07-08
add coolix receiver

glmnet:
## Description:
Coolix receiver :)

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome# https://github.com/esphome/esphome/pull/645

## Checklist:

### DIFF
--- a/components/climate/coolix.rst
+++ b/components/climate/coolix.rst
@@ -13,6 +13,9 @@ This component requires that you have setup a :doc:`/components/remote_transmitt
 Due to the unidirectional nature of IR remote controllers, this component cannot determine the
 actual state of the device, and will assume the state of the device is the latest state requested.
 
+Optionally you can add a :doc:`/components/remote_receiver` component so the climate state will be
+tracked when it is operated with the original remote controller unit.
+
 .. figure:: images/climate-ui.png
     :align: center
     :width: 60.0%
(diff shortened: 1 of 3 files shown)


components/climate/coolix.rst
```diff
@@ -36,13 +39,23 @@ Configuration variables:
...
 
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
 - **transmitter_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the remote transmitter.
 
+.. _coolix-receiver_id:
+
+Using a receiver
+----------------
+
+When using a receiver it is recommended to put the IR receiver as close as the equipment's IR receiver
```
OttoWinter:
```suggestion
When using a receiver it is recommended to put the IR receiver as close as possible to the equipment's IR transmitter
```

components/climate/coolix.rst
```diff
@@ -36,13 +39,23 @@ Configuration variables:
...
+so the chances of receiving (and not receiving) IR signals are the same as of the AC unit.
```
OttoWinter: I mean yes, that's one way of putting it, but wouldn't it be simpler to say: Just "put the receiver as close as possible"? Is there any reason the chance of receiving and transmitting have to be equal?

components/climate/coolix.rst
```diff
@@ -36,13 +39,23 @@ Configuration variables:
...
+It is recommended to use a high tolerance, like 50 or so in the receiver configuration.
```
OttoWinter: A short example configuration below would really help :)

components/climate/coolix.rst
glmnet: Ohhh what I'm trying to say is to put the receiver as close as possible to the equipment's original IR receiver, so both receivers next to each other.
This way when a signal is sent by someone using the remote control unit, it really does not depend much where this person is on the room.

PR #290, opened 2019-07-10
Add rc switch setup tips

glmnet:
## Description:
add info on troubleshooting rc switch

**Related issue (if applicable):** https://github.com/esphome/issues/issues/508

## Checklist:


components/remote_transmitter.rst
```diff
@@ -533,6 +533,13 @@ earlier, create a new template switch that sends the RF code when triggered:
 Recompile again, when you power up the device the next time you will see a new switch
 in the frontend. Click on it and you should see the remote signal being transmitted. Done!
 
+If you cannot get it to work at first time, try setting up a different node for transmitting
+and receiving, then check that your transmitter node and remote control causes the same received
+code on the receiver logs, some devices requires that the transmitted code be repeated for the
+signal to be picked as valid. Also the interval between repetitions can be important. Check
+that the pace of repetition logs are consistent between the remote controller and the transmitter
+node. You can adjust the `repeat:` settings as needed.
```
OttoWinter, line 541: This block does describe the problem well, but I think it can be replaced with a short example - just paste a short code-block with a `repeat:` set to 3 or something and tell the user to use that. I doubt many users will understand what to do from this block of text immediately.

OttoWinter: Bump.

PR #295, opened 2019-07-14
Updated esp32 description

cptquad:
Updated ping 34-39 direction to input from output as the esphome editor would not let me use these pins as output.

## Description:

## Checklist:


OttoWinter (changes requested):
No, the current text is correct.

Pins GPIO34 through GPIO39 can only be used as inputs - so they cannot be used as outputs.

PR #317, opened 2019-08-14
Add cookbook for Sonoff T1 US 3 Gang

genlink:
## Description:

Add cookbook for Sonoff T1 US 3 Gang


glmnet:
Hello!
Thanks for contributing

It seems that you are adding the same file to two different places.
Also you need to add somewhere a link to access your file.

brandond:
I have one of these myself. The pinout is exactly the same as the UK devices that are already covered here: https://esphome.io/devices/sonoff_t1_uk_3gang_v1.1.html

PR #321, opened 2019-08-23
7.50in waveshare e-paper works

rafl:
## Description:

Update documentation for a device that the current code supports.

## Checklist:

    The current code works as is, so that seems like the appropriate target.


fabaff (approved):

OttoWinter (approved):
Thanks! 👍

PR #323, opened 2019-08-25
doc improvement: its /config not config/

facastagnini:
its /config not config/
Reference  https://github.com/esphome/esphome/blob/14fd08e22591fbd2ecef6c37c4d6ed7a3cfd2bfa/docker/Dockerfile#L9

## Description:

## Checklist:


OttoWinter (changes requested):
No no, the current doc is correct.

`/config` just happens to be the path where the directory is mapped to in the Hass.io image.

For user setups (not Hass.io), you should not have to create a new dir in the root directory. A relative path at it is currently is correct.

facastagnini:
got it, thanks!

PR #329, opened 2019-08-27
Add description of id Parameter

guptamp:
see here: https://github.com/esphome/issues/issues/625

## Description:

## Checklist:

### DIFF
--- a/components/i2c.rst
+++ b/components/i2c.rst
@@ -54,7 +54,7 @@ Configuration variables:
    - platform: bme680
      i2c_id: bus_b
      address: 0x76
-     ...
+     # ...
   
 See Also
 --------
--- a/components/i2c.rst
+++ b/components/i2c.rst
@@ -35,11 +35,11 @@ Configuration variables:
 - **frequency** (*Optional*, float): Set the frequency the I²C bus should operate on.
   Defaults to ``50kHz``. Values are ``50kHz``, ``100kHz``, ``200kHz``, ... ``800kHz``
 - **id** (*Optional*, :ref:`config-id`) Manually specify the ID for this I²C bus if you need multiple I²C buses.
-  
+
 .. note::
 
-  If the device can support multiple I²C buses (ESP32 has 2) these buses need to be defined as below and sensors need to be setup specifying the correct bus:
-  
+    If the device can support multiple I²C buses (ESP32 has 2) these buses need to be defined as below and sensors need to be setup specifying the correct bus:
+
 .. code-block:: yaml
 
     # Example configuration entry
@@ -57,7 +57,7 @@ Configuration variables:
      i2c_id: bus_b
      address: 0x76
      # ...
-  
+
 See Also
 --------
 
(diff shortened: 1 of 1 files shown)


components/i2c.rst
```diff
@@ -33,6 +34,7 @@ Configuration variables:
   Defaults to ``True``.
 - **frequency** (*Optional*, float): Set the frequency the I²C bus should operate on.
   Defaults to ``50kHz``. Values are ``50kHz``, ``100kHz``, ``200kHz``, ... ``800kHz``
+- **id** (Optional, ID): Manually specify the ID for this i2c Bus if you need multiple i2c buses.
```
OttoWinter:
```suggestion
- **id** (*Optional*, :ref:`config-id`) Manually specify the ID for this I²C bus if you need multiple I²C buses.
```

Also add a small `.. note:` below for explaining how this is used on the receiving end (e.g. i2c_id)

guptamp: I was about to ask where I should put this.
Does it make sense to add this explanation to sensor core as it is applicable to at least all sensors that are i2c, spi or uart?

OttoWinter: No, sensor core is only about sensors.

Stuff like how multiple i2c buses can be used at once should go in the i2c page - that's where most users would look for first too.

glmnet: Yes! You need to say that only 1 is supported in ESP8266 and only 2 are supported in ESP32

OttoWinter: Does ESP8266 only support 1? I thought ESP8266 used a sw i2c implementation, so should allow multiple ones.

glmnet: Ahhh yes ESP8266 is sw. Sorry, my bad.
But ESP32 only supports 2 via hardware, and then esp8266 will be limited by pin availability then.
It will be worth mentioning that to address https://github.com/esphome/issues/issues/539

guptamp: From what i saw ESP32 can have 3 hw but i have only tested 2 so far.
I´m creating a new pull request with more details in a few minutes.
Any thing else i should include?

glmnet: @guptamp according to this https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf section 4.1.11 there are only 2

guptamp: I think your right and I can´t find the website I saw it before anymore.

components/i2c.rst
```diff
@@ -33,7 +34,28 @@ Configuration variables:
   Defaults to ``True``.
 - **frequency** (*Optional*, float): Set the frequency the I²C bus should operate on.
   Defaults to ``50kHz``. Values are ``50kHz``, ``100kHz``, ``200kHz``, ... ``800kHz``
+- **id** (*Optional*, :ref:`config-id`) Manually specify the ID for this I²C bus if you need multiple I²C buses.
+  
+  *Note:*
```
OttoWinter: In RST, you do notes like this:

```rst
.. note::

	Text that goes in note
```

https://esphome.io/guides/contributing.html#syntax

components/i2c.rst
```diff
@@ -33,7 +34,28 @@ Configuration variables:
...
+        scan: True
+      - id: bus_b
+        sda: 14
+        scl: 15
+        scan: True
+   # Sensors should be specified as follows
+   - platform: bme680
+     i2c_id: bus_b
+     address: 0x76
+     ...
```
OttoWinter:
```suggestion
     # ...
```

Otherwise the YAML parser will not like it

guptamp:
Any reason why this is still blocked?
Did I do the updated pull request wrong? (I´m still new to GitHub)

PR #332, opened 2019-08-29
Update esp32_camera.rst jpeg quality

glogiotatidis:
## Description:

Make clear which quality setting is the best

## Checklist:


OttoWinter (approved):
Thanks!

PR #334, opened 2019-09-01
Cookbook ape

glmnet:
## Description:

Using an Arduino as a I/O port extender / expander.

This is something that I've been using for a while, I've invested time in debugging and getting to work reliably with ESPHome so I think I should share it.

## Checklist:


cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
+Arduino Port Extender
+=====================
+
+.. seo::
+    :description: Instructions on using an Arduino board, like the Pro Mini for extending ports of a ESPHome node
+    :image: arduino_pro_mini.jpg
+    :keywords: Arduino port extender ESPHome
+
+The Arduino is a widely used microcontroller, easy to use and accesible.
+With this sketch you can use any compatible Arduino board with ESPHome and read and write pins.
```
OttoWinter: With this sketch you can control pins of a remote Arduino board through ESPHome. The Arduino acts as a port expander, allowing you to use more pins than a standard ESP8266/ESP32 has.

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+
+
+.. figure:: images/arduino_pro_mini.jpg
+    :align: center
+    :width: 75.0%
+
+
+The Arduino is connected to the ESP via I²C, common Arduinos uses pins ``A4`` and ``A5`` for this
```
OttoWinter:
```suggestion
The Arduino is connected to the ESP via I²C. Most Arduinos use the ``A4`` and ``A5`` pins for the I²C bus
```

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+so those pins are not available to read from ESPHome.
+It is recommended to use a 3.3V Arduino, however using 5V arduinos seems to work too. In the last
```
OttoWinter:
```suggestion
It is recommended to use a 3.3V I/O level Arduino, however using 5V Arduinos seems to work too. In the latter
```

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+case you should power your 5V Arduino with 3.3V otherwise you will need a level converter for the
```
OttoWinter, line 19: I doubt you can run 5V microcontrollers off of 3.3V - The chip should detect a brownout and just shut off.

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+I²C bus.
+
+Currently it is supported:
+
+    - reading digital inputs
+    - reading analog inputs
+    - writing digital outputs
+
+The Arduino sketch can be retrieved from `Here <https://github.com/glmnet/esphome_devices/tree/master/ArduinoPortExtender/src>`__
```
OttoWinter: Can you make sure these links stay up? (If you use your normal devices repo and you change the folder structure all these links will break).

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+
+You need to download `ape.h <https://github.com/glmnet/esphome_devices/blob/master/ape.h>`__ and include the ape.h in the ESPHome configuration.
+
+.. code-block:: yaml
+
+    esphome:
+      ...
```
OttoWinter:
```suggestion
      # ...
```

YAML formatter doesn't like it otherwise.

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+      includes:
+          - ape.h
```
OttoWinter:
```suggestion
        - ape.h
```

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+
+Currently it is supported:
+
+    - reading digital inputs
+    - reading analog inputs
+    - writing digital outputs
+
+The Arduino sketch can be retrieved from `Here <https://github.com/glmnet/esphome_devices/tree/master/ArduinoPortExtender/src>`__
+
+You need to download `ape.h <https://github.com/glmnet/esphome_devices/blob/master/ape.h>`__ and include the ape.h in the ESPHome configuration.
```
OttoWinter: Can we rename this file to `arduino_port_expander.h` or something like that? Ape has a bit of a different meaning :)

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+                ape_analog_input(ape, 2)};
+      sensors:
+        - name: Analog A1
+          id: analog_a1
+          filters:
+            - throttle: 1s
+        - name: Analog A2
+          id: analog_a2
+          filters:
+            - throttle: 2sec
```
OttoWinter:
```suggestion
            - throttle: 2s
```

docs usually use `s`

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+
+The listed `sensors` supports all options from :ref:`Sensor <config-sensor>` like
```
OttoWinter:
```suggestion
The listed ``sensors`` supports all options from :ref:`Sensor <config-sensor>` like
```

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+automations and filters.
+
+.. note::
+
+  Sensors are polled by default every loop cycle so it is recommended to use the ``throttle`` filter
```
OttoWinter: The normal binary_sensor.gpio also does that - is there any reason why this component would suffer from this problem more?

cookbook/arduino_port_extender.rst
```diff
@@ -0,0 +1,183 @@
...
+  switch:
+    - platform: output
+      name: Switch pin 3
+      output: output_pin_3
+
+  light:
+    - platform: binary
+      name: Switch pin 4
+      output: output_pin_4
+
```
OttoWinter, line 321: It would be great to have a link to a full example here at the end - just helps tremendously when reading through a guide.

cookbook/arduino_port_extender.rst
glmnet: Of course not, should I use a gist or what do you recommend?

cookbook/arduino_port_extender.rst
glmnet: I understand your concern, this is something you should not recommend to do, however I explain it seems to work because I believe many people will fall in this case, having and ESP and a 5V Arduino.
I have 10 boards which were using a 5v arduino pro mini, powered with 12V and regulated down to 5 using a 7805 IC. They were using the NRF24 radio, and I decided to convert them to WiFi and long story short ended up here.
I changed the 7805 by the LD1117V33 which can easily do 950mA, and had no issue with any of that. 
I did some investigation just in case, found out the chips are the same but 3.3 are run at 8Mhz instead of 16Mhz.
In the desktop breadboard I have a LoLin ESP8266, I've powered two 5V pro minis and connected via I2C no problem either (with the LoLin 3v3 regulator.

cookbook/arduino_port_extender.rst
glmnet: Will do.

cookbook/arduino_port_extender.rst
glmnet: will do

cookbook/arduino_port_extender.rst
glmnet: The sensor [logs a VERBOSE](https://github.com/esphome/esphome/blob/fd1dc24ac6eaddad42fd5d9e9221542d82413db0/esphome/components/sensor/sensor.cpp#L13) message every time I call publish_state, binary_sensor filters same value logging and uses DEBUG instead.
The other way would be me adding an update interval for the analog inputs, but then I'll need to make this configurable, I found it easier using filters.

Now I see it's ok not adding this, I thought those were DEBUG messages too and they flooded my log so I filtered them this way.

I really don't know, if the throttle filter is not there wouldn't it be sending messages to Home Assistant too quickly?
Should I implement `update_interval`?

cookbook/arduino_port_extender.rst
glmnet: Otherwise I could create a tag?

glmnet: May be a gist?

OttoWinter: Yeah, that would be fine. On second thought though the current setup is fine too, we can always change it later.

cookbook/arduino_port_extender.rst
OttoWinter: Well, the users here will probably be also ones that will need to debug some things at some point (meaning these are the type of people who write code, where debugging is really good to have). This essentially makes debugging much harder, so ideally yes there would be an update interval. It's fine for now, but might we worth looking at in the future.

OttoWinter (approved):
Cookbook looks good for me 👍👍👍

Left a comment about the update interval: https://github.com/esphome/esphome-docs/pull/334/files#discussion_r336752283

It's ok as-is, you can press merge when you feel ready to.

PR #348, opened 2019-09-25
update esp32camera documentation - note about PWM

h0--:
As I was fighting with this yesterday, I think this can be part of esphome documentation.

## Checklist:

### DIFF
--- a/components/esp32_camera.rst
+++ b/components/esp32_camera.rst
@@ -91,6 +91,8 @@ Frame Settings:
 
     The camera integration in Home Assistant isn't in stable or beta HA builds, so until 0.91
     ships or goes into beta you will need to use a development version of Home Assistant.
+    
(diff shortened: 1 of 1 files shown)


components/esp32_camera.rst
```diff
@@ -91,6 +91,8 @@ Frame Settings:
 
     The camera integration in Home Assistant isn't in stable or beta HA builds, so until 0.91
     ships or goes into beta you will need to use a development version of Home Assistant.
+    
+    Camera uses PWM timer #1. If you need PWM (e. g. in ESP32 LEDC Output) you need to specify different channel there.
```
glmnet: did you get a 50% flash led or something like that? can you also add a quick example of the ledc?

glmnet (changes requested):
May be add this in a sub heading about LEDC and controlling the LED flash

PR #359, opened 2019-10-14
The 4.20in display works fine

fredrike:
## Description:

I've been running ESPhome with a 4.20in display successfully for some time.

## Checklist:


OttoWinter:
Thanks 👍

PR #366, opened 2019-10-18
add win10 wsl info

glmnet:
## Description:
Created PR per Feature request. I successfully installed the dashboard on the WSL

P.S. Feature Request 404 actually exists 😊

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/404

## Checklist:

### DIFF
--- a/guides/getting_started_command_line.rst
+++ b/guides/getting_started_command_line.rst
@@ -21,6 +21,8 @@ Installing ESPHome is very easy. All you need to do is have `Python
 
     pip install esphome
 
+Don't forget to add the esphome binary to PATH to ensure proper execution.
+
 Alternatively, there’s also a docker image available for easy
 installation (the docker hub image is only available for amd64 right now; if you have
 an RPi, please install ESPHome through ``pip`` or use :doc:`the Hass.io add-on <getting_started_hassio>`:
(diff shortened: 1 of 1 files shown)


guides/getting_started_command_line.rst
```diff
@@ -29,6 +31,14 @@ an RPi, please install ESPHome through ``pip`` or use :doc:`the Hass.io add-on <
 
     docker pull esphome/esphome
 
+Installation on Windows 10
+--------------------------
+
+It's possible to also install ESPHome command line in Windows 10, inside the `Windows Subsystem for Linux
+(WSL) <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`__ using the same ``pip`` command as
+above. The main limitation is that COM ports are not exposed properly in WSL, so only OTA flashing works.
+Or local flashing from Windows using `esphome-flasher <https://github.com/esphome/esphome-flasher>`__.
```
OttoWinter, line 40: Hmm, I don't think this is important enough to put in the main getting started guide. This guide should be bare minimum with no big tangents for specific setups.

How about moving this to the FAQ and linking that from here?

guides/getting_started_command_line.rst
```diff
@@ -21,6 +21,8 @@ Installing ESPHome is very easy. All you need to do is have `Python
 
     pip install esphome
 
+Don't forget to add the esphome binary to PATH to ensure proper execution.
```
OttoWinter, line 24: I doubt the average user will understand what this is supposed to mean. Maybe it would fit better in the FAQ entry about windows. And definitely add a short code block showing what output is seen when it's not in the PATH (command not found or whatever it is on windows) - that makes the issue searchable.

Additionally, an example of where the binary usually is would be useful (guessing something like user appdata somewhere?)

PR #374, opened 2019-10-22
Logger on_message trigger

Anonym-tsk:
## Description:
Ability to perform actions when messages appear in the log.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#729

## Checklist:

### DIFF
--- a/components/logger.rst
+++ b/components/logger.rst
@@ -159,6 +159,28 @@ Configuration options:
 -  **tag** (*Optional*, string): The tag (seen in front of the message in the logs) to print the message
    with. Defaults to ``main``.
 
+Logger Automation
+-----------------
+
+.. _logger-on_message:
+
+``on_message``
+**************
+
+This automation will be triggered when a new message is added to the log.
+In :ref:`lambdas <config-lambda>` you can get the message, log level and tag from the trigger
+using ``message``, ``level`` and ``tag``.
+
+.. code-block:: yaml
+
+    logger:
+      # ...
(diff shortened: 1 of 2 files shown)


components/logger.rst
```diff
@@ -159,6 +159,28 @@ Configuration options:
...
+-----------------
+
+.. _logger-on_message:
+
+``on_message``
+**************
+
+This automation will be triggered when a new message is added to the log.
+In :ref:`lambdas <config-lambda>` you can get the message, log level and tag from the trigger
+using ``message``, ``level`` and ``tag``.
```
OttoWinter: For this trigger it would be useful to add the type information here - most triggers use the C++ std::string class for ease of use. Logger is an exception to this (because dynamic memory std::string uses is not ideal).

Maybe add the C type here in parentheses (`const char *`, `int`)

components/logger.rst
```diff
@@ -159,6 +159,28 @@ Configuration options:
...
+
+.. code-block:: yaml
+
+    logger:
+      # ...
+      on_message:
+        level: ERROR
+        then:
+          lambda: |-
+            ESP_LOGD("DEBUG", "Triggered on_message with level %s, tag %s and message %s", level, tag, message);
```
OttoWinter: level is an int, not a string.

OttoWinter (approved):

OttoWinter (approved):

PR #376, opened 2019-10-22
Added degree symbol output on ~ for MAX7219

cyberplant:
## Description:
Added degrees symbol output when input is ~

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#764

## Checklist:


cyberplant:
~Please don't merge it yet, I have to create and upload the SVG. And learn how to do it properly in the way~ 😄

OttoWinter:
@cyberplant Ok, if I remember correctly I auto-generated those SVGs with a small script (can't find it anymore though). If you look at the SVG I think you can figure out which segments maps to what to create the layout.

cyberplant:
Yes, I like that idea more than learning how to use inkscape (again, for the 5th time).

Right now I'm super busy with work (as|and) I have holidays next week so I think I'll delay this some days, but I'll work on it soon.

cyberplant:
Hello! I've modified the SVG file manually and now looks good.

I've randomly named the image file with last two characters "8E", as 7E was already used by "0". Is that correct?

OttoWinter (approved):
Thanks very much! And sorry for the review delay.

Yes the filename is good - no problem.

cyberplant:
No problem.. thanks to you for the great ESPHome!

glmnet:
Ahhh this should be extended to the tm1637 too. Unfortunately it wasn’t committed this one before. But maybe is just adding a line to the tm1637 cpp file

PR #378, opened 2019-10-25
change docs to suggest logger config

glmnet:
using sample config instead of warning at top to match other uart components

## Description:

## Checklist:


OttoWinter (commented):
Thanks!

PR #390, opened 2019-11-01
Create meamor_3way-switch.yaml

Dmace99:
This is an addition to the Sonoff 3-way switch. The meamor is a european switch using the TYWE3S ESP8266. 
Where the Sonoff only can handle 2 Amps, the Meamor can handle 10 Amps. 
A Manual will be created as well

## Description:

## Checklist:

### DIFF
--- a/devices/meamor_3way-switch.yaml
+++ b/devices/meamor_3way-switch.yaml
@@ -1,0 +1,100 @@
+esphome:
+  name: meamor_3switch
+  platform: ESP8266
+  board: esp01_1m
+#The Meamor is identical to the Sonoff UK 3, but can handle 10 Amps. 
+#You can find it on Amazon.de searching for "Meamor Lichtschalter"
+
+wifi:
+  ssid: 'MySSID'
+  password: 'MyPassword
+    static_ip: 192.168.1.2
+    gateway: 192.168.1.1
+   
+time:
+  - platform: homeassistant
+    id: homeassistant_time
+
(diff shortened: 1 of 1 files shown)


brandond:
What's the deal with interlocking the relays and LEDs? Wouldn't this make it so you can only have one of the three lights in your house on at a time?

Dmace99:
Hi Brandon,

This is a setup for a central ventilation system. It uses a low-medium-high
setting by allowing 3 different power inputs over a perilex connector.
Since i don't want to overload the system, I can only use one switch at a
time.

Kind regards,

Danny

Op vr 1 nov. 2019 om 19:06 schreef Brandon Davidson <
notifications@github.com>:

devices/meamor_3way-switch.yaml
```diff
@@ -0,0 +1,100 @@
...
+  name: meamor_3switch
+  platform: ESP8266
+  board: esp01_1m
+#The Meamor is identical to the Sonoff UK 3, but can handle 10 Amps. 
+#You can find it on Amazon.de searching for "Meamor Lichtschalter"
+
+wifi:
+  ssid: 'MySSID'
+  password: 'MyPassword
+    static_ip: 192.168.1.2
```
OttoWinter, line 11: This is not valid YAML - this file fails to validate.

Also, please create a reference to this file on the generic sonoff page. Otherwise noone will ever find it.

PR #399, opened 2019-11-02
Add MAX31865, update MAX31855

DAVe3283:
## Description:

Update documentation for the new MAX31865 sensor and the improvements to the MAX31855 sensor.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#832

## Checklist:


components/sensor/max31855.rst
```diff
@@ -30,16 +30,17 @@ Connect ``GND`` to ``GND``, ``VCC`` to ``3.3V`` and the other three ``MISO`` (or
         name: "Living Room Temperature"
         cs_pin: D2
         update_interval: 60s
+        reference_temperature:
```
OttoWinter: Does the normal user care about this value? If not, it should be removed here (should be minimal sample that the user can copy-paste)

components/sensor/max31865.rst
```diff
@@ -0,0 +1,70 @@
...
+As the communication with the MAX31865 is done using SPI, you need
+to have an :ref:`spi bus <spi>` in your configuration with both **miso_pin** and **mosi_pin** set.
+
+- ``VIN`` connects to 5V (``3V3`` will output 3.3V), or directly connect ``3V3`` to 3.3V
+- ``GND`` connects to ground
+- ``CLK`` connects to the SPI **clk_pin**
+- ``SDO`` connects to the SPI **miso_pin**
+- ``SDI`` connects to the SPI **mosi_pin**
+- ``CS`` connects to a free GPIO pin
+- ``RDY`` is not used by this library
```
OttoWinter:
```suggestion
- ``RDY`` is not used by ESPHome
```

ESPHome is strictly speaking not a library, but a framework.

components/sensor/max31865.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Configuration variables:
+------------------------
+
+- **name** (**Required**, string): The name for the temperature sensor.
+- **cs_pin** (**Required**, :ref:`Pin Schema <config-pin_schema>`): The Chip Select pin of the SPI interface.
+- **update_interval** (*Optional*, :ref:`config-time`): The interval to check the sensor. Defaults to ``60s``.
+- **reference_resistance** (**Required**, float): Reference resistor on the PCB. Adafruit's PT100 (#3328) uses 430 Ω, their PT1000 (#3648) uses 4300 Ω.
+- **rtd_nominal_resistance** (**Required**, float): Nominal resistance of the RTD at 0°C. PT100 is 100 Ω, PT1000 is 1000 Ω.
+- **filter** (*Optional*, string): The mains power frequency to reject (``50 Hz`` or ``60 Hz``). Defaults to ``60 Hz``.
```
OttoWinter: I think we should rename this option to `frequency_filter` or `mains_filter`.

Just `filter` is too close to `filters` used for sensors, and can be confusing.

components/sensor/max31855.rst
DAVe3283: Good call. I will clean this up this evening.

components/sensor/max31865.rst
DAVe3283: Understood. I will implement your suggestions this evening.

components/sensor/max31865.rst
DAVe3283: I like `mains_filter` as it helps avoid the implication that arbitrary frequencies can be filtered, and aligns with the datasheet specified intent of the filter:

> Noise from 50Hz or 60Hz power sources (including harmonics  of  the  ac  power’s  fundamental  frequency)  is  attenuated by 82dB.

OttoWinter (approved):

PR #409, opened 2019-11-09
Add troubleshoot step.

fabianotessarolo:
Adds a solution for cases where Docker is used and the serial port does not work as expected.

## Description:

## Checklist:


OttoWinter (commented):
Hmm, I understand this is an issue for some docker users, however it's a little too much information to include in the getting started guide - it should be bare minimum.

Could you create an FAQ entry and link to that here?

fabianotessarolo:
Hey, Otto, yep I see, off course, I will close my PR and I write there.

PR #410, opened 2019-11-09
document the new has_position parameter for the template cover

ssieb:
## Description:
Add documentation for the new has_position parameter for the template cover.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#821

## Checklist:

### DIFF
--- a/components/cover/template.rst
+++ b/components/cover/template.rst
@@ -58,8 +58,9 @@ Configuration variables:
 - **assumed_state** (*Optional*, boolean): Whether the true state of the cover is not known.
   This will make the Home Assistant frontend show buttons for both OPEN and CLOSE actions, instead
   of hiding one of them. Defaults to ``false``.
-- **has_position** (*Optional*, boolean): Whether this cover will publish its position.  If true, you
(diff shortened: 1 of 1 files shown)


components/cover/template.rst
```diff
@@ -58,6 +58,8 @@ Configuration variables:
 - **assumed_state** (*Optional*, boolean): Whether the true state of the cover is not known.
   This will make the Home Assistant frontend show buttons for both OPEN and CLOSE actions, instead
   of hiding one of them. Defaults to ``false``.
+- **has_position** (*Optional*, boolean): Whether this cover will publish its position.  If true, you
```
OttoWinter:
```suggestion
- **has_position** (*Optional*, boolean): Whether this cover will publish its position as a floating point number.
  By default (``false``), the cover only publishes OPEN/CLOSED position. By setting this to true you can also send any position in between in ``lambda``.
```

The current text here doesn't really say what this value means. "position" is quite ambiguous here as it could refer to OPEN/CLOSED or, what you actually mean, percentage positions.

OttoWinter (approved):

PR #414, opened 2019-11-14
Update custom.rst

manju-rn:
As advised in #847 the component must also be returned hence the documentation update

## Description:

## Checklist:

### DIFF
--- a/components/sensor/custom.rst
+++ b/components/sensor/custom.rst
@@ -333,9 +333,10 @@ Our YAML configuration needs an update too:
       lambda: |-
         auto my_sensor = new MyCustomSensor();
         App.register_component(my_sensor);
-        return {my_sensor->temperature_sensor, my_sensor->pressure_sensor};
+        return {my_sensor, my_sensor->temperature_sensor, my_sensor->pressure_sensor};
 
       sensors:
+      - name: "My Custom Sensor Component"
       - name: "My Custom Temperature Sensor"
         unit_of_measurement: °C
         accuracy_decimals: 1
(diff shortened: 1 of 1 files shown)


components/sensor/custom.rst
```diff
@@ -333,23 +333,22 @@ Our YAML configuration needs an update too:
       lambda: |-
         auto my_sensor = new MyCustomSensor();
         App.register_component(my_sensor);
-        return {my_sensor->temperature_sensor, my_sensor->pressure_sensor};
+        return {my_sensor, my_sensor->temperature_sensor, my_sensor->pressure_sensor};
```
OttoWinter, line 336: This is only true for classes that extend `Component` **and** `Sensor`.

The class above only inherits from `Component`, and thus should not appear in this list (it will lead to a compile error if you put it there).

manju-rn: Agree. extended the Sensor to example MyCustomSensor

OttoWinter: No, that is not good either.

While having a class both *extend* from Sensor and have Sensor members does work, it's not good code and confusing. The updated docs you pushed now will make Home Assistant add an empty entity with no name, when that is not used.

manju-rn: I see your point.  How do you propose I resolve the situation what was mentioned in  #847 ?  I could not get a sensors to display in Home Assistant without passing the component.

OttoWinter: @manju-rn In that case, yes you should add it to the list.

However, the problem is that the way your code works is *not* recommended. As explained in the previous comment, having a Sensor object also have Sensor members is not good API design.

PR #416, opened 2019-11-19
Add instructions to password protect the dashboard

renehonig:
@vamichael, your long awaited password instructions...

## Description:

## Checklist:

### DIFF
--- a/guides/getting_started_command_line.rst
+++ b/guides/getting_started_command_line.rst
@@ -177,6 +177,28 @@ To start the ESPHome dashboard, simply start ESPHome with the following command
 
 After that, you will be able to access the dashboard through ``localhost:6052``.
 
+You can protect the dashboard with a password (which is a recommended security practice):
+
+.. code-block:: bash
+
+  # On Docker:
+  docker run --rm --net=host -v "${PWD}":/config -it esphome/esphome esphome . dashboard --password MY_PW
+  
+  # For docker-compose it is easier to use a .env file:
+    esphome:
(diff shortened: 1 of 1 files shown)


OttoWinter (commented):
1. This is only about docker - not about pip-based installs.
2. I'm kinda 50%-50% on this - I agree it's good to mention here, but it adds a lot of lines to the getting started guide that's supposed to be a dead-simple introduction. I think pushing the actual instructions into a FAQ entry and linking it here would be better.
3. While I understand the use of env files here, it unnecessarily complicates the setup - most docker compose examples use secrets defined in the YAML itself, so this does not really help anybody (those who know about env files will use it anyway, all others will not convert their config to use env files).
4. The note about the "private window" is confusing - it took me some time to understand what was meant there. This should rather be solved in software, with a logout button.

renehonig:
1. That true, we can add instructions for pip too.
2. In my opinion security typically does not get the attention it deserves. It is great that ESPHome has a number of security features build in (kudos!), but documentation was lacking. Skipping security for the sake of briefness is probably a bad idea. No problem having a link to faq. Btw, one reason for this to be long is that there currently is no docker-compose example, so maybe we should add that one earlier.
3. same here, secrets in the generic configurations files is typically a bad idea. This is how they get leaked in public repos. And the .env instructions couldn't be simpler.
4. Understand. When testing it looked like no credentials were required, so I wasted some time with that no realising authentication is 'cached', hence the warning.

PR #419, opened 2019-11-21
Update template.rst

KristopherMackowiak:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** https://github.com/esphome/esphome/pull/877

## Checklist:

### DIFF
--- a/components/cover/template.rst
+++ b/components/cover/template.rst
@@ -66,6 +66,11 @@ Configuration variables:
   tilt position.
 - **tilt_lambda** (*Optional*, :ref:`lambda <config-lambda>`):
   Lambda to be evaluated repeatedly to get the current tilt position of the cover.
+- **position_action** (*Optional*, :ref:`Action <config-action>`): The action that should
+  be performed when the remote (like Home Assistant's frontend) requests the cover be set to a specific
+  position.
+- **position_lambda** (*Optional*, :ref:`lambda <config-lambda>`):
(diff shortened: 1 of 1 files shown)


components/cover/template.rst
```diff
@@ -60,12 +60,16 @@ Configuration variables:
   of hiding one of them. Defaults to ``false``.
 - **has_position** (*Optional*, boolean): Whether this cover will publish its position as a floating point number.
   By default (``false``), the cover only publishes OPEN/CLOSED position.
-  By setting this to true you can also send any position in between using the publish action.
+  By setting this to true you can also send any position in between using the publish action. 
+  Parameter useless if you set the POSITION_ACTION (is set to TRUE).
```
OttoWinter, line 64: We should validate that in code - it's an easy validator to write. The docs don't need to be cluttered with this.

glmnet:
Fixed by https://github.com/esphome/esphome-docs/pull/544

components/cover/template.rst
glmnet: Well actually there is not much to validate, if sb sets `has_position` to false and also set `position_action` then it has position overriden to true, this is done in python side. But I don't know how easy will it be to validate since the default value of `has_position` is false, so we'd need to differentiate from default value and explicit false value just to give an error.

PR #425, opened 2019-11-30
added write section to setup addition internal amt90e32 registers

Diramu:
## Description:

New register access for atm90e32.   additional registers can be setup on startup.  And a interface to read write register from Home Assistant is described.

## Checklist:


components/sensor/atm90e32.rst
```diff
@@ -44,6 +44,7 @@ The `CircuitSetup 6-Channel Energy Monitor <https://circuitsetup.us/index.php/pr
     sensor:
       - platform: atm90e32
         cs_pin: 5
+	write:
```
OttoWinter: ? This is not valid YAML

Diramu: cutting error. fixed  
The write example was move to another section

PR #431, opened 2019-12-04
Add doc for color correction of rgb* lights

nickw444:
## Description:

I began implementing color correction for RGB lights in `esphome/esphome`, and thought it would be worth checking for previous feature requests for exactly this. 

Fortunately I managed to find [this comment](https://github.com/esphome/feature-requests/issues/463). I think it would be worthwhile adding this to the documentation for others like me who might need this in the future.

## Checklist:

### DIFF
--- a/components/light/rgb.rst
+++ b/components/light/rgb.rst
@@ -35,6 +35,32 @@ The ``rgb`` light platform creates an RGB light from 3 :ref:`float output compon
         pin: D1
       # Repeat for green and blue output
 
+Color Correction
+----------------
+
+It is often favourable to calibrate/correct the color produced by an LED strip light as the
+perceived intensity of different colors will generally vary. This can be done by using
+:ref:`max_power` on individual output channels:
+
+.. code-block:: yaml
+
+    # Example configuration entry
+    light:
+      - platform: rgb
+        name: "Living Room Lights"
+        red: output_component1
+        green: output_component2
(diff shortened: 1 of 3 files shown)


components/light/rgb.rst
```diff
@@ -35,6 +35,32 @@ The ``rgb`` light platform creates an RGB light from 3 :ref:`float output compon
         pin: D1
       # Repeat for green and blue output
 
+Color Correction
+----------------
+
+It is often favourable to calibrate/correct the color produced by an LED strip light as the
+perceived intensity of different colors will generally vary. This can be done by using
+:ref:`max_power` on individual output channels:
```
OttoWinter:
```suggestion
:ref:`max_power <config-output>` on individual output channels:
```

max_power is not a ref link, so it can't be linked to. Unfortunately, list items can't be tagged as link targets either, so the best we can do is link to the general output header, where max_power is located.

OttoWinter (approved):

OttoWinter:
Oops, there's one issue left: You're creating this PR against the next branch, but this is a doc update not connected to a new feature. So it should be filed against the `current` branch.

nickw444:
> Oops, there's one issue left: You're creating this PR against the next branch, but this is a doc update not connected to a new feature. So it should be filed against the `current` branch.

No worries! Just rebased 👍

PR #435, opened 2019-12-07
Use .timestamp instead of .time in time docs

davidjb:
## Description:

`.time` is deprecated since https://github.com/esphome/esphome/pull/531; this updates the reference in the documentation so that warnings like the following can be avoided:

```
src/main.cpp: In lambda function:
src/main.cpp:462:20: warning: 'esphome::time::ESPTime::<anonymous union>::time' is deprecated (declared at src/esphome/components/time/real_time_clock.h:35): .time is deprecated, use .timestamp instead [-Wdeprecated-declarations]
       return (time.time);
 ```

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#531

## Checklist:


OttoWinter (approved):
Thanks!

PR #443, opened 2019-12-15
Add example of converting dB to %

rlowens:
## Description:

## Checklist:

### DIFF
--- a/components/sensor/wifi_signal.rst
+++ b/components/sensor/wifi_signal.rst
@@ -37,6 +37,18 @@ Configuration variables:
     Signal strength readings are only available when WiFi is in station mode. Readings are not valid
     if the device is acting as an access point without any station mode connection.
 
+Example: Converting dB to %
+---------------------------
+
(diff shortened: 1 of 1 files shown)


components/sensor/wifi_signal.rst
```diff
@@ -37,6 +37,18 @@ Configuration variables:
     Signal strength readings are only available when WiFi is in station mode. Readings are not valid
     if the device is acting as an access point without any station mode connection.
 
+Example: Converting dB to %
+---------------------------
+
+The conversion from dB to % isn't well-defined (see disussion `"here" <https://www.adriangranados.com/blog/dbm-to-percent-conversion>`__) but the Arduino wire library and Tasmota both use this formula: RSSI_Percent = min(max(2 * (RSSI + 100.0), 0.0), 100.0) (that is, add 100 and multiply by 2, then cap at 100% max and 0% min), which truncates any dB lower than -100 or higher than -50. The ESP8266 goes as low as -120 and high as +4 in my experiance, but the very low and high values don't add much real information.
```
glmnet, line 43:
```suggestion
The conversion from dB to % isn't well-defined (see disussion `"here" <https://www.adriangranados.com/blog/dbm-to-percent-conversion>`__) but the Arduino wire library and Tasmota both use this formula: ``RSSI_Percent = min(max(2 * (RSSI + 100.0), 0.0), 100.0)`` (that is, add 100 and multiply by 2, then cap at 100% max and 0% min), which truncates any dB lower than -100 or higher than -50. The ESP8266 goes as low as -120 and high as +4 in my experience, but the very low and high values don't add much real information.
```

PR #457, opened 2019-12-31
Fix SGP30 incorrect baseline reading/writing

panuruj:
## Description:
The SGP30 is the eCO2 and TVOC sensor needs calibration with the eCO2 and TVOC baseline. The current implementation only reads the LSB of the 2-byte value of each of the eCO2 and TVOC baselines. This ends up in an incorrect baseline reading as well as writing at the initialization.

Based on the [SGP30 datasheet](https://www.mouser.com/datasheet/2/682/Sensirion_Gas_Sensors_SGP30_Datasheet_EN-1148053.pdf)
> The SGP30 also provides the possibility to read and write the baseline values of the baseline correction algorithm. This feature is used to save the baseline in regular intervals on an external non-volatile memory and restore it after a new power-up or soft reset of the sensor. The command “Get_baseline” returns the baseline values for the two air quality signals. The sensor responds with 2 data bytes (MSB first) and 1 CRC byte for each of the two values in the order CO2eq and TVOC. These two values should be stored on an external memory. After a power-up or soft reset, the baseline of the baseline correction algorithm can be restored by sending first an “Init_air_quality” command followed by a “Set_baseline” command with the two baseline values as parameters in the order as (TVOC, CO2eq). 

**Related issue (if applicable):** fixes https://github.com/esphome/issues/issues/950

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#936

## Checklist:

  - [ x ] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [ x ] Link added in `/index.rst` when creating new documents for new components or cookbook.


panuruj:
I'll rebase first.

PR #459, opened 2020-01-03
web_server - css_include and js_include: add new options and example

Elkropac:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#932

## Checklist:


glmnet:
This is nice, can you make it more obvious please? I had to check the code to figure out exactly what was going on.
I suggest to add statements like "The file contents will be inlined in the rendered page between  `<script type='text-javascript'>` tags.

Likewise, "The file contents will be inlined in the rendered page header between `<style>` tags"

PR #474, opened 2020-01-19
Update automations.rst

gordonthree:
Added a tip for calling script_execute from within a lambda.

## Description:

## Checklist:


glmnet:
Is this checked? shouldn't it be `execute()` with `()`

PR #475, opened 2020-01-20
add a note about the log level of dump: all

drewp:
It would have saved me a bunch of time just now if I knew that remote_receiver's dump=all logging is at level DEBUG.


glmnet (approved):

glmnet:
Thanks

PR #484, opened 2020-01-29
Update board for Sonoff and link to Tasmota

sillyfrog:
## Description:
Update the recommended `board` and and adds the `arduino_version` to the generic Sonoff setup.

Also updated links to Tasmota for the Sonoff pinouts.

**Related issue (if applicable):** This addresses the issues as outlined here: https://github.com/esphome/issues/issues/931

## Checklist:


brandond:
The only difference between the two boards is a flash mode of DOUT vs QIO, which I guess makes sense.

PR #486, opened 2020-02-06
Add a warning for missing name

micw:
## Description:

Minor change to docs: In MQTT warn about that unnamed (internal) components don't work with mqtt.

Reason: took me several hours to figure it out (including several hours of chat in gitter). The log says the unnamed component has topics bit it just ignores any mqtt message and does not send any.

## Checklist:

### DIFF
--- a/components/mqtt.rst
+++ b/components/mqtt.rst
@@ -249,6 +249,10 @@ MQTT Component Base Configuration
 All components in ESPHome that do some sort of communication through
 MQTT can have some overrides for specific options.
 
+.. warning::
+
+    Components without **name** are internal only, thus they will not send or receive mqtt messages.
+
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


glmnet:
Ahh yes, the same applies to `api` components without name are treated as `internal` too, I wonder how can this be more obvious for newcomers 

@OttoWinter 
I'd like to make the `name` optional only if `internal` is set to `true`
Otherwise a less impact will be to write a warning on the validator

brandond:
As @glmnet mentioned, unnamed components are implicitly internal, which affects mqtt, api, and web access. This is covered in the docs, although you have to look for it:

```internal (Optional, boolean): Mark this component as internal. Internal components will not be exposed to the frontend (like Home Assistant). Only specifying an id without a name will implicitly set this to true.```

PR #513, opened 2020-03-19
Add config types to homepage guides list

pauln:
## Description:
Adds a link to the Config Types page to the Guides section of the homepage.  This is a useful page, but is hard to find.

I've used the existing "puzzle" icon, as config types are an important piece of the puzzle (and I couldn't think of an alternative which would fit better).

## Checklist:

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -40,6 +40,8 @@ Guides
     Supporters, guides/supporters, heart.svg
     DIY Examples, guides/diy, earth.svg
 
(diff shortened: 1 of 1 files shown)


glmnet:
Thanks for this attempt, however I believe this won't help much.

I better home page for ESPHome is needed indeed. I'm gonna leave this one open to see if have any feedback from other people.

PR #514, opened 2020-03-22
pzemac energy support

### LINKED ISSUE #457: Add pzem004t total energy value (Energy field)

It would be great (after pzem004t support arrived :rocket:) to also have support for the **total energy value** which is stored in the pzem004t and doesn't reset with a power cut.

This is a super useful value. It's included in all versions of the pzem004t (the old v1+2 and the new v3).

I use it to 'sync' my pzem004t to my mains energy meter :man_scientist: 

I tried to get the same information with the integration component. Problem is that I have my `update_interval:` set to `1s` for the pzem004t component to get real time information. When I add a `restore: true` to the `integration` component I'm worried it will wear out the flash from the esp really quick because it will write the value every second to the flash of the esp (if I understand the docs right).In general I also want avoid to store things on the esp flash.

Another good thing with the value stored inside the pzem004t is that a malfunction of the esp (hardware defect for example) or a communication failure (uart or wifi for example) can't stop it. The pzem004t will continue to measure power usage and stores it - no matter of the other failures. If things are fixed again the total energy consumption reported from the pzem004t will be the correct one. :confetti_ball: 

**Additional context**

https://github.com/esphome/feature-requests/issues/49

rradar:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/457

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** https://github.com/esphome/esphome/pull/933

## Checklist:

### DIFF
--- a/components/sensor/pzemac.rst
+++ b/components/sensor/pzemac.rst
@@ -44,10 +44,12 @@ to some pins on your board and the baud rate set to 9600.
       - platform: pzemac
         current:
           name: "PZEM-004T V3 Current"
-        voltage:
-          name: "PZEM-004T V3 Voltage"
         power:
           name: "PZEM-004T V3 Power"
+        voltage:
+          name: "PZEM-004T V3 Voltage"
+        energy:
+          name: "PZEM-004T V3 Energy"
         frequency:
           name: "PZEM-004T V3 Frequency"
         power_factor:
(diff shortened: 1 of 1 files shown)


glmnet:
Reverted as this is already in next where it should have been originally.

sendorm:
What do you mean this is already in? Can we use energy or not?

glmnet:
yes, in ESPHome dev branch

sendorm:
Is there a plan on putting this on master branch?

glmnet:
https://github.com/esphome/feature-requests/issues/658

sendorm:
I still don't get the reasoning of reverting this from master. Not everybody likes to use the dev branch. If one day the master does get an update do the reversion then.

glmnet:
The documentation was added to master by mistake. ESPHome master current version is 1.14.4 and dev is bleeding edge code.
Docs are in sync:
* Docs next https://next.esphome.io -> code dev 
* Docs current https://esphome.io -> code master (v1.14.4 now)

So the commit which enables energy is this one https://github.com/esphome/esphome/pull/933 which is already merged in dev but yet not merged in master (talking about code) so once we do a release current and master gets updated simultaneously.

sendorm:
Ok thank you.

PR #520, opened 2020-03-24
Document two missing hmc5883l config settings

balrog-kun:
There are two HMC5883L settings: ``oversampling`` and ``range``, that are supported by the esphome sensor but were missing from the docs.


glmnet:
Thanks!

PR #531, opened 2020-04-02
remove link to Adafruit

ebaschiera:
the link is not pointing at the right product page, and Adafruit looks like is not selling (never sold?) this device

## Description:

## Checklist:


glmnet:
Is the mkelec link working???

ebaschiera:
> Is the mkelec link working???

No, it doesn't. I removed that too! 👍

PR #538, opened 2020-04-10
Add support for command-line-substitutions

AlexMekkering:
## Description:
This adds support for command line substitutions. These can be provided by --substitution|-s <key> <value> items which are optional can be provided more then once.
The purpose is to add or override substitutions from the command line which allows for more generic 'template' configuration files which can be used for multiple devices of the same kind which can be deployed with some device-specific substitutions (like e.q. $devicename).
**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1014

## Checklist:

### DIFF
--- a/guides/configuration-types.rst
+++ b/guides/configuration-types.rst
@@ -173,11 +173,52 @@ validating your configuration, ESPHome will automatically replace all occurrence
 by their value. The syntax for a substitution is based on bash and is case-sensitive: ``$substitution_key`` or
 ``${substitution_key}`` (same).
 
-You can also add or override substitutions from the command line by adding e.g. ``-s devicename mydevice``
-which overrides the ``devicename`` substitution and gives it value ``mydevice``.
-Command line substitutions take precedence over the ones in your configuration file.
-This can also be used to create more generic 'template' configuration files which can be used for multiple devices,
-based on substitutions which are provided on the command line.
+You can also define or override substitutions from the command line by adding e.g. ``-s devicename mydevice``
+which overrides the ``devicename`` substitution and gives it value ``mydevice``. This can be issued multiple times,
+so e.g. with the following ``example.yaml`` file:
+
+.. code-block:: yaml
+
+    substitutions:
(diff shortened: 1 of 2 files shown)


guides/configuration-types.rst
```diff
@@ -173,6 +173,12 @@ validating your configuration, ESPHome will automatically replace all occurrence
 by their value. The syntax for a substitution is based on bash and is case-sensitive: ``$substitution_key`` or
 ``${substitution_key}`` (same).
 
+You can also add or override substitutions from the command line by adding e.g. ``-s devicename mydevice``
+which overrides the ``devicename`` substitution and gives it value ``mydevice``.
+Command line substitutions take precedence over the ones in your configuration file.
+This can also be used to create more generic 'template' configuration files which can be used for multiple devices,
+based on substitutions which are provided on the command line.
+
```
glmnet: Can you add an example of many substitutions if they are possible? I believe they are based on what I understood in the Python.

AlexMekkering: Thanks @glmnet, please see my changes.

glmnet:
Reviewing this I just found out there is a doc page for the command line interface, did you see that before? 

https://esphome.io/guides/cli.html

Just found it by looking for your changes, which are not easy to find either in "configuration types" Somebody sent a PR to add a link to configuration types into the main page but that main page is full of icons and I really get lost too.

The thing is I entered command line in the search box and ended up with this.

Anyway I believe this should be added to the cli page too, may be you can link from one to the other page for this nice sample.

Thanks

glmnet:
Ahh theres a link here https://esphome.io/guides/faq.html 

I really believe we need to improve this... anyway, im not good a docs

AlexMekkering:
@glmnet  I've re-structured it somewhat and added the global cli options on the cli page with a reference to the cli substitutions, so please check.

glmnet:
This is great!! Thank you!

PR #542, opened 2020-04-14
add support for state array

glmnet:
## Description:

Modifies served javascript for ESPHome so the state event handler can process array of events.

**Related issue (if applicable):** https://github.com/esphome/issues/issues/806

## Checklist:


glmnet:
fixed via https://github.com/esphome/esphome/pull/1221

PR #553, opened 2020-04-23
specific MacOS Docker command to launch dashboard

oncleben31:
## Description:

Update documentation to explain why the standard Docker command to launch the dashboard is not working on MacOS and give an alternative command.

## Checklist:

### DIFF
--- a/guides/faq.rst
+++ b/guides/faq.rst
@@ -261,8 +261,12 @@ Command reference:
     # Map /dev/ttyUSB0 into container
     docker run --rm -v "${PWD}":/config --device=/dev/ttyUSB0 -it esphome/esphome ...
 
-    # Start dashboard on port 6052
+    # Start dashboard on port 6052 (general command)
+    # Warning: this command is currently not working with Docker on MacOS. (see note below)
     docker run --rm -v "${PWD}":/config --net=host -it esphome/esphome
+
+    # Start dashboard on port 5062 (MacOS specific command)
+    docker run --rm -p 6052:6052 -e ESPHOME_DASHBOARD_USE_PING=true -v "${PWD}":/config -it esphome/esphome
 
     # Setup a bash alias:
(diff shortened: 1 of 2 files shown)


MLKiiwy:
👍 very usefull indeed.

Also will be nice to add the corresponding docker-compose config

```yml
version: "3"

services:
  esphome:
    environment:
      - ESPHOME_DASHBOARD_USE_PING=true
    image: esphome/esphome
    volumes:
      - ./:/config:rw
    ports:
      - 6052:6052
    restart: always
```

oncleben31:
@MLKiiwy I'm not sure to understand why it's needed and where to put this info ?
I'm not a docker specialist can you explain ?

MLKiiwy:
it's just a bonus, not mandatory.

Lot of people use docker-compose files to avoid having aliases for docker commands. 

Here they put an example for esphome that does not work: https://github.com/esphome/esphome-docs/blob/a99fe66be8da081d52b1cd6560295ae5b5e44a99/guides/faq.rst#docker-reference

Line 298 you can just add:

```
And the corresponding docker-compose file looks like this:

```yml
version: "3"

services:
  esphome:
    environment:
      - ESPHOME_DASHBOARD_USE_PING=true
    image: esphome/esphome
    volumes:
      - ./:/config:rw
    ports:
      - 6052:6052
    restart: always

```

```

glmnet:
I believe these commands are for MacOS indeed, I believe Otto uses Apple, he definitely doesn't use Windows so last option will be Linux, but unlikely.

From: "doesn't work as expected" there is not much to help anybody

oncleben31:
@OttoWinter Any reason explaining why this PR is not merged or commented ?

OttoWinter:
Hmm, ok yeah weird. I never had that issue with macOS back when I used it (switched to Kubuntu some time ago, and don't have a mac system to test on).

Thanks! I'll merge this.

PR #561, opened 2020-04-27
Fix typos

mje-nz:
## Description:
Fix typos

## Checklist:


glmnet:
Thanks!

PR #581, opened 2020-05-07
Update gpio.rst

3ative:
Added "Example 2 (Use "on_turn_on" instead of template) configuration entry" to "Momentary Switch" section

## Description:
By using **on_turn_on** for a _Momentary Switch_, there's no need for a **template** in the switch section/code.
- I have added a 2nd example.

## Checklist:

### DIFF
--- a/components/switch/gpio.rst
+++ b/components/switch/gpio.rst
@@ -83,6 +83,19 @@ or closes the gate. The relay simulates the button press for 500ms.
         - switch.turn_on: relay
         - delay: 500ms
         - switch.turn_off: relay
+.. code-block:: yaml
+
+    # Example 2 (Use "on_turn_on" instead of template) configuration entry
+    switch:
+      - platform: gpio
+        pin: 25
+        id: relay
+        name: "Gate Remote"
+        icon: "mdi:gate"
+        on_turn_on:
+          - switch.turn_on: relay
+          - delay: 500ms
+          - switch.turn_off: relay
 
 .. figure:: images/gate-remote-ui.png
     :align: center


glmnet:
A bit hacky... is the `- switch.turn_on` really needed?

glmnet:
yet another way to do it, the way I do it LOL

````yaml

switch:
  - platform: output
    name: "Porton boton"
    id: porton_boton
    output: gpio_relay_1
    on_turn_on:
      then:
        - delay: 1s
        - switch.turn_off: porton_boton
````

**Comment by frenck** [comment]
@glmnet I like that suggestion as it is the closest to the source that handles it.

Should we update this PR? Or do you want to open an alternative PR?

I'm happy to jump in as well?

**Comment by 3ative** [comment]
@frenck 
How is @glmnet 's suggestion any different from my original? 
- Why don't you like mine ?

**Comment by frenck** [comment]
@3ative I didn't state I didn't like yours.

I stated I think the suggestion from @glmnet is a general better one:

>  I like that suggestion as it is the closest to the source that handles it.

Hence I shared that here.

**Comment by 3ative** [comment]
@frenck 
Both solutions are valid. Except mine follows the existing example structure, as laid out in the docs and eliminates the need for a 'list' and a 'template' 
My OP was to add an additional 'Example' and not re-write / replace.

**Comment by frenck** [comment]
> Both solutions are valid

I didn't state that yours wasn't valid either. Please stop guessing/assuming things.

I just stated I think that the example provided by @glmnet, from general code maintenance and the perspective on where one should put responsibilities and how one should separate concerns: his example, in general, is a good one.

**Comment by 3ative** [comment]
@frenck 
I wasn't suggesting, guessing or assuming your intent, I was merely stating I liked @glmnet example.
Just that it doesn't follow existing Doc example(s) - and therefore is an different approach. Which should therefore be an additional 'example' for another Docs section.

**Comment by frenck** [comment]
Yes, so lets do that... (and now we are back at my original response in this PR 🤷 )

**Comment by 3ative** [comment]
@frenck 
Ok, cool let's indeed have both, mine as stated: _"Example 2 (Use "on_turn_on" instead of template) configuration entry"_ in the _"To create momentary switches,..."_ header and @glmnet 's maybe in a more suitable section of the docs.

File: components/switch/gpio.rst [hunk 1]
```diff
@@ -83,6 +83,19 @@ or closes the gate. The relay simulates the button press for 500ms.
...
+
+    # Example 2 (Use "on_turn_on" instead of template) configuration entry
+    switch:
+      - platform: gpio
+        pin: 25
+        id: relay
+        name: "Gate Remote"
+        icon: "mdi:gate"
+        on_turn_on:
+          - switch.turn_on: relay
```
frenck, line 96: This is the `on_turn_on`... you want to turn it on again?

```suggestion
```

frenck (changes requested):

3ative:
> Should we update this PR? ...

I wasn't aware that a PR could be changed or 'updated' - if that is indeed the case then, I'm happy with that. 👍
I see what you mean. Indeed the '- switch.turn_on: relay' is not needed. 

Thanks, when I posted this PR I didn't know as much as I do now.

frenck:
@3ative Could you remove that unneeded line?

You haven't opened up the PR for maintainers to contribute to, so can't help you out on this one.

3ative:
@frenck 
I'm unable to find the option to edit and remove the unwanted line... A little help? 
- How do I make the new changes... or open for maintainers. Cheers.

frenck:
Oh, I see, you've removed the upstream fork. This means this PR can't go anywhere at this point anymore and can only be solved by opening a new PR.

Closing this PR on this end for now.

../Frenck

3ative:
@frenck 
Thanks for the support and helping me understand how use PR correctly.

PR #584, opened 2020-05-09
Add H801 PIR sensor to cookbook

RichFoley:
## Description:
I've added instructions to use the GPIO pins of the H801 as an input. I think this is useful to explicitly state since having a motion sensor attached to this device makes it much more useful for me. This is my first contribution to this repository, so I apologize for any missteps and appreciate feedback.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** n/a

## Checklist:


glmnet:
I was wondering if this is working for you without disabling the logger serial output?

i.e.

````yaml
logger:
  baud_rate: 0
````

because I don't see you explaining that too, and since this is a cookbook it might be worth mentioning it too.

RichFoley:
I did not need to disable the logger for the pin in the example, but that might be required for the other pin. TX/RX are labelled backwards on this board so I was using RX (on the ESP, the one that's labelled TX). That may explain why there was not conflict.

I should be receiving another of these boards soon that I can test this on.

PR #585, opened 2020-05-09
Max7219 new documentation

rspaargaren:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/14#

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1053

## Checklist:


rspaargaren:
@glmnet index.rst has been added to the documentation

components/display/max7219digit.rst
```diff
@@ -0,0 +1,181 @@
...
+Some special MAX7219 digit code can be added as follows:
+
+.. code-block:: yaml
+
+    display:
+      - platform: max7219digit
+        # ...
+        lambda: |-
+          // Print 0 at position 0 (left)
+          it.print(0,0, id(digit_font), "Hello!");
```
glmnet: Do you mean the other print here? The one with the embedded font?

components/display/max7219digit.rst
```diff
@@ -0,0 +1,181 @@
...
+.. code-block:: yaml
+
+    display:
+      - platform: max7219digit
+        # ...
+        lambda: |-
+          // Print Hello at position 0 (left)
+          it.print(0,0, id(digit_font), "Hello!");
+          it.invert_on_off(true);
+          it.line(0,0,32,8);
```
glmnet: This seems going off buffer

Also example said it will clear the pixels. But when I run it, it inverted the pixels. 

Also line functions and other graphics accept a 5th parameter which be default is COLOR_ON. But you can pass color off so you actually clear pixels this way.

glmnet (changes requested):
Also add some small headings. I don’t understand if the sample code is for the text above or below.

components/display/max7219digit.rst
rspaargaren: I have removed this part as it is already explained in the code block below and does not clarify anything

components/display/max7219digit.rst
rspaargaren: code has been changed, the output of the example is quite ugly but works.

rspaargaren:
I just realized that the last part of the documentation in relation to the invert function needs to be changed to correspond with the latest revision. I will update that part.

glmnet:
@rspaargaren I have updated a bit the docs, the latest example is not working, it says:

> This code will only affect the line drawn on the screen. The line will wipe the pixels from top left to right bottom.
The background is not affected as the Lambda is closed with an ``invert_on_of(false)`` code.

However in my setup it draws the line in "on" pixels across the screen, and given the end limits are 32, 8 it does also scroll due to the line.

If I use this code:

```yaml
      // Print Hello at position 0 (left)
      it.print(0,0, id(digit_font), "Hello!");
      it.line(0, 0, 31, 7, COLOR_OFF);
```

Then it does what I understand from that description.

Please load my formatting changes and update docs or the code to make it work as you want.

glmnet (changes requested):
See my new comments

PR #590, opened 2020-05-12
Update for multiple receiver document

Eenoo:
Clarity on multiple receivers.

## Description:

## Checklist:

### DIFF
--- a/components/remote_receiver.rst
+++ b/components/remote_receiver.rst
@@ -22,6 +22,15 @@ which will trigger when they hear their own configured signal.
     remote_receiver:
       pin: GPIO32
       dump: all
+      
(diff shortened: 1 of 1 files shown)


Eenoo:
Documentation update for https://github.com/esphome/feature-requests/issues/679

I had this issue as well.

glmnet:
Thanks but we try to keep example configuration as simple as possible, if we were to show all features in the first example it will be always too complicated, e.g. showing tolerance, and extended version of the `pin:` schema, etc.

The key here is the latest configuration option, it states it clearly: **id** ... Use this if you have multiple remote receivers.

Then if this is the first time you are writing yaml it will bite you, yes, but it's the learning curve I guess.

PR #594, opened 2020-05-15
Add documentation for ST7789V display module (as on TTGO T-Display)

kbx81:
## Description:

**Related issue (if applicable):** fixes/updates/replaces #918 

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1050

## Checklist:


pulimento:
Very nice doc! Thank you! I think it's OK

deftdawg:
@kbx81 

The example from the doc has an error... `RGB565` is not a valid image format, so the example cannot be built.

There is code for that in a ttgo custom_component which is likely where it originated from, however that code was never merged into ESPHome.

PR #500, opened 2020-02-22
Update docs to reflect ssd1327 support

igg:
## Description:

**Related issue (if applicable):** esp1327 support esphome/feature-requests#608

**Pull request in [esphome](https://github.com/esphome/esphome):**

PR #596, opened 2020-05-16
SSD1325 documentation update for grayscale support

kbx81:
## Description:

Documents the patch that adds grayscale support to the SSD1325 display component.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1064

## Checklist:

### DIFF
--- a/components/display/ssd1325.rst
+++ b/components/display/ssd1325.rst
@@ -53,6 +53,7 @@ Configuration Variables
   - ``SSD1325 128x64``
   - ``SSD1325 96x16``
   - ``SSD1325 64x48``
+  - ``SSD1327 128x128`` **# Note the number seven!**
 
 - **reset_pin** (:ref:`Pin Schema <config-pin_schema>`): The RESET pin.
 - **cs_pin** (*Optional*, :ref:`Pin Schema <config-pin_schema>`): The pin on the ESP that that the CS line is connected to.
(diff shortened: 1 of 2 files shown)


kbx81:
@glmnet should #500 have been merged into `next` instead of `current`?

components/display/ssd1325.rst
```diff
@@ -60,6 +65,68 @@ Configuration variables:
 - **pages** (*Optional*, list): Show pages instead of a single lambda. See :ref:`display-pages`.
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
 
+Configuration examples
+**********************
+
+As of version 1.15, ESPHome supports grayscale (and color) displays. To utilize the grayscale capabilities
+of this display module, you'll likely want to add a ``color:`` section to your YAML configuration; please see
+:ref:`color <config-color>` for more detail on this configuration section. As this is a grayscale display, it
+only uses the white color element (as opposed to red, green, and/or blue).
```
glmnet:
```suggestion
To utilize the grayscale capabilities of this display add a ``color:`` section to your YAML configuration; see
:ref:`color <config-color>` for more details. As this is a grayscale display, it only uses the white color element.
```

I don't see why we need to explain this is a new feature, docs should be about current version, unless we broke something may be, just may be

glmnet: you're welcome to update the changelog to mention this

kbx81: Thanks @glmnet! I made this change and added another code example, as well.

PR #604, opened 2020-05-17
Remove Count Water Bottles by vigonotion example from DIY examples page

JeffResc:
## Description:
Remove the "Count Water Bottles" by vigonotion example from DIY examples page. The existing URL returns a 404 and after doing numerous Google searches, I could not find the new URL to this project's page. If anyone has the new URL, I would like to replace it with that, but unfortunately, I could not find it so I've removed the URL.

## Checklist:


vigonotion:
That link should work, at least it does for me. I know my Google results are broken, but you should absolutely be able to access this link. Can you try accessing the main page https://vigonotion.com/ ?

JeffResc:
@vigonotion I believe there is something wrong with the DNS, SSL or Virtual Server settings (I can’t specifically tell). When connecting, I get an SSL error stating that the certificate name is invalid (the certificate is for linealisol.gtksoftware.com, not vigonotion.com). Then, once I accept this warning, it redirects me to their site as well, and not yours.

Let me know if that information helps :)

vigonotion:
Yes you're right. Somehow my A record got messed up, and because I have an IPv6 connection, it uses the (correct) AAAA record. That's why I was seeing my page but you didn't. Thank you so much for finding this! I would probably never have noticed this!

The wrong result is probably in your cache for some time, maybe you could try checking again tomorrow?

JeffResc:
I cleared my cache manually and all looks good. Glad I was able to point this out to you, going to close out this pull request now…

PR #613, opened 2020-05-20
MCP23S08-Docs

SenexCrenshaw:
## Description:
Docs for MCP23SXX I/O Expander - SPI

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1068

## Checklist:


components/mcp23Sxx.rst
```diff
@@ -0,0 +1,129 @@
...
+    # Individual outputs
+    switch:
+      - platform: gpio
+        name: "MCP23S08 Pin #0"
+        pin:
+          mcp23S08: mcp23S08_hub
+          # Use pin number 0
+          number: 0
+          # One of INPUT, INPUT_PULLUP or OUTPUT
+          mode: INPUT_PULLUP
```
glmnet: INPUT?

jesserockz:
```suggestion
          mode: OUTPUT
```

components/mcp23Sxx.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. _mcp23S17-label:
+
+MCP23S17
+--------
+
+The MCP23S17 component allows you to use MCP23S17 I/O expanders
+(`datasheet <http://ww1.microchip.com/downloads/en/DeviceDoc/20001952C.pdf>`__,
+`Digi-Key <https://www.digikey.com/product-detail/en/microchip-technology/MCP23S17-E-SP/MCP23S17-E-SP-ND/894276>`__) in ESPHome.
+It uses the :ref:`SPI Bus <i2c>` for communication.
```
jesserockz:
```suggestion
It uses the :ref:`SPI Bus <spi>` for communication.
```

jesserockz (approved):

alexyao2015:
@SenexCrenshaw I think this would look better if it was all merged under one documentation page. I know they are split under two different codebases, but it makes sense to keep documentation together because it is the same device. Maybe reformat it so it looks like this? https://esphome.io/components/display/ssd1306.html?highlight=spi

alexyao2015:
Also the documentation here says that it's an MCP23Sxxx but has a picture of MCP230xxx which is confusing. Is it a different device altogether? If that's the case, remove the whole spi nomenclature?

PR #620, opened 2020-05-24
Fix broken YAML in MQTT on_json_message example

napieraj:
## Description:
The example YAML for the `on_json_message` trigger of the MQTT component [(mqtt.html#on-json-message-trigger)](https://esphome.io/components/mqtt.html#on-json-message-trigger) attempts to map `then:` to the topic - this doesn't work and fails validation:

```
ERROR Error while reading config: Invalid YAML syntax:

mapping values are not allowed here
  in "/config/esphome/esp32_devkit.yaml", line 32, column 11:
          then:
              ^
```

**Related issue (if applicable):** fixes

## Checklist:

### DIFF
--- a/components/mqtt.rst
+++ b/components/mqtt.rst
@@ -362,7 +362,7 @@ Please note that it's a good idea to check if the key exists in the Json Object 
       # ...
       on_json_message:
         topic: the/topic
-          then:
+        then:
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Thanks

PR #626, opened 2020-05-28
Not every nodeMCU board supports 12V

Laess3r:
## Description:
It could destroy the board if users just try and power it using 12V. There are boards that don't support 12V but only up to 5V.


rradar (approved):

rradar:
Better save than sorry @Laess3r :wink:

PR #628, opened 2020-05-28
Added Etekcity Voltson cookbook

gitolicious:
## Description:
Configure an Etekcity Voltson and calculate the missing Current sensor

## Checklist:


glmnet:
Thanks!

PR #634, opened 2020-06-05
Revert "Update docs to reflect ssd1327 support"

glmnet:
Reverts esphome/esphome-docs#500 due to wrong base branch


glmnet:
@kbx81 thanks for pointing the base branch mistake, we still need this docs in next, would you like to push on your PR?
cc @igg

PR #654, opened 2020-06-20
LG Climate IR

square99:
## Description:
 Add LG Climate IR.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1097

## Checklist:

### DIFF
--- a/components/climate/climate_ir_lg.rst
+++ b/components/climate/climate_ir_lg.rst
@@ -1,0 +1,81 @@
+LG IR Remote Climate
+========================
+
+.. seo::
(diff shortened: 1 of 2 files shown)


glmnet:
This Pull Request should be created against next branch, not current.

For climates based on IR Climate, docs are much simpler now, it's just a new row on a table and maybe some specific information if the devices requires it.

See https://next.esphome.io/components/climate/ir_climate.html

square99:
I'll re-check

PR #667, opened 2020-07-02
Update bang_bang doc, changelog fix

kbx81:
## Description:

Adds a note about the `thermostat` component to the top of the `bang_bang` component's document in an effort to mitigate confusion around the use and behavior of the component.

**Related issue (if applicable):** esphome/esphome#1061 esphome/esphome#1105

## Checklist:

  - [ ] Link added in `/index.rst` when creating new documents for new components or cookbook. **[N/A]**

### DIFF
--- a/changelog/v1.15.0.rst
+++ b/changelog/v1.15.0.rst
@@ -53,7 +53,7 @@ Notable Changes & New Features
 
 - Dashboard Interface Revamped
 - New Thermostat Controller implements ESPHome actions for all available Home Assistant climate actions,
-  climate modes, fan modes, and fan swing modes (#1061)
+  climate modes, fan modes, and fan swing modes (#1105)
 - Color (and grayscale) display support! (#1050)
 
 All changes
--- a/changelog/v1.15.0.rst
+++ b/changelog/v1.15.0.rst
@@ -53,7 +53,7 @@ Notable Changes & New Features
 
 - Dashboard Interface Revamped
 - New Thermostat Controller implements ESPHome actions for all available Home Assistant climate actions,
-  climate modes, fan modes, and fan swing modes (#1105)
+  climate modes, fan modes, and fan swing modes (:esphomepr:`1105`)
 - Color (and grayscale) display support! (#1050)
 
 All changes
--- a/components/climate/bang_bang.rst
+++ b/components/climate/bang_bang.rst
@@ -144,8 +144,8 @@ hysteresis window is normal; so, while it may seem unlikely, this situation is a
 The :doc:`thermostat <thermostat>` component differs in that there is hysteresis around *each* set point. For example,
 if the ``target_temperature_low`` set point is 20 °C, and the (default) hysteresis value of 0.5 °C is used,
 ``heat_action`` is called at a temperature of 19.5 °C and ``idle_action`` is called at 20.5 °C. If cooling, as defined
-by ``target_temperature_high``, is set to 22 °C, ``cool_action`` is called at 21.5 °C and ``idle_action`` would be
-called at 22.5 °C. Again, it is essentially two bang-bang controllers in one.
+by ``target_temperature_high``, is set to 22 °C, ``cool_action`` would be called at 22.5 °C and ``idle_action`` called
+at 21.5 °C. Again, it is essentially two bang-bang controllers in one.
 
 Great. But what if you do not have a dual-function system?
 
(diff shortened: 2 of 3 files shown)


kbx81:
@igg would love your feedback on these changes. 😄

glmnet:
Me too

igg:
Sorry for the wait.
I think the sentence "as the temperature is essentially guaranteed to always be between the two set points" should go - this is only true for a heater and a cooler controlled by a single bang_bang doing the ping-pong oscillation. A single-function system is expected to heat (or cool) for some time past it's turn-off point.

Also this sentence is related to the same property:
"Some oscillation at the "edges" of the hysteresis window is normal; so, while it may seem unlikely, this situation is actually quite common."
This could maybe be a stronger point, because in real systems it's pretty much guaranteed to happen. Maybe something like:
"Oscillation at the "edges" of the hysteresis window, or going past the set-point in either direction is expected: consider a heater that gets turned off by it's set-point. The heating element will still hot, and will continue to heat the air until the element and the air reach the same temperature."

I think the rest is great!

kbx81:
Sweet, thanks @igg! These are great points; I'll make these changes. 😄🍻

igg:
@dseven, can you chime in too and see if this new description makes better sense than before for when to use `thermostat` and when to use `bang_bang`? Basically, would this have led you on the right path?
This link is a reasonable preview, unless someone has a better way:
https://github.com/kbx81/esphome-docs/blob/update-bang-bang-doc/components/climate/bang_bang.rst

kbx81:
Netlify preview is here: https://deploy-preview-667--esphome.netlify.app/components/climate/bang_bang.html#bang-bang-vs-thermostat
...it's much closer to the Real Thing™️. 😄

dseven:
Sorry for the delay. I just got to take a look at the doc update. It does make sense. It's been a while since I was in the utterly confused state, so I forget the details. but what the new doc describes seems plausible as what got me there :)

changelog/v1.15.0.rst
```diff
@@ -53,7 +53,7 @@ Notable Changes & New Features
 
 - Dashboard Interface Revamped
 - New Thermostat Controller implements ESPHome actions for all available Home Assistant climate actions,
-  climate modes, fan modes, and fan swing modes (#1061)
+  climate modes, fan modes, and fan swing modes (#1105)
```
OttoWinter:
```suggestion
  climate modes, fan modes, and fan swing modes (:esphomepr:`1105`)
```

That automatically creates a link

components/climate/bang_bang.rst
```diff
@@ -90,6 +97,87 @@ Advanced options:
...
+that is turned off after it reaches its set point. The heating element will remain hot (potentially for quite a while),
+and as such will continue to heat the air until the element fully cools down to match the ambient air/room temperature.
+
+The :doc:`thermostat <thermostat>` component differs in that there is hysteresis around *each* set point. For example,
+if the ``target_temperature_low`` set point is 20 °C, and the (default) hysteresis value of 0.5 °C is used,
+``heat_action`` is called at a temperature of 19.5 °C and ``idle_action`` is called at 20.5 °C. If cooling, as defined
+by ``target_temperature_high``, is set to 22 °C, ``cool_action`` would be called at 22.5 °C and ``idle_action`` called
+at 21.5 °C. Again, it is essentially two bang-bang controllers in one.
+
+Great. But what if you do not have a dual-function system?
```
OttoWinter:
```suggestion
Great. But what if you do not have a dual-function system?
```

Hmm, i don't really understand this sentence. If you have a single function system this doc recommends using bang_bang right?

kbx81: Not necessarily -- the recommendation is made at the end in the "Which is Right for Me?" section.
We could remove the line, it's not terribly important. It's just there to help transition the focus from dual-function systems to user interface differences.

changelog/v1.15.0.rst
kbx81: Awesome! Thank you for pointing that out @OttoWinter 😄

components/climate/bang_bang.rst
kbx81: What if we replaced it with this?:
"Now that we've highlighted the behavioral differences between these two components, let's take a look at the other important difference: User Interface."

OttoWinter: Yes that would be good (Just was confused when I first read that sentence, I didn't understand what it was referring to)

kbx81: Sweet! I pushed up a change with a modification similar to that. 😄  See https://github.com/esphome/esphome-docs/pull/667/commits/bd55f024fc0e9774cdea5f5cc0fb80ccd8005a49

OttoWinter (approved):

PR #669, opened 2020-07-04
Update pzemac.rst set stop_bits to 1

pciavald:
Could not get the sensor to work with stop_bits set to 2, found [this comment](https://github.com/esphome/issues/issues/817#issuecomment-552798845) recommending to use stop_bits: 1 and now it works.


gio-dot:
Hi, same pull request was made before here: https://github.com/esphome/esphome-docs/pull/636

PR #683, opened 2020-07-14
Add exposure notifications docs

OttoWinter:
## Description:

https://github.com/esphome/esphome/pull/1135

Image is from the [Google Exposure Notifications website](https://www.google.com/covid19/exposurenotifications/)

## Checklist:


components/exposure_notifications.rst
```diff
@@ -0,0 +1,74 @@
...
+      on_exposure_notification:
+        then:
+          - lambda: |
+              ESP_LOGD("main", "Got notification:");
+              ESP_LOGD("main", "  RPI: %s", hexencode(x.rolling_proximity_identifier).c_str());
+              ESP_LOGD("main", "  RSSI: %d", x.rssi);
+
+          # Stop existing timer so that turn_off doesn't get called
+          - script.stop: start_led
+          - script.execute: start_led
```
glmnet, line 75: it might be worth adding a new verb which does this, in fact, I find it really confusing behavior when two or more instances of a script will run at the same time, I did not find documentation about this "multi instance support" is this something we want to have? why not `script.execute` implicitly stop any running instance and restarts it?

OttoWinter: Yeah a verb could be added for that (as well as a general rethink of the names of these actions)

I can't immediately think of a use case for multi-instance scripts, but I do think there are some. Might be because HA also does it this way.

I think the main reason they exist though is recursive scripts. What do you do with a `script.execute` action with itself as the parameter? You can't wait for it, and aborting would not really be a solution.

The existing system isn't *that* bad in terms of stuff it supports, we could really improve the naming here though. But that's for another PR :)

components/exposure_notifications.rst
```diff
@@ -0,0 +1,74 @@
+BLE Exposure Notification Listener
+==================================
+
+.. seo::
+    :description: Instructions for setting up exposure notification listeners for ESPHome.
+    :image: exposure_notifications.png
+
+The ``exposure_notifications`` component uses the :doc:`/components/esp32_ble_tracker` to discover
+nearby COVID-19 exposure notification bluetooth messages sent by phone's running the
```
OttoWinter:
```suggestion
nearby COVID-19 exposure notification bluetooth messages sent by phones running the
```

oops

components/exposure_notifications.rst
```diff
@@ -0,0 +1,74 @@
...
+          - lambda: |
+              ESP_LOGD("main", "Got notification:");
+              ESP_LOGD("main", "  RPI: %s", hexencode(x.rolling_proximity_identifier).c_str());
+              ESP_LOGD("main", "  RSSI: %d", x.rssi);
+
+Configuration variables:
+------------------------
+
+- **on_exposure_notification** (*Optional*, :ref:`Automation <automation>`): An automation
+  to run when a exposure notification bluetooth message it received.
```
OttoWinter:
```suggestion
  to run when an exposure notification bluetooth message is received.
```

PR #684, opened 2020-07-14
Update GH Actions CI Workflow

OttoWinter:
## Description:

Primary change is to not build twice for branches on this remote. ~~And restricts the pull request trigger to only certain events (so that it doesn't trigger on anything like comments)~~

## Checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,23 +2,24 @@ name: Lint
 
 on:
   push:
+    # On other branches the `pull_request` trigger will be used
+    branches: [current, next]
+
   pull_request:
+    # Only trigger on certain events (not when comments are added)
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -2,23 +2,24 @@ name: Lint
 
 on:
   push:
+    # On other branches the `pull_request` trigger will be used
+    branches: [current, next]
+
   pull_request:
+    # Only trigger on certain events (not when comments are added)
+    types: [opened, reopened, synchronize]
```
glmnet: these are default https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

```suggestion
```

OttoWinter: Oops, mea culpa. Thanks!

PR #685, opened 2020-07-14
Use inclusive terminology

OttoWinter:
## Description:

See also https://github.com/esphome/esphome/pull/1137

## Checklist:

### DIFF
--- a/cookbook/arduino_port_extender.rst
+++ b/cookbook/arduino_port_extender.rst
@@ -58,7 +58,7 @@ individual IOs.
           auto ape_component = new ArduinoPortExpander(i2c_component, 0x08);
           return {ape_component};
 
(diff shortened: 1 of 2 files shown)


glmnet:
I just checked this https://next.esphome.io/search.html?q=slave and https://next.esphome.io/search.html?q=whitelist

Should we change this too? (this is all new for me)

OttoWinter:
@glmnet Not a huge problem if we fix it eventually (it's not like this is incredibly time sensitive). It's just some terms that can be associated with some awful history, and the industry has been using them for way too long.

Anyway, I just merged `current` into `next`, which will also propagate those changes to next.esphome.io

glmnet:
I realized later you updated `current`. I'm just excited about this project moving on again and reacted too quickly, sorry. And I still don't understand how all the merges for new versions works, which are manual and which automated, etc.. so I'll keep watching.

PR #691, opened 2020-07-17
Add documentation for MCP9808 sensor

k7hpn:
## Description: Add documentation for MCP9808 sensor

**Related issue (if applicable):** fixes esphome/feature-requests#214

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1169

## Checklist:


components/sensor/mcp9808.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Configuration variables:
+------------------------
+
+- **temperature** (**Required**): The information for the temperature sensor.
+
+  - **name** (**Required**, string): The name for the temperature sensor.
+  - **id** (*Optional*, :ref:`config-id`): Set the ID of this sensor for use in lambdas.
+  - All other options from :ref:`Sensor <config-sensor>`.
+
```
OttoWinter, line 40: Please also add `- **address** ...` as in other i2c-based sensors (this config option is automatically inserted by `i2c.i2c_device_schema`)

OttoWinter (commented):
Looks good, thanks!

I compressed the index image a bit further (as that's loaded whenever someone visits esphome.io). Also left a comment about the address option. Once that's addressed, this would be ok to merge.

components/sensor/mcp9808.rst
k7hpn: I think I got this added, thanks for pointing it out!

PR #705, opened 2020-07-24
Added docs for pulse_counter total

FrankBakkerNl:
## Description:
Docs for total sensor of pulse_counter

**Related issue (if applicable):** closes esphome/feature-requests#763

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1173

## Checklist:

### DIFF
--- a/components/sensor/pulse_counter.rst
+++ b/components/sensor/pulse_counter.rst
@@ -47,6 +47,10 @@ Configuration variables:
 - **update_interval** (*Optional*, :ref:`config-time`): The interval to check the sensor. Defaults to ``60s``.
 
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
+
+- **total** (*Optional*): Report the total number of pulses
+  All options from :ref:`Sensor <config-sensor>`.
+
 - All other options from :ref:`Sensor <config-sensor>`.
 
 .. note::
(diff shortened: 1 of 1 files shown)


FrankBakkerNl:
@OttoWinter any idea why the checks are failing? The error I see does not seem to be related to my changes

jesserockz (approved):

components/sensor/pulse_counter.rst
```diff
@@ -73,6 +77,31 @@ count the light pulses on a power meter, you can do the following:
...
+        unit_of_measurement: 'kW'
+        name: 'Power Meter House'
+        filters:
+          - multiply: 0.06
+
+        total:
+          unit_of_measurement: 'kWh'
+          name: 'Energy Meter House'
+          filters:
+            - multiply: 0.001
```
pedrolamas, line 103: @FrankBakkerNl if the counter is outputing in kW and the total in kWh, why do you need to multiply by 0.001 here?

FrankBakkerNl: The total counter counts the actual pulses, not what is output by the kW sensor. The kW sensor also has a multiply 0.06 here. I do not remember the math 100% but I think that translates from minutes to hours and W to kW

PR #711, opened 2020-07-26
Update light.partition documentation

Sangoku:
## Description:
As pointed out in https://community.home-assistant.io/t/esphome-partition-examples/117568 the documentation is kinda misleading. 
I am just improving the docs so other ppl have an easier expirience. 

**Related issue (if applicable):** fixes https://community.home-assistant.io/t/esphome-partition-examples/117568


OttoWinter (commented):
Ok adding more examples is good. However, I'd rather have them below in a new section (for example "Different Layouts").

The top configuration block should only contain the bare minimum so that the user can understand what config options exist.

jesserockz:
I did not see this PR and this issue came up again in chat so I altered in #1008

PR #716, opened 2020-07-27
Fix old screenshots

glmnet:
## Description:

Just replaced two screenshots. Also removed one image as it was a duplicate

## Checklist:


glmnet:
Is it a good idea **current** being default branch? I see lots of people confusing it usually, not that it just happened too me 👀

OttoWinter:
@glmnet Depends on what you want to optimize. If you want to make it easy to submit documentation fixes/improvements on the `current` branch, the default branch should be `current`.

Vice versa for next branch and new integrations.

Unfortunately GitHub just doesn't give us the tools to optimize both (at least last time I checked, but things might have changed since then). I believe this is also the setting home-assistant.io repo uses.

OttoWinter (approved):
Thanks! Yeah we should probably re-do some of the screenshots for the new release (especially in https://github.com/esphome/hassio)

PR #718, opened 2020-07-28
Include new v1.15 components  from diff

glmnet:
## Description:

Completed list after feature freeze

## Checklist:


OttoWinter (approved):
:+1:

OttoWinter:
Just as an FYI to you, the release scripts will automatically cherry-pick all PRs with a milestone equal to the release being deployed; so if a change should be part of a release you can add it to the milestone;

I'll create a longer writeup of the release script later, probably when 1.15.1 is released because then all bugs should have been taken care of. (a high level overview of how the _process_ works is already available here: https://github.com/esphome/esphome-release/blob/master/NOTES.md - I'll add the `esphomerelease` commands later)

PR #719, opened 2020-07-29
Add hyphen to supported name characters

### LINKED ISSUE #418: Allow hyphens in node name
Allow hyphens in node name.

Node names currently allow underscore, which is in the set of valid mDNS service identifier and DHCP client id characters, but not in the set of (strictly) valid DNS hostname characters.  Some people have DHCP servers which create dynamic DNS names, and DNS servers which are configured to strictly verify valid hostnames.

Rather than altering network topology to accommodate mDNS, I would like to simply use DNS name resolution to reach ESP nodes.  Cosmetically, I'd like to use a separator in the node hostnames.  Hyphen is the only functional character for this.

**Additional context**
There are, realistically, other options for resolving this issue:
 * I could use a subdomain specific to the ESP devices rather than a naming convention.  That's moderately inconvenient, as it means an additional configuration step rather than just letting the DHCP server hand out an address wherever on the network the device attaches
 * I could alter the DNS server to accept underscores.  The drawback to this is that future devices could create hostnames with underscores, and underscores are not valid in real hostnames, dang it. :D 
 * I could just mash the hostnames together instead of using hyphens.  But that decreases legibility, and reduced legibility increases the possibility of mistakes later on.
 * I could rearrange my network so multicast DNS worked reliably.  That's kind of a hassle, though. :)

After a quick glance, it looks like this is just a matter of adding a hyphen to the string at https://github.com/esphome/esphome/blob/7ad593d674e488e3abbfb3208326e3cbcd404d24/esphome/config_validation.py#L46 - I'll wait to open a PR in a few hours, because Hacktoberfest. 👕

ianleeder:
## Description:
Documentation update for esphome/esphome#1223

**Related issue (if applicable):** fixes  esphome/feature-requests#418

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1223

## Checklist:

### DIFF
--- a/guides/configuration-types.rst
+++ b/guides/configuration-types.rst
@@ -28,7 +28,7 @@ names <https://venus.cs.qc.cuny.edu/~krishna/cs111/lectures/D3_C++_Variables.pdf
 
 -  … must start with a letter and can end with numbers.
 -  … must not have a space in the name.
--  … can not have special characters except the underscore (“_“).
+-  … can not have special characters except underscore (“_“) and hyphen ("-").
 -  … must not be a keyword.
 
 .. _config-pin:
--- a/guides/configuration-types.rst
+++ b/guides/configuration-types.rst
@@ -28,7 +28,7 @@ names <https://venus.cs.qc.cuny.edu/~krishna/cs111/lectures/D3_C++_Variables.pdf
 
 -  … must start with a letter and can end with numbers.
 -  … must not have a space in the name.
--  … can not have special characters except underscore (“_“) and hyphen ("-").
+-  … can not have special characters except the underscore (“_“).
 -  … must not be a keyword.
 
 .. _config-pin:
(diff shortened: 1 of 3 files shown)


guides/configuration-types.rst
```diff
@@ -28,7 +28,7 @@ names <https://venus.cs.qc.cuny.edu/~krishna/cs111/lectures/D3_C++_Variables.pdf
 
 -  … must start with a letter and can end with numbers.
 -  … must not have a space in the name.
--  … can not have special characters except the underscore (“_“).
+-  … can not have special characters except underscore (“_“) and hyphen ("-").
```
OttoWinter: Uhm no this line should stay. This section is about `id` names, so the names of variables. C++ does not allow hyphens there.

ianleeder: Roger, I got trigger-happy searching for underscore.  Fixed now.

OttoWinter:
Oops, this needs to be rebased on next as this is a new feature not a docs fix/improvement

Edit: Done

PR #724, opened 2020-08-02
Add the ability to do a forced re-calibration on the SCD30

zeroping:
## Description:

Most of the info is over in the esphome PR.

This actually needs to be documented on top of the [last documentation changes for the SCD30, which look like they may have been dropped](https://github.com/esphome/esphome-docs/pull/473)

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1237

## Checklist:

  - [x] Link added in `/index.rst` when creating new documents for new components or cookbook. (not needed)


zeroping:
@Sizurka, you might be interested, given your prior work on the SDC30 - especially since this builds onn https://github.com/esphome/esphome/pull/962.

Sizurka (approved):

zeroping:
Just a related comment here: I think it makes sense to do a better introductory write-up on CO2 sensors and their limitations, especially discussing calibration. I'm guessing that should be a separate page, since it's shared information between the SCD30, the MHZ-19, and the CCS811. Many users may naively think (like I initially did) that these sensors can work in indoor environments without some kind of calibration process. 

I'm interested in writing it up, but I wanted to get these changes in first. I wouldn't mind suggestions on where to put such a write-up - maybe a cookbook page?

PR #734, opened 2020-08-16
Add carrier_frequency for IR device example

abmantis:
## Description:
carrier_frequency is needed for most IR devices. The most common for remotes is 38kHz.

## Checklist:


glmnet:
Indeed, default value is [0Hz](https://github.com/esphome/esphome/blob/e24d5c172fe16d13a1d67e9a3f97b92e6b9bcfa7/esphome/components/remote_base/__init__.py#L404)

PR #739, opened 2020-08-22
Fixing pin definition error

AnsgarSchmidt:
## Description:
The example shows the wrong pin definition.

https://github.com/esphome/issues/issues/943

## Checklist:


OttoWinter (approved):
Hmm, I'm not sure I understand.

The example here is just based on how somebody has wired their sensor to the ESP - on ESP8266 you (mostly) can use all pins for all functions.

Could you elaborate a bit?

Thanks!

jesserockz:
I don't see the need for this as they are just user defined pins and can be anything.

PR #740, opened 2020-08-23
Add missing doc for rc_switch event

micw:
The event was added with https://github.com/esphome/esphome/pull/983 but the doc does not reflect this change.

## Description:

https://github.com/esphome/esphome/pull/983

## Checklist:


OttoWinter (approved):
Thanks!

PR #741, opened 2020-08-23
docs: necessary registration of node in HA 

torwag:
Add remark about registration in home assistant.
## Description:
Even one never plan to export data to home assistant and therefore do not need any entity of the node itself created in home assistant, it is still required to  register the node in home assistant if you want to __import__ data from home assistant into the node. That might not be clear if you come from the "other" side just trying to get data into a node. The paragraph should remind people to do that.

## Checklist:


torwag:
If this PR finds acceptance, I am willing to send in a new one add the same graph for binary_sensors and text_sensors based on home assistant.

jesserockz (approved):
Sounds good. Feel free to add the same to the other `homeassistant` ones in either this PR or new ones. I will await a reply before merging this one

PR #747, opened 2020-08-24
using docker image to build esphome-docs locally

demikl:
## Description:

Clarification on how to use the esphome-docs docker image for bulding docs locally. As a first time contributor, I had to dig into the Dockerfile for figure out which volume binding would work.

**Related issue (if applicable):** no issue opened

## Checklist:


OttoWinter (approved):
Thanks!

PR #750, opened 2020-09-03
Add uart_id and energy on Pzemdc

lucasimons:
Description:

Now the pzemdc as, pzem 017, not support uart_id and energy meter, indeed the pzemac support all this two thinghs 

Related issue (if applicable): fixes

Checklist:
 Branch: next is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against current.
 Link added in /index.rst when creating new documents for new components or cookbook.


jesserockz:
@lucasimons This PR is not correct, can you please check your branches and open a new PR with changes if it is still valid.

PR #753, opened 2020-09-06
clarify VCC, wrt #1452

### LINKED ISSUE #1452: documentation example of measuring VCC voltage does not provide sensible data
**Operating environment/Installation (Hass.io/Docker/pip/etc.):**

integrated with hass.io

**ESP (ESP32/ESP8266, Board/Sonoff):**

ESP8266/d1_mini & nodemcuv2

**ESPHome version (latest production, beta, dev branch)**

ESPHome v1.14.5

**Affected component:**

https://esphome.io/components/sensor/adc.html#adc-esp8266-vcc
https://github.com/esphome/esphome-docs/blob/current/components/sensor/adc.rst

**Description of problem:**
when configuring the sensor according to documentation, "meassures" 2.91V.
this value does not change over time and is even the same on different devices (including different power supplies like 6V or 5V via USB).

**Problem-relevant YAML-configuration entries:**
```yaml
sensor:
  - platform: adc
    pin: VCC
    name: "VCC Voltage"
```

**Logs (if applicable):**

```
[D][adc:056]: 'VCC Voltage': Got voltage=2.91V
```

**Additional information and things you've tried:**

pille:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/issues/issues/1452

### DIFF
--- a/components/sensor/adc.rst
+++ b/components/sensor/adc.rst
@@ -72,14 +72,14 @@ To measure voltages higher than 1.1V, set ``attenuation`` to one of the `followi
 ESP8266 Measuring VCC
 ---------------------
 
-On the ESP8266 you can even measure the voltage the chip is getting. This can be useful in situations
+On the ESP8266 you can even measure the voltage the *chip is getting*. This can be useful in situations
 where you want to shut down the chip if the voltage is low when using a battery.
 
 To measure the VCC voltage, set ``pin:`` to ``VCC`` and make sure nothing is connected to the ``A0`` pin.
 
 .. note::
 
-    To avoid confusion: It measures the voltage at the chip, and not at the VCC pin of the board. It is usually static at 2.91V.
+    To avoid confusion: It measures the voltage at the chip, and not at the VCC pin of the board. It should usually be around 3.3V.
     
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


components/sensor/adc.rst
```diff
@@ -77,6 +77,10 @@ where you want to shut down the chip if the voltage is low when using a battery.
 
 To measure the VCC voltage, set ``pin:`` to ``VCC`` and make sure nothing is connected to the ``A0`` pin.
 
+.. note::
+
+    To avoid confusion: It measures the voltage at the chip, and not at the VCC pin of the board. It is usually static at 2.91V.
```
OttoWinter: Maybe we can elaborate on `at the chip` more. This option measures the logic supply voltage to the chip (around 3.3V).

What I don't quite understand is why it would usually be "static at 2.91V". I don't see that here, and I suspect that depends on the power supply.

pille: i got the 2.91V on two different device models (`nodemcuv2`, `d1_mini`) with different power supplies (`USB`, `6V VCC`).
i can do some more testing, but i'm also interested what your devices show.

pille: looks like a case of extrapolating from incomplete data. my ESP8285 (tuya power socket) reports slighly varying values around 3.46V.

jesserockz (approved):

PR #760, opened 2020-09-13
Update diy.rst

kquinsland:
Added two links for custom projects that ESPHome powered:

1. A custom UART based component that parses data from weather station and gets the values into MQTT
2. Using the 'built in' UART component in ESPHome to send commands to a LG Monitor. This can easily be adopted to other LG monitors as they usually share similar payloads.

-----

This does not fix any issues, only links to a few projects.


OttoWinter (approved):
Thanks!

PR #761, opened 2020-09-14
Add battery_level note for xiaomi_hhccjcy01

axilleas:
## Description

In https://github.com/esphome/esphome/pull/1027/files#diff-7db33c55aa814ed3c8ad001b3fc60471,
the `battery_level` attribute was removed from `xiaomi_hhccjcy01` as it was [no longer being
advertised by the device](https://github.com/esphome/issues/issues/107#issuecomment-510403045).

This PR removes the `battery_level` from the `xiaomi_hhccjcy01` example and adds a note why.

Review app: https://deploy-preview-761--esphome.netlify.app/components/sensor/xiaomi_ble.html#hhccjcy01

## Checklist:

### DIFF
--- a/components/sensor/xiaomi_ble.rst
+++ b/components/sensor/xiaomi_ble.rst
@@ -38,6 +38,11 @@ Configuration example:
         battery_level:
           name: "Xiaomi HHCCJCY01 Battery Level"
 
+.. note::
+
+    Newer versions of HHCCJCY01 ship with firmware 3.2.1, and they
+    `don't send the battery level data anymore <https://github.com/esphome/esphome/pull/1288#issuecomment-695809481>`__.
+
 GCLS002
 *******
 


jesserockz:
This was re-added back with esphome/esphome#1288
Yes devices with newer firmware do not have this feature, but devices with older firmware still have `battery_level`

jesserockz:
Maybe alter this PR to add the note, but in it mention that the battery_level might not be available depending on your `hhccjcy01` firmware version?

axilleas:
> Maybe alter this PR to add the note, but in it mention that the battery_level might not be available depending on your `hhccjcy01` firmware version?

@jesserockz yep, that's what I was thinking. I'll get to it.

axilleas:
@jesserockz ok, this should be good to go :rocket:

jesserockz (approved):
Thanks

PR #772, opened 2020-09-19
Fix APDS9960 datasheet link

JonathanTreffler:
## Description:
The datasheet link in the APDS9960 sensor description currently points to a datasheet for the BME280 sensor, wich is completely unrelated.

## Checklist:


jesserockz (approved):
Thanks

PR #784, opened 2020-10-01
Remove bh7150 API reference

JeffResc:
## Description:
This PR moves to remove the API reference from the [BH1750 page](https://esphome.io/components/sensor/bh1750.html#see-also) as it seems to no longer exist. I couldn't find an equivalent link to replace it with, so I decided to remove it entirely, but if someone knows the new URL to the API reference (if it still exists), please let me know and I can replace it with that new URL.

## Checklist:

### DIFF
--- a/components/sensor/bh1750.rst
+++ b/components/sensor/bh1750.rst
@@ -59,5 +59,6 @@ See Also
 
 - :ref:`sensor-filters`
 - :doc:`tsl2561`
+- :apiref:`bh1750/bh1750.h`
 - `BH1750 Library <https://github.com/claws/BH1750>`__ by `@claws <https://github.com/claws>`__
 - :ghedit:`Edit`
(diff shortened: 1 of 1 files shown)


components/sensor/bh1750.rst
```diff
@@ -59,6 +59,5 @@ See Also
 
 - :ref:`sensor-filters`
 - :doc:`tsl2561`
```
jesserockz, line 61: Looks like it was just misspelt and should be this which links [here](https://esphome.io/api/bh1750_8h.html)
```suggestion
- :doc:`tsl2561`
- :apiref:`bh1750/bh1750.h`
```

JeffResc:
Thanks, @jesserockz. I've added your commit. I'm not too familiar with this component, in particular, so thank you for pointing that out as it wasn't too obvious to me.

jesserockz (approved):
Thanks @JeffResc. I am not either but the apidocs are generated from the source code automatically so if the component exists, the doc should be there.

PR #791, opened 2020-10-04
Individual BMXXXXX sensor configs are optional

trvrnrth:
## Description:
For the BME280, BME680, BMP085 and BMP280 sensors the individual sensors are optional.

## Checklist:


fabaff (approved):
Thanks

PR #792, opened 2020-10-04
BME680 default address is 0x76

trvrnrth:
## Description:
The BME680 default address config is 0x76, with the secondary address being 0x77.

See https://github.com/esphome/esphome/blob/002861f13b2e3a9dfd16a1157b12894daf08a6cb/esphome/components/bme680/sensor.py#L62

## Checklist:


jesserockz (approved):
Thanks

fabaff (approved):

PR #793, opened 2020-10-04
Add BME680 BSEC docs

trvrnrth:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1313

## Checklist:


components/sensor/bme680_bsec.rst
```diff
@@ -0,0 +1,192 @@
...
+The ``bme680_bsec`` sensor platform allows you to use your BME680
+(`datasheet <https://cdn-shop.adafruit.com/product-files/3660/BME680.pdf>`__,
+`Adafruit`_) temperature, pressure and humidity and gas sensors with ESPHome via the Bosch Sensortec Environmental Cluster (BSEC)
+software library. The use of Bosch's proprietary algorithms provides additional Indoor Air Quality (IAQ), CO2 equivalent and Breath
+Volatile Organic Compounds (VOC) equivalent measurements.
+
+.. note::
+
+    The Bosch BSEC library is only available for use after accepting its software license agreement. By enabling this component,
+    you are explicitly agreeing to the terms of the `BSEC license agreement`_.
```
OttoWinter: So if someone shares a build (the "`.bin`" file) with this component enabled, they would technically violate the ESPHome GPL license. Let's add a line saying that any firmware binary can't be shared when this component is enabled.

trvrnrth: That sounds reasonable to me if that's what is required to meet the GPL requirements. My understanding of the licensing specifics for this scenario aren't really sufficient to give any input so on that basis I'm perfectly happy to add that as an extra restriction.

trvrnrth: Added words to that effect in b1daada.

OttoWinter (approved):
Thanks! LGTM :+1:

PR #799, opened 2020-10-08
Optimize images

JeffResc:
## Description:
I have ImgBot set up to automatically run on my repositories and when I made a fork of this repo, it automatically ran on that one too. I noticed that ImgBot was able to reduce the size by ~10% so I figured I would just forward its PR to here. It's automatically setup to do lossless compression, so there should be no image quality degradation, either.
```
*Total -- 20,960.25kb -> 18,836.18kb (10.13%)
```

## Checklist:


fabaff (approved):
Thanks

PR #804, opened 2020-10-11
Midea Climate support

dudanov:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1328

## Checklist:


glmnet (changes requested):
Is there a reason to have two docs file? I.e. is a component useful without the other?

Otherwise please merge them in a sine file were it will be easier to go thru the docs

dudanov:
> Is there a reason to have two docs file? I.e. is a component useful without the other?
> 
> Otherwise please merge them in a sine file were it will be easier to go thru the docs

The idea is to separate the transport component (midea-dongle) common to all Midea devices ("usb" uart stick) and the components of various devices (currently supported air conditioners (midea-ac). It is known that there are other devices of this company, similarly controlled (humidifiers, water heaters, etc.)

glmnet:
Well, in that case merging the docs will be preferred as there is no need to separate docs, you can even make the dongle component be auto loaded so you only have to configure the climate component itself, this makes a cleaner yaml file, you can still add the `midea_dongle:` element explicitly if you want to override some setting or the uart used, etc.

components/climate/midea_ac.rst
```diff
@@ -0,0 +1,92 @@
...
+  - All other options from :ref:`Sensor <config-sensor>`.
+- **humidity_setpoint** (*Optional*): The information for the humidity indoor
+  sensor (experimental).
+
+  - **name** (**Required**, string): The name of the sensor.
+  - **id** (*Optional*, :ref:`config-id`): Set the ID of this sensor for use in lambdas.
+  - All other options from :ref:`Sensor <config-sensor>`.
+- **beeper** (*Optional*, bool, default: **false**): Beeper feedback on command.
+- **swing_horizontal** (*Optional*, bool, default: **false**): Enable **swing horizontal** option.
+- **swing_both** (*Optional*, bool, default: **false**): Enable **swing both** option.
```
glmnet: This is not how we document defaults, see other components for examples.
Usually a statement at the end like: "Defaults to ``false``."

glmnet (approved):

components/climate/midea_ac.rst
dudanov: Done.

PR #809, opened 2020-10-12
Update docs about MQTT lambdas

matikij:
## Description:
Update docs about mqtt subscription in lambdas, add docs on mqtt connected in lambdas

Note that I am not 100% sure that subscription will work, but I would expect it to: when I tried to call the code in docs I got:

```
src/main.cpp:968:15: error: 'class esphome::Application' has no member named 'get_mqtt_client'
       if (App.get_mqtt_client()->is_connected()) {
```

**Related issue (if applicable):** fixes - no issue yet, simple fix

## Checklist:

### DIFF
--- a/components/mqtt.rst
+++ b/components/mqtt.rst
@@ -334,9 +334,15 @@ Configuration variables:
 
     This action can also be used in :ref:`lambdas <config-lambda>`:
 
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Thanks

PR #813, opened 2020-10-17
Update waveshare_epaper.rst

nejc-cc:
## Description:
I tried using a d1_mini_pro for driving an e-ink display. Wouldn't work. Found out why:
Got "[E][display:017]: Could not allocate buffer for display!" from logs.
The Waveshare display 7.50inV2 has a resolution of 800x480 pixels and ESP8266 runs out of RAM.

I hope this is the right way for me to propose change in documentation (only that). First time doing this :)

## Checklist:


jesserockz (approved):
Thanks

PR #820, opened 2020-10-26
Added manual for Itead Sonoff iFAN03

sliwma:
## Description:

**Related issue (if applicable):** fixes github.com/esphome/feature-requests/issues/784

## Checklist:

### DIFF
--- a/cookbook/ifan03.rst
+++ b/cookbook/ifan03.rst
@@ -53,14 +53,14 @@ Then you need to set it up with yaml.
 .. code-block:: yaml
 
     esphome:
-  name: fan_ifan03
-  platform: ESP8266
-  board: esp8285
-  includes:
-    - ifan03.h
+      name: fan_ifan03
+      platform: ESP8266
+      board: esp8285
+      includes:
+        - ifan03.h
     wifi:
-  ssid: <YOUR_SSID>
-  password: <YOUR_PASSWORD>
+      ssid: <YOUR_SSID>
+      password: <YOUR_PASSWORD>
 
     captive_portal:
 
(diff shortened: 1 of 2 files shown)


jesserockz:
While I expect this works, it would be better to pass through the switches to the custom component:
`new IFan03Output(id(relay_fan1), id(relay_fan2), id(relay_fan3));` and then turn on the switches from the custom component instead of directly using `digitalWrite`

sliwma:
If only I had some time... It is based on the iFan02 config and works for me.

ianchi:
Hi @sliwma @jesserockz, I have some doubts with this recipe.
 

+ what is the purpose of `update_fan_speed`? 
It never gets "called" so the sync never takes place. At least I couldn't see where.

+ why use a custom component? I find it more complicated than necessary. I've accomplished the same with a simple `switch.turn_on` `switch.turn_off` in the `output write_action`.

I leave an example:

```yaml
output:
  - platform: template
    type: float
    id: fan_output
    write_action:
      - if:
          condition:
            lambda: return (state < 0.3);
          then:
            # OFF
            - switch.turn_off: relay_fan1
            - switch.turn_off: relay_fan2
            - switch.turn_off: relay_fan3
      - if:
          condition:
            lambda: return ((state >= 0.3) && (state < 0.6));
          then:
            # LOW SPEED
            - switch.turn_on: relay_fan1
            - switch.turn_off: relay_fan2
            - switch.turn_off: relay_fan3
      - if:
          condition:
            lambda: return ((state >= 0.6) && (state < 0.9));
          then:
            # MID SPEED
            - switch.turn_off: relay_fan1
            - switch.turn_on: relay_fan2
            - switch.turn_off: relay_fan3
      - if:
          condition:
            lambda: return (state >= 0.9);
          then:
            # MAX SPEED
            - switch.turn_off: relay_fan1
            - switch.turn_off: relay_fan2
            - switch.turn_on: relay_fan3
```

sliwma:
Hey, as mentioned - I just modified the iFan02 config to make my iFan03 fully functional.
Your solution looks better tho.

PR #828, opened 2020-10-30
Add separate current/target Tuya Climate multipliers

Yarikx:
## Description:
Some Tuya thermostats have different multipliers for current and target temperatures.
So the current approach (having a single `temperature_multiplier`) cannot handle both cases.

This is breaking change, it introduces separate multipliers for both values

**Related issue (if applicable):

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):
https://github.com/esphome/esphome/pull/1345

## Checklist:

### DIFF
--- a/components/climate/tuya.rst
+++ b/components/climate/tuya.rst
@@ -49,7 +49,7 @@ Configuration variables:
 - **target_temperature_datapoint** (**Required**, int): The datapoint id number of the target temperature.
 - **current_temperature_datapoint** (**Required**, int): The datapoint id number of the current temperature.
 - **temperature_multiplier** (**Optional**, float): A multiplier to modify the incoming and outgoing temperature values - :ref:`see below <temperature-multiplier>`.
-  If the device has different multipliers for current and target temperatures, **temperature_multiplier** can be replaced with either or both of:
+  If the device has different multipliers for current and target temperatures, **temperature_multiplier** can be replaced with both of:
 
     - **current_temperature_multiplier** (**Optional**, float): A multiplier to modify the current temperature value.
     - **target_temperature_multiplier** (**Optional**, float): A multiplier to modify the target temperature value.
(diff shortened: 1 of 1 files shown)


components/climate/tuya.rst
```diff
@@ -48,7 +48,8 @@ Configuration variables:
 - **switch_datapoint** (**Required**, int): The datapoint id number of the climate switch.
 - **target_temperature_datapoint** (**Required**, int): The datapoint id number of the target temperature.
 - **current_temperature_datapoint** (**Required**, int): The datapoint id number of the current temperature.
-- **temperature_multiplier** (**Optional**, float): A multiplier to modify the incoming and outgoing temperature values - :ref:`see below <temperature-multiplier>`.
+- **current_temperature_multiplier** (**Optional**, float): A multiplier to modify the current temperature value - :ref:`see below <temperature-multiplier>`.
```
jesserockz:
```suggestion
- **temperature_multiplier** (**Optional**, float): A multiplier to modify the incoming and outgoing temperature values - :ref:`see below <temperature-multiplier>`.
- **current_temperature_multiplier** (**Optional**, float): A multiplier to modify the current temperature value - :ref:`see below <temperature-multiplier>`.
```

jesserockz (changes requested):
Please leave the existing config option in the docs. Maybe add an exclusion note somewhere?

Yarikx:
Moved new config options under the `temperature_multiplier` item. Explained usage a bit more.

components/climate/tuya.rst
```diff
@@ -49,6 +49,10 @@ Configuration variables:
 - **target_temperature_datapoint** (**Required**, int): The datapoint id number of the target temperature.
 - **current_temperature_datapoint** (**Required**, int): The datapoint id number of the current temperature.
 - **temperature_multiplier** (**Optional**, float): A multiplier to modify the incoming and outgoing temperature values - :ref:`see below <temperature-multiplier>`.
+  If the device has different multipliers for current and target temperatures, **temperature_multiplier** can be replaced with either or both of:
```
jesserockz: The validator I made makes them inclusive so both are required.

```suggestion
  If the device has different multipliers for current and target temperatures, **temperature_multiplier** can be replaced with both of:
```

jesserockz (approved):

PR #833, opened 2020-11-01
Link to Winsen ZE08-CH2O custom component

cretep:
Add link to C++ library and example ESPHome YAML config for reading Winsen's Formaldehyde sensor over serial.

Original code; other implementations I found do not switch the component into question-answer mode (to control update frequency) and they read each UART byte individually. Most Winsen sensors communicate over a serial protocol they have developed, but it is not consistent enough throughout their components to become a common 'winsen' library.


fabaff (approved):
Thanks

PR #834, opened 2020-11-01
Add contrast option docs to PCD8544

dzervas:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1348

## Checklist:

### DIFF
--- a/components/display/pcd8544.rst
+++ b/components/display/pcd8544.rst
@@ -39,6 +39,7 @@ VCC to 3.3V and GND to GND.
         reset_pin: D0
         cs_pin: D8
         dc_pin: D1
+        contrast: 60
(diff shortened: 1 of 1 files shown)


components/display/pcd8544.rst
```diff
@@ -39,6 +39,7 @@ VCC to 3.3V and GND to GND.
         reset_pin: D0
         cs_pin: D8
         dc_pin: D1
+        contrast: 60
```
jesserockz: It might be a good idea to have the example be the same as the default value
```suggestion
        contrast: 0x7f
```

jesserockz (approved):
LGTM

PR #838, opened 2020-11-08
Add warning about pn532 moved config key

jesserockz:
## Description:
Adds a warning about the removed `pn532` key to point users to `pn532_spi` or `pn532_i2c`

Related changes #798 / esphome/esphome#1302

## Checklist:


glmnet:
I'd say it will be better to bring back the pn532 with a dummy python implementation that will just error out showing which is the current configuration for this component, see also https://github.com/esphome/esphome/blob/dev/esphome/components/remote_transmitter/switch.py

jesserockz:
I never even saw that and already did that in python. Just thought I would put a message on the docs too, but if its not needed then all good =)

glmnet:
> I never even saw that and already did that in python. Just thought I would put a message on the docs too, but if its not needed then all good =)

I see the comment you've left there. That's good enough, people will quickly figure out how to get it working right away. I'm not a fan of polluting docs with version evolution stuff, docs applies to latest version, period.

jesserockz:
Sounds good

PR #845, opened 2020-11-14
Fix MCP23008 example 

djtef:
## Description:
Change the pin mode from INPUT_PULLUP to OUTPUT for the switch in the MCP23008 example

## Checklist:

### DIFF
--- a/components/mcp230xx.rst
+++ b/components/mcp230xx.rst
@@ -36,8 +36,7 @@ The MCP23008 component (`datasheet <http://ww1.microchip.com/downloads/en/device
           mcp23008: mcp23008_hub
           # Use pin number 0
           number: 0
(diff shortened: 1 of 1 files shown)


oncleben31 (approved):

djtef:
I observed that a switch can be defined as an input gpio without any error in yaml validation : 
```
switch:
  - platform: gpio
    pin: 
        number : 25
        mode: INPUT
    name: "Test switch as INPUT"
```
So the error in the example documentation I fixed is not detected by the yaml validation and can compile.
Do you think it is normal ?

djtef:
And same problem with an output gpio, this is considered as valid :
```
output:
  - platform: gpio
    pin: 
        number : 25
        mode: INPUT
```

jesserockz (approved):
Thanks

PR #851, opened 2020-11-18
Add troubleshooting help with light partition

TheLastGimbus:
## Description:
When I wanted to divide my single strip into multiple parts to use as separate HA entities, I've ran into some issues: original light component kept conflicting with the new partition.

I figured out that the best solution is to just mark original light as internal, and control all segments myself - not to be forced to have single entity that will turn on/off whole strip

So I added this in documentation of partition :smile: :+1: 

## Checklist:


fabaff (approved):
Thanks

PR #855, opened 2020-11-22
Fix teleinfo

djtef:
## Description:
Teleinfo component 
Correct yaml example (according to test1.yaml in esphome repo) and add a warning to disable loggin via UART

## Checklist:

### DIFF
--- a/components/sensor/teleinfo.rst
+++ b/components/sensor/teleinfo.rst
@@ -107,7 +107,8 @@ Configuration variables:
     logger:     
       baud_rate: 0   # disable logging via UART, help to avoid numerous crash with ESP_LOGD
       level: INFO   # INFO for less log, put DEBUG to view all the linky's "étiquettes" received  in the logs
-      esp8266_store_log_strings_in_flash: False     # recommanded for ESP8266 https://esphome.io/components/sensor/custom.html
+      esp8266_store_log_strings_in_flash: False     #  :doc:`recommanded for ESP8266 </components/sensor/custom>`
+      
     
 
 See Also
--- a/components/sensor/teleinfo.rst
+++ b/components/sensor/teleinfo.rst
@@ -97,6 +97,18 @@ Configuration variables:
 - **uart_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the :ref:`UART Component <uart>` if you want
   to use multiple UART buses.
 
+.. note::
+
+    On ESP8266, the logging via UART must be disabled to avoid crashes leading to boot loop :
+    
+    .. code-block:: yaml
+    
+    logger:     
+      baud_rate: 0   # disable logging via UART, help to avoid numerous crash with ESP_LOGD
+      level: INFO   # INFO for less log, put DEBUG to view all the linky's "étiquettes" received  in the logs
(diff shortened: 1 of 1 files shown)


components/sensor/teleinfo.rst
```diff
@@ -60,17 +60,17 @@ simply press -/+ buttons on the counter and look for `Standard mode` or
     sensor:
       - platform: teleinfo
         tags:
-         - name: "HCHC"
+         - tag_name: "HCHC"
```
0hax, line 63: Good catch, thanks for fixing this.

components/sensor/teleinfo.rst
```diff
@@ -97,6 +97,20 @@ Configuration variables:
...
+    
+.. code-block:: yaml
+
+    logger:     
+      baud_rate: 0   # disable logging via UART, help to avoid numerous crash with ESP_LOGD
+      level: INFO   # INFO for less log, put DEBUG to view all the linky's "étiquettes" received  in the logs
+      esp8266_store_log_strings_in_flash: False     #  :doc:`recommanded for ESP8266 </components/sensor/custom>`
+      
+    
+
```
0hax, line 113: Maybe we should tell something like this:
```
On ESP8266, only 2 UARTs are available, one of it is TX only. This means only one UART is available for Teleinformation as RX is needed.
Since the only UART available for Teleinformation is already used by the logger, we need to either set the logger to use the second UART which is TX only or totally disable the logger.
A third solution would be to use another pin not dedicated to UART by the mean of Software UART, but unfortunately, this has been very unstable during development and it's not recommended.
To disable logger or make it use another UART, you can have a look to <path/to/uart/wiki>
```
Does that looks good to you ?

0hax (changes requested):
Thanks for your contribution, I have made few review comments.

components/sensor/teleinfo.rst
djtef: When I read this [documentation ](https://arduino-esp8266.readthedocs.io/en/latest/reference.html#serial) i understand there are 2 UARTs indeed, but it is possible to map 2 others pins to the first one by calling `Serial.swap()`. Don't you think we can have both logging and teleinfo by swapping each time we want to read teleinfo ?

0hax: You are right, it's possible to use UART0 with GPIO1/GPIO3 or GPIO15/GPIO13 by using swap() function.
And it looks possible to use GPIO15/GPIO13 for UART0 in esphome using `UART0_SWAP` in the configuration.
But the swap() function is only called at configuration time, it's not changed dynamically after the configuration.
So I'm not sure if maintainers want us to use swap() from components directly and switch from one pin to another.
I think it's would also require re-configuring the UART to modify the baudrate for example each time we swap.
Maybe it's doable, but it sounds error prone to me.

djtef: I agree with you, maybe it should be done at core level to be transparent for the end user. I think it's dangerous to implement the swap inside a component itself. But the concept can be usefull if you want to use 2 UART on the ESP8266. Maybe we need to create a pull request about this subject.

PR #856, opened 2020-11-22
 Adding gif animation docs

Syndlex:
## Description:

**Related issue (if applicable):** fixes esphome/feature-requests#893

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1378

## Checklist:

### DIFF
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -376,6 +376,32 @@ be supplied to modify the color used to represent the on and off bits respective
 You can also use this to invert images in two colors display, use ``COLOR_OFF`` then ``COLOR_ON``
 as the additional parameters.
 
+Animation
+*********
+
+Animation inherits all options from the image component.
+It adds an additional method to change the shown picture of a gif.   
+
+.. code-block:: yaml
+
+    animation:
+      - file: "animation.gif"
+        id: my_animation
+        resize: 100x100
+
+The animation can be rendered just like the image component with the image() function of the display component.
+
+To show the next frame of the animation call ``id(my_animation).next_frame()``
+This can be combined with all automation:
+
+.. code-block:: yaml
+
+    interval:
+      - interval: 500ms
+        then:
+          lambda: |-
+            id(my_animation).next_frame();
+
 .. _display-pages:
 
 Display Pages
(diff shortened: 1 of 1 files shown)


components/display/index.rst
```diff
@@ -376,6 +376,32 @@ be supplied to modify the color used to represent the on and off bits respective
...
+.. code-block:: yaml
+
+    animation:
+      - file: "animation.gif"
+        id: my_animation
+        resize: 100x100
+
+The animation can be rendered just like the image component with the image() function of the display component.
+
+To show the next frame of the animation call ``id(my_animation).next_frame()``
```
glmnet, line 394: Wouldn't be an easy example to do 

```
 lambda: |-
      // Draw the image my_image at position [x=0,y=0]
      // with front color red and back color blue
      id(my_animation).next_frame();
      it.image(0, 0, id(my_image), id(red), id(blue));
````

so this would indeed update the frame everytime the render lambda is called for the display.

Syndlex: yes this would update the frame on every draw cycle.
This is a very good point.
 
It has some benefits and some drawbacks.
It's way easier!
But i had this problem in mind: 
The draw cycle depends on the display implementation.
This can be too fast for some animations.
It could even be yield to flickering on some displays. (I only tested this on this display https://github.com/esphome/esphome/pull/1377)
 
Maybe a better way is to implement animation as a Pollingcomponent and add an update field.
But this would  make the animation less flexible.

glmnet: Hmm most displays use update interval for calling the lambda

Syndlex: that's correct.
But the interval update is called depends on the Display implementation:
All current displays have a interval of 1s or 5s
Waveshare has 1s intervals
https://github.com/esphome/esphome/blob/dev/esphome/components/waveshare_epaper/display.py#L55
Nextion has 5s intervals
https://github.com/esphome/esphome/blob/dev/esphome/components/nextion/display.py

I think this value is set without the thought of Animations since there was no need to update the Display more than once a second.

I will change the Document according to your suggestion. Since its easier for most Users. Updating in a interval of 500ms wouldn't even work with current Display implementations.

PR #861, opened 2020-11-24
pulse counter: Show how calculations are made

fredrike:
## Description:

The examples for pulse counting are not 100% easy to follow. This adds more information.

## Checklist:

### DIFF
--- a/cookbook/power_meter.rst
+++ b/cookbook/power_meter.rst
@@ -31,10 +31,10 @@ And... that should already be it :)
 
     .. code-block::
 
-    S0 ------------ VCC
-    S0 --+-- 10k -- GND
-         |
-         +--------- GPIO12
+      S0 ------------ VCC
+      S0 --+-- 10k -- GND
+          |
+          +--------- GPIO12
 
 For ESPHome, you can then use the
 :doc:`pulse counter sensor </components/sensor/pulse_counter>` using below configuration:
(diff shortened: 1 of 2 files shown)


fredrike:
Would it be ok to mention a S0 connection like this for in the power_meter doc?

>
> Some meters have an available S0 port, if that is the case the photo diode can be replaced with the following connection.

```
S0 ------------- VCC
S0 ---+-- 10k -- GND
      |
      +--------- GPIO12
```

fabaff:
> Would it be ok to mention a S0 connection like this for in the power_meter doc?

Sure, everything that helps should be included.

fredrike:
I'm not sure why linting fails..

fredrike:
Changed to current as the upstream changes already have been merged.

cookbook/power_meter.rst
```diff
@@ -24,6 +24,18 @@ And... that should already be it :)
     :align: center
     :width: 80.0%
 
+.. note::
+
+    Some energy meters have an exposed S0 port (which essentially just is a switch that closes), if 
+    that is the case the photo diode can be replaced with the following connection.
```
frenck:
```suggestion
    Some energy meters have an exposed S0 port (which essentially just is a switch that closes), if
    that is the case the photodiode can be replaced with the following connection.
```

frenck: Additionally this kinda unrelated to the PR presented... Should this have been a separate PR?

fredrike: Well it could, but this is more of an "update doc" pr so perhaps we could just update the name.

frenck (approved):
Thanks, @fredrike 👍

PR #865, opened 2020-11-29
Update custom_component.rst

danoh:
There must be mqtt: component configured to compile the example.

## Description:

Improve documentation of CustomMQTTDevice, make implicit, that you have to configure the mqtt broker to use the component.

## Checklist:


jesserockz:
I don't think we need to add `mqtt` to the generic yaml... 
I think if you have your node configured already before you start writing custom components you should already know about setting up `mqtt`.
It also does not have the `api` component for the API custom component.

PR #870, opened 2020-12-04
Ili9341 doc update

kbx81:
## Description:

Updates #723 per changes requested there by @glmnet 

## Checklist:


kbx81:
FWIW, yesterday I was able to test this component and documentation on the ESP-WROVER-KIT that I have and it appears to work as expected.￼ 🎉

components/display/ili9341.rst
```diff
@@ -0,0 +1,128 @@
...
+      - platform: ili9341
+        model: TFT 2.4
+        cs_pin: 14
+        dc_pin: 27
+        led_pin: 32  ### see note below ###
+        reset_pin: 33
+        rotation: 0
+        
+        lambda: |-
+          it.fill(ili9341::COLOR_BLACK);
```
glmnet: Does this works? Should it work?

components/display/ili9341.rst
```diff
@@ -0,0 +1,128 @@
...
+          it.print(0, 0, id(font), "Hello World!");
```
glmnet: It'd be nice to showcase color print?

components/display/ili9341.rst
```diff
@@ -0,0 +1,128 @@
...
+
+    # Example minimal configuration entry
+    display:
+      - platform: ili9341
+        model: TFT 2.4
+        cs_pin: 14
+        dc_pin: 27
+        led_pin: 32  ### see note below ###
+        reset_pin: 33
+        rotation: 0
```
glmnet: If rotation 0 is default remove it from here. On the first example we tend to do the minimum working configuration for a component, so usually optional parameters are not specified.

glmnet (commented):
Looks good! Thanks for reviewing specially the samples given works. People always expect to copy paste code and get it working right away

components/display/ili9341.rst
kbx81: I will confirm...could be that I missed part of it. I'll check and fix as necessary.

components/display/ili9341.rst
kbx81: Agreed -- will update.

components/display/ili9341.rst
kbx81: Consider it done. 😄

components/display/ili9341.rst
kbx81: A find/replace got the better of me. This is fixed and confirmed to work.

glmnet:
Thanks!

PR #874, opened 2020-12-05
Updated Mcp3008 to support reference_voltage and voltage_sampler::VoltageSampler

SenexCrenshaw:
## Description:
Updated the sensor to return the voltage read to match how the ADC component works.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1387

## Checklist:

  - [] Link added in `/index.rst` when creating new documents for new components or cookbook.

This is just an update to the current doc


jesserockz (changes requested):
Please check the example indentation.
Also please don't inflate the example so much. Its best left with the minimum requirements to use the component. eg. don't need the "fridge" config as well. Best if you just add the `reference_voltage` line into the existing example.

SenexCrenshaw:
> Please check the example indentation.
> Also please don't inflate the example so much. Its best left with the minimum requirements to use the component. eg. don't need the "fridge" config as well. Best if you just add the `reference_voltage` line into the existing example.

Sounds good. ill do it over the weekend.

SenexCrenshaw:
Made a few changes and I hope this completed the doc. It was lacking to say the least! Sorry about that

SenexCrenshaw:
@jesserockz I changed the png image to jpg and now the changelog doesnt match throwing a lint error. I could put the file back and have both or do I edit the changelog to point to the right file?

jesserockz:
I think the PNG file is corrupt but because it's still present in the branch it is causing errors. I am only on my phone so can't check properly.

SenexCrenshaw (commented):
Thanks!

SenexCrenshaw (commented):
Looks good, Thanks!
@jesserockz  is this good to go now?

jesserockz (approved):

PR #884, opened 2020-12-12
Adding link to debugger setup for ESP32

seantrue:
## Description: Link to blog post describing ESP32 debugging setup on Ubuntu

**Related issue (if applicable): None

## Checklist:

  - [X ] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [X ] Link added in `/index.rst` when creating new documents for new components or cookbook.


fabaff:
In not so sure if we should include that blog post. platformio says that Python 2.7 is not recommended but the blogs post tells users to install `python2.7` and `libpython2.7`. Python 2.x is EOL for almost a year now.

seantrue:
> In not so sure if we should include that blog post. platformio says that Python 2.7 is not recommended but the blogs post tells users to install `python2.7` and `libpython2.7`. Python 2.x is EOL for almost a year now.

I hear you. It's a work around for dependencies inside the openocd/espressif tool chain. Clearly labeling this an issue in the blog post would be easy. Debugging without the workaround is not ...

PR #892, opened 2020-12-18
Update sonof_basic.rst - incorrect example

bvansambeek:
Board not correct and arduino version not necessary

## Description: The example uses incorrect board type. Got it from the cookbook which gives a good example with light: https://esphome.io/cookbook/sonoff-basic-light-switch.html

## Checklist:

### DIFF
--- a/devices/sonoff_basic.rst
+++ b/devices/sonoff_basic.rst
@@ -54,18 +54,19 @@ exposes all of the basic functions.
     esphome:
       name: <NAME_OF_NODE>
       platform: ESP8266
-      board: esp8285
-      arduino_version: 2.4.2
+      board: esp01_1m
 
     wifi:
       ssid: <YOUR_SSID>
       password: <YOUR_PASSWORD>
 
-    api:
-
     logger:
 
+    api:
+      password: "<YOUR_PASSWORD>"
+
     ota:
+      password: "<YOUR_PASSWORD>"
 
     binary_sensor:
       - platform: gpio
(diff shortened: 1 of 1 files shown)


devices/sonoff_basic.rst
```diff
@@ -54,18 +54,19 @@ exposes all of the basic functions.
...
 
-    api:
-
     logger:
 
+    api:
+      password: "<YOUR_PASSWORD>"
+
     ota:
+      password: "<YOUR_PASSWORD>"
```
jesserockz: I don't see a need to add/change these lines..

jesserockz (changes requested):

devices/sonoff_basic.rst
bvansambeek: Thanks fot your time to look at this minor pull request.

It is not good practice to run those without password, although they are both optional.
Both components have the password included in the example.
If you do not want to add the optional passwordt here, then you should remove them there as well:
https://esphome.io/components/ota.html and https://esphome.io/components/api.html

The format and order of the components correspons with the formating of the 'Create New Node' wizard.

Let me know if you still do not see the need to add/change the api and ota lines and I will adjust accordingly.

jesserockz: It's not that I am against the passwords, but I just did not see a point in adding them here to the sonoff_basic example as it should be the minimal config for how the sonoff basic works, and people should look into each component to learn about them. There are a lot of support questions on discord which are solved by pointing to the docs. =)

bvansambeek: I am not very experienced in pull requests, appologies. Do I need to change it before it can get merged?

jesserockz (approved):

PR #895, opened 2020-12-21
Update apds9960.rst

DrRob:
Added a "troubleshooting" section containing an explanation that some modules require power to the VL pin for gesture sensing to work.

## Description:

Some APDS9960 modules, if wired as described on this page, won't detect gestures.  A "troubleshooting" section is added to explain why and what to do.

## Checklist:


jesserockz (approved):
Thanks

PR #910, opened 2021-01-03
DS1307 real time clock component

badbadc0ffee:
## Description:

initial brief documentation for DS1307 RTC component

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1441

## Checklist:

### DIFF
--- a/components/time.rst
+++ b/components/time.rst
@@ -112,10 +112,10 @@ Configuration variables:
 - **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run at specific intervals using
   a cron-like syntax. See :ref:`time-on_time`.
 
-.. _ds1307-sync_to_rtc_action:
+.. _ds1307-write_action:
 
-``ds1307.sync_to_rtc`` Action
------------------------------
+``ds1307.write`` Action
+-----------------------
 
 This :ref:`Action <config-action>` triggers a synchronization of the current system time to the RTC hardware.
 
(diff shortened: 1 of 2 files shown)


components/time.rst
```diff
@@ -88,6 +88,78 @@ Configuration variables:
...
+
+DS1307 Actions
+--------------
+
+The DS1307 component supports :ref:`actions <config-action>` that can be used to synchronize the RTC hardware and
+the system clock.
+
+.. _ds1307-write_action:
+
+``ds1307.write`` Action
```
glmnet, line 123: we maintain a list of actions (don't ask me why) in the automation pages, a link here would be nice

components/time.rst
```diff
@@ -88,6 +88,78 @@ Configuration variables:
...
+
+.. code-block:: yaml
+
+    on_...:
+      - ds1307.read
+
+      # in case you need to specify the DS1307 id
+      - ds1307.read:
+          id: ds1307_time
+
```
glmnet, line 162: I expected to read more, like how to use this in conjuntion with another time source so this source is used as a fallback mostly.

components/time.rst
badbadc0ffee: sure, n.p.

components/time.rst
badbadc0ffee: I'd like to elaborate a bit more on this once the PR esphome/esphome#1442 has settled: I'd rather give a more complete example once there are elegant ways to synchronize clocks instead of the - kind of improvised - possibilities that now exist in more or less bulky lambda expression.

That would then look similar to this:

```yaml
esphome:
  on_boot:
    then:
      ds1307.read:

time:
  - platform: ds1307
    update_interval: never
  - platform: homeassistant
    on_time_sync:
      then:
        ds1307.write:
```

PR #911, opened 2021-01-03
Update mirabella-genio-bulb.rst to show potential use of GPIO14 instead of GPIO13 for specific monochromatic dimmable globes 

rpimeekle:
## Description:

Added a tip to use GPIO14 for the PWM brightness component after struggling with why nothing worked using GPIO13 (this is for the monochromatic Mirabella Genio 9W WiFi LED Item # 1002333 - as per https://mirabellagenio.net.au/es-%2F-bc-warm-white-specs).  
Tasmota recommendations don't quite work for this globe (as per <https://tasmota.github.io/docs/devices/Mirabella-Genio-Bulb/> under the heading "Dimmable warm white or cool white bulbs").  

Instead I used:
- Module type => Generic (18)
- D5 GPIO14 => PWM1 (37)

## Checklist:

### DIFF
--- a/cookbook/mirabella-genio-bulb.rst
+++ b/cookbook/mirabella-genio-bulb.rst
@@ -120,7 +120,7 @@ Thanks to the `existing work <https://github.com/arendst/Sonoff-tasmota/wiki/Mir
 3.1 Monochromatic Bulbs
 ***********************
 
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Thanks

PR #912, opened 2021-01-04
Update nextion.rst

wernight:
Links to the touch component for the data binding from the touch to ESPHome.

## Description:

Nextion allows 2 way data binding (kind of). This is just a link from one way to the other.

## Checklist:


wernight:
I could move this to be against another branch but there is not invalid to have it added directly to the master.

jesserockz (approved):
Cheers

PR #924, opened 2021-01-09
update time (sync) documentation

badbadc0ffee:
## Description:

**Pull request sin [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1442 / esphome/esphome#1443

## Checklist:

### DIFF
--- a/components/time.rst
+++ b/components/time.rst
@@ -194,7 +194,7 @@ to an external hardware real time clock chip.
     Components should trigger ``on_time_sync`` when they update the system clock. However, not all real time components
     behave exactly the same. Components could e.g. decide to trigger only when a significant time change has been
     observed, others could trigger whenever their time sync mechanism runs - even if that didn't effectively change
-    the system time. Some (such as SNTP) could even trigger when another real time component is responsigle for the
+    the system time. Some (such as SNTP) could even trigger when another real time component is responsible for the
     change in time.
 
 Home Assistant Time Source
--- a/components/time.rst
+++ b/components/time.rst
@@ -189,6 +189,14 @@ to an external hardware real time clock chip.
           then:
             - logger.log: "Synchronized system clock"
 
+.. note::
+
+    Components should trigger ``on_time_sync`` when they update the system clock. However, not all real time components
+    behave exactly the same. Components could e.g. decide to trigger only when a significant time change has been
+    observed, others could trigger whenever their time sync mechanism runs - even if that didn't effectively change
+    the system time. Some (such as SNTP) could even trigger when another real time component is responsigle for the
+    change in time.
+
 Home Assistant Time Source
 --------------------------
 
@@ -227,6 +235,13 @@ Configuration variables:
 
     If your are using :ref:`wifi-manual_ip` make sure to configure a DNS Server (dns1, dns2) or use only IP addresses for the NTP servers.
     
+.. warning::
+
+    Due to limitations of the SNTP implementation, this component will trigger ``on_time_sync`` only once when it detects that the
+    system clock has been set, even if the update was not done by the SNTP implementation!
+    This must be taken into consideration when SNTP is used together with other real time componnents, where another time source could
+    update the time before SNTP synchronizes.
+
 GPS Time Source
 ---------------
 
--- a/components/time.rst
+++ b/components/time.rst
@@ -260,10 +260,10 @@ Configuration variables:
 - **address** (*Optional*, int): Manually specify the I²C address of the RTC. Defaults to ``0x68``.
 - All other options from :ref:`base_time_config`.
 
-.. _ds1307-write_action:
+.. _ds1307-write_time_action:
 
-``ds1307.write`` Action
-***********************
+``ds1307.write_time`` Action
+****************************
 
 This :ref:`Action <config-action>` triggers a synchronization of the current system time to the RTC hardware.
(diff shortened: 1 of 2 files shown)


components/time.rst
```diff
@@ -10,7 +10,183 @@ Time
...
-Currently only SNTP (internet-based), Home Assistant and GPS time sources are supported.
+
+.. _base_time_config:
+
+Base Time Configuration
+-----------------------
+
+All time configuration schemas inherit these options.
+
+Configuration variables:
```
glmnet, line 21:
```suggestion
Configuration variables:
************************
```

components/time.rst
```diff
@@ -10,7 +10,184 @@ Time
...
+
+- **id** (*Optional*, :ref:`config-id`): Specify the ID of the time for use in lambdas.
+- **timezone** (*Optional*, string): Manually tell ESPHome what time zone to use with `this format
+  <https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html>`__ (warning: the format is quite complicated)
+  or the simpler `TZ database name <https://en.wikipedia.org/wiki/List_of_tz_database_time_zones>`__ in the form
+  <Region>/<City>. ESPHome tries to automatically infer the time zone string based on the time zone of the computer
+  that is running ESPHome, but this might not always be accurate.
+- **on_time** (*Optional*, :ref:`Automation <automation>`): Automation to run at specific intervals using
+  a cron-like syntax. See :ref:`time-on_time`.
+- **on_time_sync** (*Optional*, :ref:`Automation <automation>`): Automation to run when the time source
```
glmnet, line 32: Overall PR is ok, but we can beware of terminology better:

Sync is already used, mostly in code, to indicate the ESP clock is updated with some time source, which is what's happening here. I don't understand this short description, wouldn't something like: 

Automation to run after the time source synchronizes the internal clock. See ...

components/time.rst
```diff
@@ -10,7 +10,184 @@ Time
...
+
+.. _time-on_time_sync:
+
+``on_time_sync`` Trigger
+************************
+
+This automation can be used to run synchronize several real time clocks.
+See the :ref:`DS1307 configuration example <ds1307-config_example>` for a scenario
+where a network time synchronization from a home assistant server trigger a write
+to an external hardware real time clock chip.
```
glmnet: Also this explanation can be improved, the example is good and makes sense, but before that I get confused. 

```suggestion
This automation is triggered after a time source successfully retrieves the current time.
See the :ref:`DS1307 configuration example <ds1307-config_example>` for a scenario
where a network time synchronization from a home assistant server trigger a write
to an external hardware real time clock chip.
```

glmnet (commented):
Here are some notes regarding my understanding on the docs, my English isn't too good, but still

components/time.rst
badbadc0ffee: How about `on_time_update`? - That could certainly also be missunderstood because the time is essentially updated every second... The same would apply to `on_time_change`. Maybe `on_time_set`?
Im very open to suggestions!

glmnet: It's ok on sync. I was thinking out loud. Ds1370 has read and write but those are actions. 
It's a read action call causing the sync event trigger which is odd but May be ok, a sync action will be confusing. 
Btw ds1370 can store data? Wouldn't read write be confusing then?
As 1.16 is in beta we can change that before making it a breaking change. Not a request but a comment

badbadc0ffee: You are right: DS1307 has a small (56 bytes) NVRAM section to store data. My idea was to read all data incl. time an NVRAM in one call but it is probably really better to have separate calls to be able to update the data section without an update to the time registers. I'll change that later today.

My proposal for the actions would be  `read_time` and `write_time` to make it clear the these actions only store or retrieve the time information. I'll add an interface for the general purpose registers later.

A `read_time` will indeed trigger an `on_time_sync` event which is intentional: other components can register to be updated whenever the RTC is used to update the system clock.
Optionally, I am still thinking of another wording for the `on_time_sync` event. I think `sync` is consistent, because the event shall be triggered whenever `synchronize_epoch_()`  is called to update the system time.
I'd also be 100% OK to rename it to something like `on_time_set`. Please post your opinions or other options here. Otherwise I'll leve it as it is. ;-)

glmnet: `on_time_sync` ok

components/time.rst
```diff
@@ -10,7 +10,184 @@ Time
...
+This automation is triggered after a time source successfully retrieves the current time.
+See the :ref:`DS1307 configuration example <ds1307-config_example>` for a scenario
+where a network time synchronization from a home assistant server trigger a write
+to an external hardware real time clock chip.
+
+    .. code-block:: yaml
+
+        on_time_sync:
+          then:
+            - logger.log: "Synchronized system clock"
```
glmnet, line 191:
```suggestion

.. note::

    The SNTP component will trigger this event only once, the first time any time component synchronizes. This works well when SNTP is the only time component, but if you another time source updating the time before SNTP synchronizes, that event will be triggered.

```

glmnet (commented):
do you agree with this explanation here?

badbadc0ffee:
> do you agree with this explanation here?

I do! I'll just split it and put a note here and another one to the SNTP documentation.

PR #926, opened 2021-01-09
Update st7789v.rst

deftdawg:
The original ST7786V component work by folks outside of ESPHome included RGB565 (2 byte pixels), however that customization to 'Image' was never merged into ESPHome, so ESPHome Image only supports RGB24 (3 byte pixels); as such the examples on this page fail validation.  This change replaces RGB565 with RGB24.

## Description:

## Checklist:


kbx81:
Looks like this has already been taken care of -- closing

PR #929, opened 2021-01-10
Update spi.rst

NdS-Research-Facilities:
## Description:
updates sample as per https://discord.com/channels/429907082951524364/429907082955718657/797861820831760414

## Checklist:

### DIFF
--- a/custom/spi.rst
+++ b/custom/spi.rst
@@ -11,21 +11,23 @@ them publish values.
 Please refer to the SPI library docs for more information.
 
 .. code-block:: cpp
+#include "esphome.h"
 
-    #include "esphome.h"
 
-    class MyCustomComponent : public Component {
-     public:
-      void setup() override {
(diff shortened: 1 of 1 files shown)


custom/spi.rst
```diff
@@ -11,21 +11,23 @@ them publish values.
...
-      }
-    };
+#include "esphome.h"
+
+
+class MyCustomComponent : public Component, public spi::SPIDevice<spi::BIT_ORDER_MSB_FIRST, spi::CLOCK_POLARITY_LOW,
+                                                 spi::CLOCK_PHASE_LEADING, spi::DATA_RATE_2MHZ>{
+ public:
+  void setup() override {
+
```
mmakaay, line 21: I'm not a fan of the way in which the methods are formatted here.
I would suggest to have a white line between `setup()` and `loop()` and get rid of the empty line at the start of the method. That makes it a lot more clear what goes together.

PR #930, opened 2021-01-10
Update pzemac.rst due to wrong PIN reference

emeggiolaro:
## Description:

## Checklist:

### DIFF
--- a/components/sensor/pzemac.rst
+++ b/components/sensor/pzemac.rst
@@ -35,8 +35,8 @@ to some pins on your board and the baud rate set to 9600.
 
     # Example configuration entry
     uart:
-      rx_pin: D1
-      tx_pin: D2
+      rx_pin: GPIO3
+      tx_pin: GPIO1
       baud_rate: 9600
 
     sensor:


jesserockz (approved):
Example pins are just examples. A lot of boards and devices have different pins for different functions. 
All though I do somewhat agree that maybe we should use the `GPIOx` variant everywhere and not the board specific pins `Dx` etc.

@glmnet Thoughts? Should we enforce `GPIOx` instead of `Dx` for docs?

glmnet:
yes, I'd rather use `GPIOxx` instead of `Dx` however this PR is not replacing just PIN names, it's actually using default RX TX pins (on both ESP8266 and ESP32) which are not default in given example, in this case I would just remove the `TX:` `RX:` config lines as they are default and we do not put defaults in examples.

Using default uart pins will mean you needing to disable serial logging, so this PR should remove RX and TX and add `baud_rate: 0` to `logger:`, otherwise the sample is OK

RELATED: We could add comments indicating `# ESP8266` or `# ESP32` pin setup, most lovely would be to have a button to view samples for one device or another like some tutorials sites they show you javascript / typescript etc.

components/sensor/pzemac.rst
```diff
@@ -35,8 +35,8 @@ to some pins on your board and the baud rate set to 9600.
 
     # Example configuration entry
     uart:
-      rx_pin: D1
-      tx_pin: D2
+      rx_pin: GPIO3
+      tx_pin: GPIO1
```
frenck, line 39:
```suggestion
```

components/sensor/pzemac.rst
```diff
@@ -35,8 +35,8 @@ to some pins on your board and the baud rate set to 9600.
...
       baud_rate: 9600
```
frenck, line 40:
```suggestion
      baud_rate: 9600
      
    logger:
      baud_rate: 0
```

frenck (changes requested):
Added 2 review suggestions to this PR, as per review comments by @glmnet.

@emeggiolaro, could you take a look? Thanks already! 👍 


../Frenck 🚂

PR #943, opened 2021-01-16
Update adc.rst

GeekVisit:
## Description:
Clarification on use of multiple ADC sensors. 

## Checklist:

### DIFF
--- a/components/sensor/adc.rst
+++ b/components/sensor/adc.rst
@@ -8,7 +8,7 @@ Analog To Digital Sensor
 The Analog To Digital (``adc``) Sensor allows you to use the built-in
 ADC in your device to measure a voltage on certain pins. On the ESP8266
 only pin A0 (GPIO17) supports this. On the ESP32 pins GPIO32 through
-GPIO39 can be used.
+GPIO39 can be used. 
 
 .. figure:: images/adc-ui.png
     :align: center
@@ -51,6 +51,8 @@ Configuration variables:
             filters:
               - multiply: 3.3
 
+
+
 .. _adc-esp32_attenuation:
 
 ESP32 Attenuation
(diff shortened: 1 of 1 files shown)


components/sensor/adc.rst
```diff
@@ -8,7 +8,7 @@ Analog To Digital Sensor
 The Analog To Digital (``adc``) Sensor allows you to use the built-in
 ADC in your device to measure a voltage on certain pins. On the ESP8266
 only pin A0 (GPIO17) supports this. On the ESP32 pins GPIO32 through
-GPIO39 can be used.
+GPIO39 can be used.
```
fabaff:
```suggestion
GPIO39 can be used.
```

components/sensor/adc.rst
```diff
@@ -88,6 +90,13 @@ To measure the VCC voltage, set ``pin:`` to ``VCC`` and make sure nothing is con
         pin: VCC
         name: "VCC Voltage"
 
+Multiple ADC Sensors
+---------------------
+
+You can only use as many ADC sensors as your device can support.  The Esp8266 only has one ADC and can only handle one sensor at a time. For example, on the ESP8266, you can measure the value of an analog pin (A0 on ESP8266) or VCC (see below) but NOT both simultaneously. Using both at the same time will result in incorrect sensor values.
```
fabaff:
```suggestion
You can only use as many ADC sensors as your device can support. The ESP8266 only has one ADC and can only handle one sensor at a time. For example, on the ESP8266, you can measure the value of an analog pin (A0 on ESP8266) or VCC (see below) but NOT both simultaneously. Using both at the same time will result in incorrect sensor values.
```

fabaff (dismissed):

frenck (approved):
Thanks, @GeekVisit 👍 

../Frenck 🚂

PR #947, opened 2021-01-17
Add IWOOLE Table Lamp cookbook entry

Deinara:
## Description:
I would like to add a cookbook entry for this table lamp. I've added some images of my experience and how to connect a serial programmer.

## Checklist:


Deinara:
I have tried adding a blank line after the figures, but I can't seem to make the linter warning go away

Deinara:
It took a while to figure out how to resolve all the linter warnings and how reStructuredText works, but the page is now in a state where I'm happy with it.
I hope this page will be useful to some.

cookbook/iwoole_rgbw_table_lamp.rst
```diff
@@ -0,0 +1,174 @@
...
+        ssid: "IWOOLE Table Lamp Fallback Hotspot"
+        password: "safefallbackpassword"
+    captive_portal:
+    logger:
+    api:
+      password: "ialwaysforgetthisone"
+    ota:
+      password: "enteryourownpasswordhere"
+
+
```
jesserockz: I think this would be the only thing I would change/remove as most people will be starting from scratch and know what they are doing, or using the wizard which generates this part.

```suggestion
```

jesserockz (changes requested):

cookbook/iwoole_rgbw_table_lamp.rst
Deinara: Yeah, I can see how that would remove some clutter. Thanks for the suggestion.

jesserockz (approved):

PR #949, opened 2021-01-19
Fix human-readable uptime example so it compiles

oddsockmachine:
## Checklist:


fabaff (approved):
Thanks

PR #966, opened 2021-01-28
Adding that mpr121_id is a valid option for binary_sensor

minideezel:
## Description:

I had to figure out how to have multiple MPR121's defined, figured I would document the fix so it would be easier for the next guy. Let me know if I did something wrong, this is my first contribution to the docs.

**Related issue (if applicable):** fixes N/A

## Checklist:

### DIFF
--- a/components/binary_sensor/mpr121.rst
+++ b/components/binary_sensor/mpr121.rst
@@ -32,7 +32,6 @@ required to be set up in your configuration for this sensor to work.
 
     binary_sensor:
       - platform: mpr121
-        mpr121_id: mpr121_component
         id: touch_key0
         channel: 0
         name: "Touch Key 0"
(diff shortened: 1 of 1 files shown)


components/binary_sensor/mpr121.rst
```diff
@@ -32,6 +32,7 @@ required to be set up in your configuration for this sensor to work.
 
     binary_sensor:
       - platform: mpr121
+        mpr121_id: mpr121_component
```
glmnet: This is ok however we don't clutter the docs first example so better this should not go here 

Btw a new check in esphome will make it fail to compile if you declare two hubs and don't specify the one you want to use. So future people will have a better time than you. 

The definition below is ok

glmnet (changes requested):
Remove line then good to merge

minideezel:
Updated!

glmnet:
Thanks

PR #982, opened 2021-02-03
Updated documentation for the SGP30 sensor improvements.

### LINKED ISSUE #668: SGP30: report "baseline" as a sensor as well as it's already done in the log

After evaluating CCS811 vs SGP30, I decided to use the SGP30 for my Portable Air Quality useful toy.
The SGP30 re-calibrates its baseline value frequently, as reported in the ESP_LOGI() but it is reset at every boot.

The quality of the SGP30 returned values greatly depends on a baseline calibration updated as much as possible. The manufacturer suggests every 7 days, but initially (in my experience for the first couple of months or so) it should be updated almost daily.
It would be nice to have the SGP30 baseline value reported as a sensor, so one can periodically save the last value and then restore it after the next boot (which in my case is frequent, as the wifi is weak and the connection is lost frequently).

**Additional context**

As my build can also work offline from my home HASSIO server (i.e. at my sons or nephews home), it would be very nice for me to be able to save it also locally as a `globals: restore_value: yes` if some `save_globals` action could be called (not often!) in some way. Otherwise I will need to bind the Air Quality ESP32 probe with a dedicated RasPI-Zero, only to have a "portable" HassIO server.

### LINKED ISSUE #673: SGP30: update_interval should be 1s, otherwise raise a log warning message

As per SGP30 datasheet:
"The on-chip baseline compensation algorithm has been optimized for this sampling rate (1 Hz). The sensor shows best performance when used with this sampling rate."

With update_interval 10s the SGP30 shows strange behaviours, such as swap eCO2/tVOC baselines and/or return only one of the two real values.

**Additional context**

The ESPHome documentation may also suggest to use median and/or SWMA filters to slow down the update frequency, while obtaining more mediated values.

huubeikens:
## Description:

Updated the SGP30 sensor with configuration options to persistently store the eCO2/TVOC baselines so that it survives a power cycle. It removes the need to recompile and OTA the configuration file when the baselines have been determined after the calibration period (of 12 hours).

As a convenience, the baseline values can also be configured as sensors so that they are published to Home assistant.

Last, the update interval default is set to 1 second which is advised by the datasheet.

closes esphome/feature-requests#673
closes esphome/feature-requests#668

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1510

## Checklist:

### DIFF
--- a/components/sensor/sgp30.rst
+++ b/components/sensor/sgp30.rst
@@ -24,8 +24,9 @@ The :ref:`I²C Bus <i2c>` is required to be set up in your configuration for thi
         tvoc:
           name: "Workshop TVOC"
           accuracy_decimals: 1
+        store_baseline: yes
         address: 0x58
-        update_interval: 5s
+        update_interval: 1s
 
 Configuration variables:
 ------------------------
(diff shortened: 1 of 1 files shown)


jesserockz:
Could you please rebase your changes from the `next` branch and then change the target of this PR

huubeikens:
Done rebasing. Since I'm not an experience git user (yet), I'm a bit unsure about the mentioned branch conflicts. Please let me know if you need me to do some fixing.

jesserockz (approved):

PR #991, opened 2021-02-07
Fix st7735 yaml example

rradar:
## Description:

Fix st7735 yaml example

## Checklist:


fabaff (approved):
Thanks

PR #992, opened 2021-02-08
Fix API Reference link for display buffer

zuzu59:
## Description:

The link pointing to API Reference was wrong


fabaff (approved):
Thanks

PR #995, opened 2021-02-10
Add sleep duration to enter deep sleep action

nuttytree:
## Description:

**Related issue (if applicable):** fixes esphome/feature-requests#339

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1526

## Checklist:


glmnet:
So if I use this action with a time that will permanently override the configured run duration on the component right?
May be you can add a warning about that

nuttytree:
It actually will not permanently change the duration because upon coming out of deep sleep it will re-initialize the component putting it back to what is configured in the yaml.

PR #998, opened 2021-02-13
Fix teleinfo

schmurtzm:
## Description:
PR about teleinformation sensor :
 - Small correction on the example.
 - add some recommandations to avoid crash due to hardware UART conflicts with Logger.


components/sensor/teleinfo.rst
```diff
@@ -60,17 +60,17 @@ simply press -/+ buttons on the counter and look for `Standard mode` or
     sensor:
       - platform: teleinfo
         tags:
-         - name: "HCHC"
+         - tag_name: "HCHC"
```
0hax, line 63: This `tag_name` issue has already been fixed in #871

components/sensor/teleinfo.rst
```diff
@@ -97,6 +97,34 @@ Configuration variables:
 - **uart_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the :ref:`UART Component <uart>` if you want
   to use multiple UART buses.
 
+
+.. warning::
+
+    UART hardware is recommanded for this sensor. The default Esphome's logger also use one hardware UART.
+    The ESP8266 only has two hardware UARTs, one of which is transmit-only (details in `Logger Component section <components/logger>`__).
```
0hax, line 104: I would say something like:
```
On ESP8266, only 2 UARTs are available, one of it is TX only. This means only one UART is available for Teleinformation as RX is needed.
Since the only UART available for Teleinformation is already used by the logger, we need to either set the logger to use the second UART which is TX only or totally disable the logger.
A third solution would be to use another pin not dedicated to UART by the mean of Software UART, but unfortunately, this has been very unstable during development and it's not recommended.
To disable logger or make it use another UART, you can have a look to <path/to/uart/wiki>
```
We discussed it in #855

components/sensor/teleinfo.rst
```diff
@@ -97,6 +97,34 @@ Configuration variables:
...
+
+- Disable the logging via UART :
+    
+.. code-block:: yaml
+
+    logger:     
+      baud_rate: 0   # disable logging via UART, help to avoid numerous crash with ESP_LOGD
+      level: INFO   # INFO for less log, put DEBUG to view all the linky's "étiquettes" received in the logs
+      esp8266_store_log_strings_in_flash: False     #  :doc:`recommanded for ESP8266 </components/sensor/custom>`
+
```
0hax, line 117: I'm fine adding some examples on how to disable UART logging like you did but this is subject to changes so maybe it's better to simply reference the uart wiki page and let people figure out how to disable uart logging from there?

components/sensor/teleinfo.rst
```diff
@@ -4,7 +4,7 @@ Teleinformation from Linky electrical counter.
 .. seo::
     :description: Instructions for setting up French Teleinformation
     :image: teleinfo.jpg
-    :keywords: teleinfo
+    :keywords: teleinfo, linky
```
0hax, line 7: Good idea!

0hax (commented):
I have added few review comments inline. 
Let me know what you think.

components/sensor/teleinfo.rst
schmurtzm: Oh! I didn't saw your PR because I was looking for something from @djtef ...
So yes, this include this little correction too.

components/sensor/teleinfo.rst
schmurtzm: Yes we could. But I think that the 2 examples can make gain a lot of time to the users. The part of the documentation about the uart is not the most simple to handle for beginners.

PR #1018, opened 2021-02-19
Update esphome-configs URL

shish:
For some reason my browser sees an I/O timeout when trying to load esphome-configs.io - and when I use curl, it shows that the URL has changed. When linking directly to the new site, it works fine in my browser.
```
$ curl esphome-configs.io
<a href='http://esphome-devices.com'>Moved Permanently</a>.
```


glmnet:
Ok seems they moved. Thanks

PR #1039, opened 2021-03-05
Uart invert option for ESP32

Mynasru:
## Description:
Option to invert the logic levels of the RX and TX pins.

**Related issue (if applicable):** fixes [1145](https://github.com/esphome/feature-requests/issues/1145)

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1586

## Checklist:


glmnet (changes requested):
To better separate the settings from esp8266 please add an interline with the legend

ESP32 Options:

See for example: https://deploy-preview-1039--esphome.netlify.app/components/esphome.html where `esp8266_restore_from_flash` is documented.

Thanks

glmnet (approved):

PR #1043, opened 2021-03-07
Update i2c.rst

webeling67:
Clarify that example config is for ESP32

## Description:
I felt in the trap to assume the example config would work for ESP8266, hence the little change in the example comment

## Checklist:


glmnet:
It's ok. Actually i2c minimum config is just

```
i2c:
```

Then default pins will be selected

PR #1046, opened 2021-03-08
Specify format for BSSID entries.

marcteale:
## Description:
Clarify format of BSSID entry.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#
n/a

## Checklist:

### DIFF
--- a/components/wifi.rst
+++ b/components/wifi.rst
@@ -176,8 +176,8 @@ Configuration variables:
 - **eap** (*Optional*): See :ref:`eap`.
 - **channel** (*Optional*, int): The channel of the network (1-14). If given, only connects to networks
   that are on this channel.
-- **bssid** (*Optional*, string): Optionally define a BSSID (MAC-Address) of the network to connect to.
-  This can be used to further restrict which networks to connect to.
+- **bssid** (*Optional*, string): The connection's BSSID (MAC address). BSSIDs must consist of six
+  two-digit hexadecimal values separated by colon characters ("``:``"). All letters must be in upper case.
 - **hidden** (*Optional*, boolean): Whether this network is hidden. Defaults to false.
   If you add this option you also have to specify ssid.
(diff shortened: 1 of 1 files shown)


jesserockz (commented):
Do we really need this? Especially without actually giving an example?
ESPHome will give a config validation error on trying to compile with an incorrectly formatted mac address anyway and also give the example of `XX:XX:XX:XX:XX:XX`

marcteale:
Either the configuration validation needs to be updated, or the documentation, or both.  While the validation rejects an improperly formatted MAC address (e.g. a hyphen-delimited entry), it accepts lowercase letters.  Firmware compiled with a lowercase entry will not be able to find the specified BSSID.  I also feel it's poor UX if the only way to determine the correct format is to receive a compile time error.

That said, if the edit would benefit from an example, I'm happy to make the change.

jesserockz:
Thats unfortunate. If that is the case, then the config validation should be enforcing or converting to uppercase.

jesserockz:
So I just tested the config, and it converts them to hex integers, and the main.cpp file is no different if you use upper or lower case...

marcteale:
I just did some testing of my own and got the same results.  I think the connection failures that spawned this pull request were the result of a dying/flaky WAP.

frenck:
Looking at the discussion above, and the contents of this PR:

 @marcteale If you really think this is an issue, please raise an issue in the issue tracker, so we can keep track of it and discuss it until resolved. A documentation PR might not be the right place to track it :)

Personally, as for the contents of this PR, I think it doesn't harm to add it. At best, people know what to do before being hit by the config validation. So I would vote on adding it. @jesserockz What do you think?

frenck (approved):

jesserockz (approved):
Ok, lets get this in then =)

PR #1050, opened 2021-03-10
add font to usage example

azrael783:
## Description:
Added the font part, because without this part I could not upload to sample configuration to the ESP8266. As a beginner it was not directly clear for me what to add due to the error message I got in the UI. 

## Checklist:

### DIFF
--- a/components/display/waveshare_epaper.rst
+++ b/components/display/waveshare_epaper.rst
@@ -49,6 +49,11 @@ configuration.
 .. code-block:: yaml
 
     # Example configuration entry
+    font:
+      - file: 'some_font_file.ttf' # place the "some_font_file.ttf" in /config/esphome
+        id: font1
+        size: 8
+    
     spi:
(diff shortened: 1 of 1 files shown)


components/display/waveshare_epaper.rst
```diff
@@ -49,6 +49,11 @@ configuration.
 .. code-block:: yaml
 
     # Example configuration entry
+    font:
+      - file: 'some_font_file.ttf' # place the "some_font_file.ttf" in /config/esphome
```
frenck: Suggestion to use an actual font name, and removing the comment (as the shown folder is very specific to a certain installation method and thus not universal).

```suggestion
      - file: 'fonts/Comic Sans MS.ttf'
```

frenck (changes requested):

frenck (approved):
Thanks, @azrael783 👍 

../Frenck 🚂

PR #1053, opened 2021-03-11
Wifi scan results

gpambrozio:
## Description:

Adds documentation to the new text sensor that adds the latest wifi can results.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1605

## Checklist:


gpambrozio:
Opened #1054 based on `next` instead of `current`

PR #1056, opened 2021-03-14
Documentation for fan speed levels

blejdfist:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1541
## Checklist:


components/fan/speed.rst
```diff
@@ -30,15 +30,7 @@ Configuration variables:
...
-- **speed** (*Optional*): Set the float values for each speed setting:
-
-  - **low** (**Required**, float): Set the value for the low speed
-    setting. Must be in range 0 to 1. Defaults to 0.33.
-  - **medium** (**Required**, float): Set the value for the medium speed
-    setting. Must be in range 0 to 1. Defaults to 0.66.
-  - **high** (**Required**, float): Set the value for the high speed
-    setting. Must be in range 0 to 1. Defaults to 1.
-
+- **speed_levels** (*Optional*): Set the number of supported speed levels. Defaults to 100.
```
glmnet:
```suggestion
- **speed_levels** (*Optional*, int): Set the number of supported discrete speed levels. Each step will
  represent a calculated percentage value to the output of the fan. E.g. ``2`` will send 50% and
  100% while ``100`` will allow 1% increments in the output. Defaults to ``100``.
```

components/fan/index.rst
```diff
@@ -7,14 +7,13 @@ Fan Component
 
 With the ``fan`` domain you can create components that appear as fans in
 the Home Assistant frontend. A fan can be switched ON or OFF, optionally
-has a speed setting (``LOW``, ``MEDIUM``, ``HIGH``) and can have an
-oscillate output.
+has a speed level (1..max_supported_speed_level) and can have an
```
glmnet:
```suggestion
has a speed level between 1 and the maximum supported speed level of the fan, and can have an
```

max_supported_speed_level sound like a configuration setting which doesn't exists

glmnet (commented):
I think these updates will make users understand easier what speed_levels are

components/fan/speed.rst
blejdfist: This one is a bit confusing, but I know what you are aiming at. I'll try to rephrase it later after work.

glmnet (approved):

PR #1057, opened 2021-03-15
thermostat min/max temperature clarification

larsonmpdx:
## Description:

**Related issue (if applicable):** see discussion in https://github.com/esphome/issues/issues/1507

## Checklist:

### DIFF
--- a/components/climate/thermostat.rst
+++ b/components/climate/thermostat.rst
@@ -153,7 +153,7 @@ These temperatures are used when the device first starts up.
 **At least one of** ``default_target_temperature_low`` **and** ``default_target_temperature_high``
 **must be specified.**
 
-``min_temperature`` and ``max_temperature`` from the base climate component **will limit the range**
+``min_temperature`` and ``max_temperature`` from the base climate component **are used to define the range**
 of allowed temperature values in the thermostat component.  See :doc:`/components/climate/index`
 
 Heating and Cooling Actions
@@ -272,8 +272,8 @@ Advanced Options
 
     While this platform uses the term temperature everywhere, it can also be used to regulate other values.
     For example, controlling humidity is also possible with this platform.  Note that ``min_temperature`` and
-    ``max_temperature`` from the base climate component will limit the range of adjustability and the defaults
-    will probably not make sense for control of things like humidity.  See :doc:`/components/climate/index`
+    ``max_temperature`` from the base climate component are used the define the range of adjustability and the
+    defaults will probably not make sense for control of things like humidity.  See :doc:`/components/climate/index`
 
 Bang-bang vs. Thermostat
 ------------------------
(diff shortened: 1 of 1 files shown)


glmnet:
@kbx81 could you review this?

kbx81 (approved):
LGTM

components/climate/thermostat.rst
```diff
@@ -153,6 +153,9 @@ These temperatures are used when the device first starts up.
 **At least one of** ``default_target_temperature_low`` **and** ``default_target_temperature_high``
 **must be specified.**
 
+``min_temperature`` and ``max_temperature`` from the base climate component **will limit the range**
+of allowed temperature values in the thermostat component.  See :doc:`/components/climate/index`
+
```
kbx81, line 158: I guess the only additional comment I would have is to perhaps add a sentence to address the people who ask "why"? For example:
_"These values are used to ensure any given set point is **always** within the desired range -- you wouldn't  want a broken automation in HA to set your living room's thermostat to -999999 degrees in the middle of winter, would you?"_

Totally optional and absolutely not necessary...but...we're engineers...and inevitably people jump on the discord server and ask "why?"

larsonmpdx: thanks see update

components/climate/thermostat.rst
```diff
@@ -153,6 +153,9 @@ These temperatures are used when the device first starts up.
 **At least one of** ``default_target_temperature_low`` **and** ``default_target_temperature_high``
 **must be specified.**
 
+``min_temperature`` and ``max_temperature`` from the base climate component **are used to define the range**
+of allowed temperature values in the thermostat component.  See :doc:`/components/climate/index`
```
frenck: Use a single space after periods :)

```suggestion
of allowed temperature values in the thermostat component. See :doc:`/components/climate/index`
```

See also: <https://docs.microsoft.com/en-us/style-guide/punctuation/periods>

components/climate/thermostat.rst
```diff
@@ -268,7 +271,9 @@ Advanced Options
 .. note::
 
     While this platform uses the term temperature everywhere, it can also be used to regulate other values.
-    For example, controlling humidity is also possible with this platform.
+    For example, controlling humidity is also possible with this platform.  Note that ``min_temperature`` and
+    ``max_temperature`` from the base climate component are used the define the range of adjustability and the
+    defaults will probably not make sense for control of things like humidity.  See :doc:`/components/climate/index`
```
frenck:
```suggestion
    For example, controlling humidity is also possible with this platform. Note that ``min_temperature`` and
    ``max_temperature`` from the base climate component are used the define the range of adjustability and the
    defaults will probably not make sense for control of things like humidity. See :doc:`/components/climate/index`
```

frenck (approved):
Thanks, @larsonmpdx 👍 

../Frenck 🚂

PR #1061, opened 2021-03-17
Add output part to binary light example

klaasnicolaas:
## Description:

This adds the output part to the binary light example. This would make the example more complete and potentially avoid some frustration.

On request by @balloob 😜 

## Checklist:

### DIFF
--- a/components/light/binary.rst
+++ b/components/light/binary.rst
@@ -20,6 +20,11 @@ The ``binary`` light platform creates a simple ON/OFF-only light from a
         name: "Desk Lamp"
         output: output_component1
 
+    output:
+      - id: light_output
+        platform: gpio
+        pin: GPIO16
+
 Configuration variables:
 ------------------------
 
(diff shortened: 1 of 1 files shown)


balloob (approved):

components/light/binary.rst
```diff
@@ -20,6 +20,11 @@ The ``binary`` light platform creates a simple ON/OFF-only light from a
         name: "Desk Lamp"
         output: output_component1
 
+    output:
+      - id: light_output
```
jesserockz, line 24:
```suggestion
      - id: output_component1
```

jesserockz (changes requested):
Better make sure the ids match then 👍

components/light/binary.rst
balloob: In that case, maybe better to update the name of the light config?

jesserockz (approved):

PR #1074, opened 2021-03-23
Update st7735.rst

wjcarpenter:
Corrects some mis-statements in the component documentation. It didn't make any sense to say a config item is required and then to say it has a default. It would be better to make those config items optional in the first place.

Also adds the implemented but undocumented "use_bgr" config item.

Thanks for your work in bringing this display controller into esphome.

## Description:

## Checklist:

### DIFF
--- a/components/display/st7735.rst
+++ b/components/display/st7735.rst
@@ -6,7 +6,7 @@ ST7735 Display
     :keywords: ST7735
     :image: ST7735.png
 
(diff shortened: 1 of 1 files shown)


wjcarpenter:
Sorry, made my PR against the wrong branch. I don't know how to change from "current" to "next" without re-creating the PR.

PR #1076, opened 2021-03-24
SPI: fix Example

ferbar:
## Description:

The right pins for ESP8266 and ESP32 would be GPIO12-GPIO14.
GPIO21-23 is not available on an ESP8266, on an ESP32 it is I2C, maybe it was copy&paste from I2C?

## Checklist:


fabaff (approved):
Tanks

PR #1083, opened 2021-04-04
Add the documentation for the touch screen GUI component

numo68:
## Description:

This pull request adds the documentation for a way to create a touch screen GUI where a set of virtual buttons for multiple display pages can be created. The touch screen events such as ones from the XPT2046 touch screen controller can be routed to this component where the interaction with the buttons is evaluated and provided in the binary sensor way.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1660

## Checklist:


numo68:
The parent was resurrected so I guess this should be left open as well...

PR #1086, opened 2021-04-05
Adds docs for the b-parasite soil moisture sensor

rbaron:
## Description:
[b-parasite](https://github.com/rbaron/b-parasite) is an open hardware and open source soil moisture and ambient temperature/humidity sensor. It broadcasts its sensors' readings via Bluetooth Low Energy (BLE).

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1666

## Checklist:


components/sensor/b_parasite.rst
```diff
@@ -0,0 +1,59 @@
...
+        moisture:
+          name: 'b-parasite Soil Moisture'
+        battery_voltage:
+          name: 'b-parasite Battery Voltage'
+
+Configuration variables
+-----------------------
+
+- **mac_address** (**Required**): The MAC address of the device.
+- **temperature** (**Optional**): Air temperature in Celcius.
```
OttoWinter:
```suggestion
- **temperature** (*Optional*): Air temperature in Celcius.
```

We only use bold for `Required` (if it's a different way somewhere else then that's wrong)

components/sensor/b_parasite.rst
```diff
@@ -0,0 +1,59 @@
...
+
+  - **name** (**Required**): Sensor name.
```
OttoWinter, line 43: Please add the standard line `- All other options from :ref:`Sensor <config-sensor>`.` here too (same for other sensor blocks)

rbaron: Done. Thanks!

components/sensor/b_parasite.rst
rbaron: Done.

rbaron:
Thanks once again @OttoWinter!

PR #1093, opened 2021-04-08
Add the documentation for the autorepeat filter

numo68:
## Description:

A documentation for a common autorepeat filter able to produce binary sensor changes according to a timing description

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1681

## Checklist:

### DIFF
--- a/components/binary_sensor/index.rst
+++ b/components/binary_sensor/index.rst
@@ -78,13 +78,20 @@ of these entries matters!)
           - delayed_on: 100ms
           - delayed_off: 100ms
           - delayed_on_off: 100ms
+          - autorepeat:
+            - delay: 1s
+              time_off: 100ms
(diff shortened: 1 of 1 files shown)


glmnet (approved):

glmnet:
I like this, works great, also a great solution to push buttons to e.g. increase brightness

Isn't a bit odd that you have to provide at least one item, even when all parameters have defaults?

numo68:
> I like this, works great, also a great solution to push buttons to e.g. increase brightness

Thanks

> Isn't a bit odd that you have to provide at least one item, even when all parameters have defaults?

I added a "shortcut" where a filter with no timings is equivalent to one timing with all the parameters default.

PR #1102, opened 2021-04-15
pmsx003: add config options for extra sensor data points

warthog9:
## Description:

This adds in the config options for pm1.0, pm2.5, pm10 under standard
particles, as well as the partcile counts beyond 0.3um, 0.5um, 1.0um,
2.5um, 5.0um, and 10.0um

none

esphome/esphome#1694

## Checklist:

### DIFF
--- a/components/sensor/pmsx003.rst
+++ b/components/sensor/pmsx003.rst
@@ -40,11 +40,29 @@ value:
 Configuration variables:
 ------------------------
 
-- **pm_1_0** (*Optional*): Use the concentration of particulates of size less than 1.0µm in µg per cubic meter.
+- **pm_1_0_std** (*Optional*): Use the concentration of particulates of size less than 1.0µm in µg per cubic meter at standard particle
   All options from :ref:`Sensor <config-sensor>`.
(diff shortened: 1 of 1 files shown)


glmnet:
Can you expand the top section? Is not made clear all these options which devices support them (there are 3 devices)

warthog9:
At least for the x003 sensors the only one who's data sheet doesn't list all of those is the 3003, and it has some reserved fields in the data sheet's noted response that would match up to a particle count >1.0um, but not 5.0, or 10 before it gets to the as the PMS3003 seems to put the checksum in the data block where the >5.0um would be.  Happy to note that if it's wanted.

As for the PMS5003T this seems odd since (according to the code, as I can't find a data sheet for it oddly and don't have one) it reads the temp/humidity out of bytes 24 and 26, which would also be the >5.0um and >10.0um slots so to speak.  I have also found reference to a PMS5003S, which seems to use the same format as the PMS5003T but reporting different values in bytes 24 and 26 seemingly.

The PMS5003ST according to the data sheet (now that I've found a datasheet for it) uses the same response format as the normal PMS5003 but with the 3 extra measurements at the end so I likely need to go back and update my code patch that it too has those fields in it.

Would trailing parenthetical notes for each option work to denote the differences, or would you like that noted another way?

PR #1112, opened 2021-04-22
Add extra margin at the bottom to avoid overlap with Changelog banner

manutenfruits:
## Description:
This is an infrastructure change that adds some margin at the bottom of the document, so the changelog banner doesn't overlap with the content. The only way of getting to the content is by hitting "View changelog" button which opens a new page, and isn't a great experience if all you want to do is read the docs.

I tried to find the best spot to add it, and use a sensible px number.

Before: 
localhost_8000_components_esphome html (1)

After: 
localhost_8000_components_esphome html

## Checklist:

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -5,6 +5,10 @@
   text-align: center;
 }
 
+div.document {
+  margin-bottom: 50px;
+}
+
 img.component-image {
   border: none;
   vertical-align: middle;


glmnet:
I just merged your other commit, is this going to be compatible?

manutenfruits:
Yes, they are unrelated. Even if you can dismiss the footer, it shouldn't be covering the bottom of the page.
Technically speaking, they are also compatible, as the margin is added at the bottom of the page.

frenck (approved):
LGTM ✅  Thanks, @manutenfruits!

../Frenck 🚂

PR #1113, opened 2021-04-22
Make changelog banner dismissable

manutenfruits:
## Description:
This change adds a "Dismiss" button to the changelog update footer, as not all users might be interested in checking it out. Clicking anywhere on the footer makes it disappear and updates Local Storage so it's permanent until next release.

I used a `<div role="button">` so it's a bit more accessible, and so I don't have to override too many styles, which I would have needed to do if I used a real `<button>`. Moved some styles to classes, as that's industry standard.

Desktop:
localhost_8000_components_esphome html (2)

Mobile:
localhost_8000_components_esphome html(Pixel 2)

## Checklist:

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -62,15 +62,18 @@ table.docutils {
   z-index: 9999;
   display: none
 }
-#upgrade-footer-changelog {
+.footer-button-container {
+  margin: 0 60px 0 10px;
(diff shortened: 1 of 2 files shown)


glmnet:
Thanks!

glmnet (approved):

PR #1117, opened 2021-04-26
Expand uart invert feature to ESP8266

Lewn:
## Description:

Expands the ``invert`` option added for the ESP32 in #1586 to the ESP8266.

~As the esp8266 arduino library only implements this for UART0, a combination of ``invert`` and ``tx_pin: GPIO2`` will fall back to using software UART.~ see below

To also support different pin combinations, a the inverting behaviour is also implemented for ``ESP8266SoftwareSerial``.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1727

## Checklist:

### DIFF
--- a/components/uart.rst
+++ b/components/uart.rst
@@ -51,17 +51,13 @@ Configuration variables:
 ------------------------
 
 - **baud_rate** (**Required**, int): The baud rate of the UART bus.
-- **tx_pin** (*Optional*, :ref:`config-pin`): The pin to send data to from the ESP's perspective.
-- **rx_pin** (*Optional*, :ref:`config-pin`): The pin to receive data on from the ESP's perspective.
+- **tx_pin** (*Optional*, :ref:`config-pin`): The pin to send data to from the ESP's perspective. Use the full pin schema and set ``inverted: true`` to invert logic levels.
+- **rx_pin** (*Optional*, :ref:`config-pin`): The pin to receive data on from the ESP's perspective. Use the full pin schema and set ``inverted: true`` to invert logic levels.
 - **rx_buffer_size** (*Optional*, int): The size of the buffer used for receiving UART messages. Increase if you use integration that needs to read big payloads from UART. Defaults to ``256``.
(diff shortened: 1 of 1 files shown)


Lewn:
Updated to use the full pin scheme to invert signal levels, see esphome/esphome#1727 for more details

components/uart.rst
```diff
@@ -44,18 +44,14 @@ Configuration variables:
...
+- **rx_pin** (*Optional*, :ref:`config-pin`): The pin to receive data on from the ESP's perspective. Use the full pin schema and set ``inverted: True`` to invert logic levels.
 - **rx_buffer_size** (*Optional*, int): The size of the buffer used for receiving UART messages. Increase if you use integration that needs to read big payloads from UART. Defaults to ``256``.
 - **data_bits** (*Optional*, int): The number of data bits used on the UART bus. Options: 5 to 8. Defaults to 8.
 - **parity** (*Optional*): The parity used on the UART bus. Options: ``NONE``, ``EVEN``, ``ODD``. Defaults to ``NONE``.
 - **stop_bits** (*Optional*, int): The number of stop bits to send. Options: 1, 2. Defaults to 1.
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID for this UART hub if you need multiple UART hubs.
 
-ESP32 options:
-
-- **invert** (*Optional*, boolean): Invert the logic levels of the RX and TX pins. Options: ``True`` or ``False``. Defaults to ``False``.
```
frenck: This configuration option is now completely missing?

Lewn: Correct, as the inversion is now done in the pin configuration. Do you think it be better to keep that configuration option as well? I thought it would be best to only have a single way to do the configuration.

frenck: I understand 👍 

My comment came out of the parent PR (which is more concern on that end).

frenck (approved):
✅ Approved. Can be merged as soon as the parent PR gets merged.

oxan:
GitHub doesn't let me reopen and merge this, so made & merged #1692.

PR #1121, opened 2021-04-29
GPIO Switch: simplify Momentary Switch example

ianchi:
Simplify case and avoid the need to use a secondary template switch.
Use `on_turn_on` directly to trigger the delay and turning off afterwards


glmnet (approved):

frenck (approved):
Thanks, @ianchi 👍 

../Frenck 🚂

PR #1127, opened 2021-05-02
Add sm2135 component

BoukeHaarsma23:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/443

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1736

## Checklist:

### DIFF
--- a/components/output/sm2135.rst
+++ b/components/output/sm2135.rst
@@ -114,9 +114,8 @@ Configuration variables:
 
 .. note::
 
-    The white LEDs are much brighter than the color LEDs and will fully overpower
-    the set color when the white level is set even a little bit high. You will need
-    to set the white level to 0 in order to get usable colors from this bulb.
+    The white LEDs are much brighter than the color LEDs. To get uniform brightness 
+    for both color and white you will need to limit the white led power.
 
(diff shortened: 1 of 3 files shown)


components/output/sm2135.rst
```diff
@@ -0,0 +1,138 @@
...
+
+    The white LEDs are much brighter than the color LEDs and will fully overpower
+    the set color when the white level is set even a little bit high. You will need
+    to set the white level to 0 in order to get usable colors from this bulb.
+
+.. note::
+
+    This driver does not support enabling of both the color and the white channels
+    at the same time. Therefore, the :ref:`rgbw_color_interlock` should be set to true
+    when using this driver.
```
jesserockz: Don't these 2 notes contradict each other?

The first says you can use white + rgb, but you need to turn down the white

And the second says it is not even supported...

BoukeHaarsma23: Ah my bad. Copied a bit too much here. I wanted to say that the white output typically is a lot brighter compared to the color output. I have updated this.

PR #1133, opened 2021-05-04
Explain where to upload .ttf font files

TheGroundZero:
## Description:

The documentation didn't mention .ttf font files should be uploaded in the config/esphome folder.

## Checklist:

### DIFF
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -134,12 +134,27 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the Internet and create a ``font:`` section in your configuration:
+a ``.ttf`` file from somewhere on the internet and upload it to the ``config/esphome`` folder.
+Next, create a ``font:`` section in your configuration:
 
 .. code-block:: yaml
 
     font:
       - file: "Comic Sans MS.ttf"
+        id: my_font
+        size: 20
+
+    display:
+      # ...
+
+
+If you created a ``fonts`` subfolder, reference the font file as follows:
+
+
+.. code-block:: yaml
+
+    font:
+      - file: "fonts/Comic Sans MS.ttf"
         id: my_font
         size: 20
 
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -134,7 +134,8 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the internet and upload it to the ESPHome root folder.
+a ``.ttf`` file from somewhere on the internet and place it, for example,
+inside a `fonts` folder next to your configuration file.
 
 Next, create a ``font:`` section in your configuration:
 
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -134,7 +134,7 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the internet and upload it to the ``config/esphome`` folder.
+a ``.ttf`` file from somewhere on the internet and upload it to the ESPHome root folder.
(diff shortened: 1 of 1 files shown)


components/display/index.rst
```diff
@@ -134,7 +134,8 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the Internet and create a ``font:`` section in your configuration:
+a ``.ttf`` file from somewhere on the internet and upload it to the ``config/esphome`` folder.
```
frenck: The `config/esphome` folder is very specific for Home Assistant installation of a certain type (e.g., Home Assistant OS and Supervised), and thus does not apply to all other users.

Can we change the wording to avoid that?

components/display/index.rst
```diff
@@ -147,6 +148,20 @@ a ``.ttf`` file from somewhere on the Internet and create a ``font:`` section in
       # ...
 
 
+If you created a ``fonts`` subfolder, reference the font file as follows:
```
frenck: Maybe we should just adjust the example above to have a subfolder by default.

frenck (changes requested):
Hi @TheGroundZero!

I've left 2 comments, could you take a look?

Thanks already 👍 

../Frenck

components/display/index.rst
TheGroundZero: For the sake of keeping everything organised, I'd definitely use subfolders for fonts etc.
Maybe `_fonts` is even better than `fonts` ?

frenck: > Maybe _fonts is even better than fonts ?

Why?

components/display/index.rst
```diff
@@ -147,14 +147,14 @@ Next, create a ``font:`` section in your configuration:
     display:
       # ...
 
-
-If you created a ``fonts`` subfolder, reference the font file as follows:
+You can organise your fonts by using subfolders.
+I.e., if you created a ``_fonts`` subfolder, reference the font file as follows:
```
frenck: Why not just update the example on line 143?

components/display/index.rst
TheGroundZero: It could work as a visual aid. Also keeps the folders up top in the explorer window.

components/display/index.rst
TheGroundZero: That would assume everyone's using a fonts subfolder?
This folder isn't present by default, so new people might be confused when they don't find the folder.

components/display/index.rst
frenck: Let's not try to outsmart or think for other people. The effect of doing such stuff, is that it raises the "Why?" question with our readers, just as I just had to ask.

components/display/index.rst
frenck: > That would assume everyone's using a fonts subfolder?

No? It is an example, showing it uses a path. Please adjust the above example instead. (there is no reason to have multiple examples to show "how paths work". The goal is to make clear it is a path that is accepted.

components/display/index.rst
```diff
@@ -134,19 +134,20 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the Internet and create a ``font:`` section in your configuration:
+a ``.ttf`` file from somewhere on the internet and upload it to the ESPHome root folder.
```
frenck:
```suggestion
a ``.ttf`` file from somewhere on the internet and place it, for example,
inside a `fonts` folder next to your configuration file.
```

TheGroundZero: `place` is double here?
> place, for example, place

==> ... and place it e.g., inside a `fonts` folder ...

frenck: right 🤕 
Fixed 🩹

components/display/index.rst
```diff
@@ -134,19 +134,21 @@ specific sizes, with ESPHome you have the option to use **any** TrueType (``.ttf
 at **any** size! Granted the reason for it is actually not having to worry about the licensing of font files :)
 
 To use fonts you first have to define a font object in your ESPHome configuration file. Just grab
-a ``.ttf`` file from somewhere on the Internet and create a ``font:`` section in your configuration:
+a ``.ttf`` file from somewhere on the internet and place it, for example,
+inside a `fonts` folder next to your configuration file.
```
frenck: Meh this isn't Markdown 😓 

```suggestion
inside a ``fonts`` folder next to your configuration file.
```

frenck (approved):
Better! Thanks, @TheGroundZero 👍 

../Frenck 🚂

PR #1143, opened 2021-05-08
Utalize pip3 for commands

rradar:
## Description:

Replaced `pip` commands by `pip3`

Newer *buntu based systems don't have `pip` command (only when installing python 2.x) but instead pip3 (python 3.x) is used.

## Checklist:


frenck (approved):
Thanks, @rradar 👍 

../Frenck

PR #1156, opened 2021-05-11
Sample codeblock has incorrect variables

jcallaghan:
## Description:

Four of the variables in the sample codeblock don't include underscores as defined in the variables section. This throws an error when using this sample code. To avoid others getting these errors I have updated the sample so it uses the right variables.

## Checklist:


frenck (approved):
Thanks, @jcallaghan 👍 

../Frenck

PR #1165, opened 2021-05-12
Some clarification added to external components.

mmakaay:
## Description:

Added a few clarifications to the external_components documentation. 

* Added a local example to the basic example config at the start of the documentation.
* Used list syntax for the configuration example for local components. Using a single dict does work, but it confused me a lot. It felt like `source.path:` was a config option for `external_components:` to tell it where to cache a downloaded repo or so. By not diverting the syntax in comparison with the git example, I think it takes away a level of cognitive burden.
* "By default `all`" kind of suggestst that `all` is a valid value for the `components:` option, but it isn't. Changed the wording to "By default, all available components are used" to prevent this.
* Because there is a difference in file layout between local and git-based components (git-based uses the extra `components` or `esphome/components` subfolders, which aren't used for `local`), I added a layout example for local components to clarify this.
* Some other small changes.

## Checklist:

### DIFF
--- a/components/external_components.rst
+++ b/components/external_components.rst
@@ -5,8 +5,8 @@ External Components
     :description: Instructions for setting up ESPHome External Components.
     :keywords: External, Custom, Components, ESPHome
 
-You can easily import community components using the external components feature. Bundled components
-can be overridden using this feature.
+You can easily import community or personal components using the external components feature.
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @mmakaay 👍 

../Frenck 🚂

PR #1179, opened 2021-05-17
Fix docker build failure

### LINKED ISSUE #2041: esphome-docs: docker image out of date (Python outdated)
**Operating environment/Installation (Hass.io/Docker/pip/etc.):**

Win10+Docker (esphome-docs)

**ESP (ESP32/ESP8266, Board/Sonoff):**

n/a

**ESPHome version (latest production, beta, dev branch)**

esphome-docs:current

**Affected component:**

Python version in docker image for esphome-docs (and possible other libraries/dependencies)
https://hub.docker.com/r/esphome/esphome-docs/
https://esphome.io/guides/contributing.html#contributing-to-esphome-docs

**Description of problem:**
Trying to build esphome-docs locally. Build inside Docker image fails. Python throwing syntax error due to F-strings. This feature was added with Python 3.6. The container seems to have Python 3.5

**Problem-relevant YAML-configuration entries:**
```yaml
n/a
```

**Logs (if applicable):**

```
PS C:\WINDOWS\system32> docker run --rm -v C:\esphomedocsnew\:/data/esphomedocs -p 18845:8000 -it esphome/esphome-docs
sphinx-build -M html . _build -j auto -n
Running Sphinx v1.8.4
loading translations [en]... done

Exception occurred:
  File "/usr/local/lib/python3.5/dist-packages/sphinx/registry.py", line 472, in load_extension
    mod = __import__(extname, None, None, ['setup'])
  File "/data/esphomedocs/schema_doc.py", line 26
    logger.info(f"{SCHEMA_PATH} not found. Not documenting schema.")
                                                                  ^
SyntaxError: invalid syntax
```

**Additional information and things you've tried:**

I managed to build the documentation with sphinx directly. I also had to generate the "schema.json" (with script/build_jsonschema,py) first.

martgras:
Fix the docker build error from https://github.com/esphome/issues/issues/2041

## Description:

````docker run --rm -v c:\esphomedev\esphome-docs\:/data/esphomedocs -p 8000:8000 -it esphome/esphome-docs```` fails with 
````
Running Sphinx v1.8.4
loading translations [en]... done

Exception occurred:
  File "/usr/local/lib/python3.5/dist-packages/sphinx/registry.py", line 472, in load_extension
    mod = __import__(extname, None, None, ['setup'])
  File "/data/esphomedocs/schema_doc.py", line 26
    logger.info(f"{SCHEMA_PATH} not found. Not documenting schema.")
                                                                  ^
SyntaxError: invalid syntax
````

The reason for this error is python 3.5 used in xenial. 
Switching to ubuntu focal resolves the error 

**Related issue (if applicable):** fixes https://github.com/esphome/issues/issues/2041

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM ubuntu:xenial
+FROM ubuntu:focal
+ENV TZ=UTC
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
@@ -17,6 +19,7 @@ RUN pip3 install --no-cache-dir --no-binary :all: \
         sphinx
 
 EXPOSE 8000
-WORKDIR /data/esphome-docs
+WORKDIR /data/esphomedocs
 
 CMD ["make", "webserver"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,13 @@
-FROM ubuntu:focal
-ENV TZ=UTC
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+FROM python:slim
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
-        python3 \
-        python3-pip \
-        python3-setuptools \
-        python3-pygments \
         git \
         make \
         doxygen \
         openssh-client \
         software-properties-common \
-    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+        && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
 
 RUN pip3 install --no-cache-dir --no-binary :all: \
         sphinx
@@ -22,4 +16,3 @@ EXPOSE 8000
 WORKDIR /data/esphomedocs
 
 CMD ["make", "webserver"]
-

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,4 +1,6 @@
-FROM ubuntu:xenial
+FROM ubuntu:focal
```
frenck: Instead, should use the official Docker container of Python, this makes it faster as well (especially if we used the Alpine versions instead).

frenck (changes requested):

Dockerfile
martgras: used python-slim now and removed python from apt commands
I didn't use the alpine image  because I found many recommendations against it . E.g. https://pythonspeed.com/articles/alpine-docker-python/

frenck: That is because of the wheels, however, we have the wheels available for Alpine :)
(as Home Assistant builds them). But nice for a future update.

martgras: to be honest I'm glad if I get python working at all and also not a docker pro  :)  So I'm pretty agnostic which images to use 

Should I rather use this for the PR ? 
A quick test didn't show a huge difference to the python-slim build but it is indeed pretty small 146MB
````
FROM python:alpine

ARG BUILD_ARCH
ENV WHEELS_LINKS=https://wheels.home-assistant.io/alpine-3.13/${BUILD_ARCH}/

RUN \
    apk add --no-cache \
    curl \
    git \
    make \
    doxygen \
    openssh-client 

RUN pip3 install --no-cache-dir --no-binary :all: \
    sphinx

EXPOSE 8000
WORKDIR /data/esphomedocs

CMD ["make", "webserver"]
````

frenck: Nah let's stick with what we have for now, maybe in the future; We could add dev containers based on the current implementation as well (which isn't possible with Alpine).

frenck (approved):
Thanks, @martgras 👍

PR #1191, opened 2021-05-21
Update sonoff_basic.rst

workingmanrob:
Use same recommended device config section as the info here - https://esphome.io/devices/sonoff.html

## Description: 

Slight difference/conflict as to what to set as the sonoff basic board type.  Just converted all my sonoff basics to use the same esphome block as the recommendation for the generic sonoff and everything is working well.

## Checklist:

### DIFF
--- a/devices/sonoff_basic.rst
+++ b/devices/sonoff_basic.rst
@@ -54,7 +54,8 @@ exposes all of the basic functions.
     esphome:
       name: <NAME_OF_NODE>
       platform: ESP8266
-      board: esp01_1m
+      board: esp8285
+      arduino_version: 2.4.2
 
     wifi:
       ssid: <YOUR_SSID>


devices/sonoff_basic.rst
```diff
@@ -54,7 +54,8 @@ exposes all of the basic functions.
     esphome:
       name: <NAME_OF_NODE>
       platform: ESP8266
-      board: esp01_1m
+      board: esp8285
+      arduino_version: 2.4.2
```
frenck, line 58: I wonder if this Arduino downgrade is still needed. This is based on information from 2019...

workingmanrob: I have a serial console connected to the one on my desk.

Will test without that arduino version line this afternoon and let you know how it goes..

workingmanrob: Did not go well.

Never reconnected to wifi and lucky for me I had the serial console so I could easily flash back the 2.4.2 version.

These devices are very finicky but using that declared arduino_version does seem to make a difference for the better.

frenck: Thanks! 👍

frenck (approved):
Thanks, @workingmanrob 👍

PR #1197, opened 2021-05-24
Extend Teleinfo functionality for Linky Standard mode

Jpsy:
## Description:

Component Teleinfo:
Standard mode provides a set of tags that does not only contain a value but also a timestamp, which is relevant data. This PR allows to read out the timestamp AND/OR the value of a tag (not only the value). The timestamp is read by associating a sensor with a tag and setting the data_class of the sensor to "timestamp".

PR also cleans up the documentation and adds a bunch of little details that were missing, including a circuit diagram.

(none)

https://github.com/esphome/esphome/pull/1829

## Checklist:


frenck (commented):
Hi there @Jpsy 👋 

There are unrelated commits in this PR. It seems like this commit was based on a different branch. Please perform a git rebase onto the upstream `next` branch to clean it up.

PR #1198, opened 2021-05-24
Update due to new version 1.18.0

azrael783:
## Description:
It seems that with version 1.18.0 it is not longer `measurement_time` but `measurement_duration` which should be reflected in the documentation also.

## Checklist:


frenck (approved):
Thanks, @azrael783 👍 

../Frenck 🚂

PR #1202, opened 2021-05-25
Remove `NFC` from RDM6300 as it is RFID not NFC

jesserockz:
## Description:

The RDM6300 is RFID not NFC, these two terms should not be confused.

## Checklist:


jesserockz:
Just realised the RC522 and PN532 docs are all over the place too. Will update this later

PR #1209, opened 2021-05-31
Update Sonoff 4CH docs

Chupaka:
## Description:

Swap RX/TX in pin list and add a warning about connecting some pins to GND on boot
Tested with `SONOFF 4CH R3 V1.1 2020-04-13`

## Checklist:


Chupaka:
Huh, warning looks a bit scary, probably we may use a notice instead. On the other hand, that stopped my device from booting when I was days and kilometers away from it, so it wouldn't harm.

frenck:
I agree on the warning a bit overkill. These should be used sparingly, please make it less intrusive.

Chupaka:
@frenck done!

frenck (approved):
Thanks, @Chupaka 👍

PR #1229, opened 2021-06-09
Update remote_transmitter.rst

pilleatus:
changed remote receiver code example to inverted mode, because that is usually needed..

## Description:

## Checklist:


frenck:
Doesn't this apply to the current documentation/stable version? If so, this PR should be targeting the `current` branch.

PR #1246, opened 2021-06-16
Override document title for index

ludeeus:
## Description:

Override the document title (used in tabs) on page load for the index

## Checklist:

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -36,6 +36,9 @@
     </div>
   </div>
   <script>
+    if (window.location.pathname === "/") {
+      window.document.title = "ESPHome"
+    }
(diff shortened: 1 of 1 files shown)


glmnet:
Ahh, I was about to mention this... this is why I said I failed when I tried removing the the ESPHome at top. Btw this is affecting SE previews too: 
image

Not sure if this fix will apply for that case.

ludeeus:
It will not, will look into that

ludeeus:
It actually looks like this solved the results as well, I did not expect that

glmnet:
I also see right title on google now. 

Makes sense the crawler run scripts to get the actual page title.

PR #1257, opened 2021-06-17
DIY examples: add ESPHome Heart Rate Display configuration

koenvervloesem:
Thanks to the raw BLE characteristics lambda in ESPHome 1.19, this project of me now works out-of-the-box. I haven't seen anything like this yet, so it looks like a nice addition to the DIY examples.


frenck (approved):
Thanks, @koenvervloesem 👍

PR #1262, opened 2021-06-18
Change set_supports_auto_mode for custom climate

DomiStyle:
## Description:

Change example to set_supports_heat_cool_mode according to https://github.com/esphome/esphome/pull/1933

## Checklist:


jesserockz (approved):
Awesome, thanks

PR #1276, opened 2021-06-26
Update fingerprint_grow.rst

haade-administrator:
error  tostring, to_string in description


frenck (approved):
Thanks, @haade-administrator 👍

PR #1278, opened 2021-06-26
Display: Add example how to use color

gladhorn:
The docs mentioned only binary displays, but color works just fine.

## Description:
Add snippet how to draw in color. The section was mentioning that only black and white is possible which is outdated.

## Checklist:


gladhorn:
I'm a bit unsure what the preferred way is to instantiate Color.

oxan:
> I'm a bit unsure what the preferred way is to instantiate Color.

This is fine.

oxan (approved):

PR #1296, opened 2021-07-09
Update Toshiba doc for new Toshiba AC IR protocol

kbx81:
## Description:

Updates Toshiba IR Climate documentation based on https://github.com/esphome/esphome/pull/1987

## Checklist:

  - [ ] Link added in `/index.rst` when creating new documents for new components or cookbook. **N/A**


components/climate/ir_climate.rst
```diff
@@ -154,11 +139,63 @@ Configuration variables:
         header_high: 3265us # AC Units from LG in Brazil, for example use these timings
         header_low: 9856us
 
+
+.. _toshiba:
+
+``toshiba`` Climate
+---------------------
```
glmnet: Too many dashes?

components/climate/ir_climate.rst
```diff
@@ -62,8 +62,8 @@ controller unit.
       - platform: coolix       # adjust to match your AC unit!
         name: "Living Room AC"
 
-Configuration variables:
-------------------------
+Configuration Variables
```
glmnet: By undocumented convention this title ends in : (see all other pages) if you want you can make it not be a title. (But keep the :)

It's actually expected the : for automatic parsing of it.

glmnet (changes requested):
Just some details.

components/climate/ir_climate.rst
kbx81: Too many dashes. I thought the linter catches these...

components/climate/ir_climate.rst
kbx81: Noted! 😄

glmnet (approved):
Thanks

PR #1309, opened 2021-07-16
Corrected value for "inverted"

azrael783:
## Description:
According to the pin schema the value for "inverted" should be true or false, but not yes.

## Checklist:

### DIFF
--- a/components/switch/gpio.rst
+++ b/components/switch/gpio.rst
@@ -59,7 +59,7 @@ To create an active-low switch (one that is turned off by default), use the :ref
       - platform: gpio
         pin:
           number: 25
-          inverted: yes
+          inverted: True
 
 Momentary Switch
 ----------------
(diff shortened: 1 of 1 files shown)


oxan:
Actually, you can use both yes/no, true/false, on/off and enable/disable ;) But yes, it's probably good to consistently use true/false in the examples.

components/switch/gpio.rst
```diff
@@ -59,7 +59,7 @@ To create an active-low switch (one that is turned off by default), use the :ref
       - platform: gpio
         pin:
           number: 25
-          inverted: yes
+          inverted: True
```
frenck: This change actually makes no difference, as in YAML both are equal in the YAML 1.1 specifications used.

Correct styling would be `true`/`false` in lower caps.


```suggestion
          inverted: true
```

frenck (approved):
Thanks, @azrael783 👍

PR #1315, opened 2021-07-19
docs for pm1006 component

Habbie:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2038

Image sources (please let me know if this is not okay):
* https://twitter.com/wuhan_cubic/status/984695460248961024
* https://www.ikea.com/nl/nl/p/vindriktning-luchtkwaliteitsensor-70498242/

## Checklist:


Habbie:
Oops, rebased onto `next` now.

components/sensor/pm1006.rst
```diff
@@ -0,0 +1,58 @@
...
+
+The implementation was inspired by https://github.com/Hypfer/esp8266-vindriktning-particle-sensor; you can also see the pinout there.
+
+Configuration variables:
+------------------------
+
+- **pm_2_5** (*Optional*): Use the concentration of particulates of size less than 2.5µm in µg per cubic meter.
+  All options from :ref:`Sensor <config-sensor>`.
+
+- **rx_only** (*Optional*, boolean): Whether the sensor is connected **only by RX pin** to the UART bus.
```
lassebm: This variable is unused and can be removed.

Habbie: oops, yes, I removed that somewhere through development

components/sensor/pm1006.rst
```diff
@@ -0,0 +1,54 @@
...
+
+    # Example configuration entry
+    uart:
+      rx_pin: D2
+      baud_rate: 9600
+
+    sensor:
+      - platform: pm1006
+        pm_2_5:
+          name: "Particulate Matter <2.5µm Concentration"
```
nohn: AFAIK it's >= 2.5µm

nohn (changes requested):
I would mention that the ESP should be mounted under the fan not in the large space above the sensor. The airflow is in from the upper side of the Vindrinktsdfsdf through the sensor, through the fan and out through the the lower side of the Vindrinktsdfsdf. The fan is a radial fan. So if the ESP is put on the upper side, it's in the airflow. If put below the fan, there's on obstruction to airflow at all caused by the ESP.

nohn (changes requested):
IKEA does not grant use of their images. I'd suggest not using the image file from their website but shoot one by yourself.

Habbie:
> The fan is a radial fan.

This is a good point.

> I would mention that the ESP should be mounted under the fan not in the large space above the sensor.

I did not even realise a D1 Mini might fit there, but it does.

components/sensor/pm1006.rst
Habbie: Hmm, almost all other sensors are documented with `<`. The datasheet (http://www.jdscompany.co.kr/download.asp?gubun=07&filename=PM1006_LED_PARTICLE_SENSOR_MODULE_SPECIFICATIONS.pdf) doesn't usefully answer the question. I'm tempted to just drop `<` like in the sm300d2 docs.

nohn: I checked, and you are correct.

Habbie: Ok, I'll drop the `<`, to match the language in the datasheet.

Habbie: done

Habbie:
Picture updated (thanks @nohn), text also updated based on @nohn's comments.

https://community.home-assistant.io/t/ikea-vindriktning-air-quality-sensor/324599 is full of great pictures of how people have done the mod; can I link to it from the docs?

OttoWinter:
> can I link to it from the docs?

Sure, do you want to include it in this PR?

Habbie:
> Sure, do you want to include it in this PR?

Yes! Will do it soon.

Habbie:
> Sure, do you want to include it in this PR?

Pushed!

PR #1326, opened 2021-07-23
Added more details to where to put the secrets

mrred2k:
Secrets.yaml in the Homeassistant folder didnt work within esphome.
Added the information here to be more precise and get less people confused.

## Description:

## Checklist:

### DIFF
--- a/guides/faq.rst
+++ b/guides/faq.rst
@@ -11,7 +11,7 @@ Tips for using ESPHome
 1. ESPHome supports (most of) `Home Assistant's YAML configuration directives
    <https://www.home-assistant.io/docs/configuration/splitting_configuration/>`__ like
    ``!include`` and ``!secret``. So you can store all your secret WiFi passwords and so on
-   in a file called ``secrets.yaml`` within the directory where the configuration file is.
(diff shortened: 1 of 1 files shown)


guides/faq.rst
```diff
@@ -11,7 +11,7 @@ Tips for using ESPHome
 1. ESPHome supports (most of) `Home Assistant's YAML configuration directives
    <https://www.home-assistant.io/docs/configuration/splitting_configuration/>`__ like
    ``!include`` and ``!secret``. So you can store all your secret WiFi passwords and so on
-   in a file called ``secrets.yaml`` within the directory where the configuration file is.
+   in a file called ``secrets.yaml`` within the directory where the esphome configuration files are.
```
frenck, line 14: Actually, the previous statement was correct.

> ``secrets.yaml`` within the directory where the configuration file is.

As they need to be next to each other, the configuration file that uses the secrets and the secrets file.

> within the directory where the esphome configuration files are.

This assumes there is only one directory, which doesn't have to be the case.

frenck:
Thanks for willing to contribute @mrred2k 👍 

But going to friendly decline this change for the above reasning.

../Frenck

PR #1332, opened 2021-07-25
Replace State with Position for Covers

EnochPrime:
## Description:
The use of state for covers was deprecated and has now been remove in v1.20.0. This updates the documentation to align with that change.

## Checklist:

### DIFF
--- a/components/cover/index.rst
+++ b/components/cover/index.rst
@@ -128,7 +128,6 @@ Configuration variables:
 
 - **id** (**Required**, :ref:`config-id`): The cover to control.
 - **stop** (*Optional*, boolean): Whether to stop the cover.
-- **state** (*Optional*, string): The state to set the cover to - one of ``OPEN`` or ``CLOSE``.
 - **position** (*Optional*, float): The cover position to set.
 
   - ``0.0`` = ``0%`` = ``CLOSED``
@@ -164,11 +163,11 @@ advanced stuff.
       id(my_cover).publish_state(COVER_OPEN);
       id(my_cover).publish_state(COVER_CLOSED);
 
-- ``state``: Retrieve the current state of the cover.
+- ``position``: Retrieve the current position of the cover.
 
   .. code-block:: yaml
 
-      if (id(my_cover).state == COVER_OPEN) {
+      if (id(my_cover).position == COVER_OPEN) {
         // Cover is open
       } else {
         // Cover is closed
--- a/components/cover/index.rst
+++ b/components/cover/index.rst
@@ -128,6 +128,7 @@ Configuration variables:
 
 - **id** (**Required**, :ref:`config-id`): The cover to control.
 - **stop** (*Optional*, boolean): Whether to stop the cover.
+- **state** (*Optional*, string): The state to set the cover to - one of ``OPEN`` or ``CLOSE``.
 - **position** (*Optional*, float): The cover position to set.
 
   - ``0.0`` = ``0%`` = ``CLOSED``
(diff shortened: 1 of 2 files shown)


components/cover/template.rst
```diff
@@ -92,20 +92,18 @@ with the ``cover.template.publish`` action.
...
       - cover.template.publish:
           id: template_cov
-          state: OPEN
+          position: OPEN
 
       # Templated
       - cover.template.publish:
           id: template_cov
-          state: !lambda 'return COVER_OPEN;'
+          position: !lambda 'return COVER_OPEN;'
```
jesserockz: This still applies to the action. This was not removed/changed.

components/cover/template.rst
```diff
@@ -92,20 +92,18 @@ with the ``cover.template.publish`` action.
...
 
 Configuration options:
 
 - **id** (**Required**, :ref:`config-id`): The ID of the template cover.
-- **state** (*Optional*, :ref:`templatable <config-templatable>`):
-  The state to publish. One of ``OPEN``, ``CLOSED``. If using a lambda, use ``COVER_OPEN`` or ``COVER_CLOSED``.
 - **position** (*Optional*, :ref:`templatable <config-templatable>`, float):
-  The position to publish, from 0 (CLOSED) to 1.0 (OPEN)
+  The position to publish, from 0 (CLOSED) to 1.0 (OPEN). If using a lambda, use ``COVER_OPEN`` or ``COVER_CLOSED``.
 - **tilt** (*Optional*, :ref:`templatable <config-templatable>`, float):
```
jesserockz: Same here.

components/cover/index.rst
```diff
@@ -128,7 +128,6 @@ Configuration variables:
 
 - **id** (**Required**, :ref:`config-id`): The cover to control.
 - **stop** (*Optional*, boolean): Whether to stop the cover.
-- **state** (*Optional*, string): The state to set the cover to - one of ``OPEN`` or ``CLOSE``.
```
jesserockz: This still applies and was not removed.

jesserockz (changes requested):

components/cover/template.rst
EnochPrime: You are correct, this compiles fine. I will revert this change.

components/cover/template.rst
EnochPrime: This part perhaps needs a little clarification. The note afterward describes an alternative way that one could write a lambda. I believe it would be a reasonable assumption that `id(template_cov).state = COVER_OPEN;` would be valid, however, this throws a compile-time error.

> class esphome::template_::TemplateCover' has no member named 'state'

components/cover/index.rst
EnochPrime: Same response as to the template cover.

components/cover/template.rst
jesserockz: Lambda/c++ variables do not always match the yaml configuration variables.
In YAML, `position` is a float, whereas `state` is a `string` `OPEN` or `CLOSED`

The removed field is in c++ `.state` and the `state` string in YAML is converted to 1.0 or 0.0 and then set to `position`

The `position` lambda for a template cover should not be doing `id(template_cov).state = COVER_OPEN;`, it should be `return`ing the position/state that the cover is in: `return COVER_OPEN;`

EnochPrime: If there are more instances where the C++ and YAML differ then if and how that should be documented has a broader impact that would be out of scope for this pull request.

As the documentation was correct in this instance (and the next) I will revert the change.

oxan:
Hi! Unfortunately in the meantime #1463 already fixed this issue, so I'll close this PR as it's no longer necessary. Nevertheless, thanks for your contribution!

PR #1351, opened 2021-07-30
Replace Color::BLACK by COLOR_BLACK

soundstorm:
Color::BLACK does not exist, haven't checked `random_color()`

## Description:

`Color::BLACK` does somehow not exist

## Checklist:


oxan:
See also esphome/issues#2068 and esphome/esphome#2093.

oxan:
`Color::BLACK` has been added back to the dev release and will be released with v1.21, so I don't think this update applies anymore.

PR #1352, opened 2021-07-31
Improve documentation with information about modbus_id variable

vicfergar:
Improve documentation with information about modbus_id variable

## Description:

**Related issue (if applicable):** [#1896](https://github.com/esphome/issues/issues/1896)

## Checklist:


jesserockz (approved):
Technically `modbus` is auto loaded so its not explicitly needed in yaml, but this is fine to add to the docs. Thanks

PR #1353, opened 2021-07-31
Adding references to other already existing tuya components

alexbarcelo:
Minor improvement. There are several Tuya components, it makes sense to reference all of them from the "parent" Tuya thingy.


jesserockz (approved):

jesserockz:
Thanks!

PR #1366, opened 2021-08-04
Add FAQ advising against underscores in hostnames.

davet2001:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2079

## Checklist:

### DIFF
--- a/guides/faq.rst
+++ b/guides/faq.rst
@@ -343,6 +343,14 @@ with a lot of work).
 
 Always back up all your files!
 
+
+Why shouldn't I use underscores in my device name?
+--------------------------------------------------
+
+The top level ``name:`` field in your .yaml file defines the node name(/hostname) on the local network.  According to `RFC1912 <https://datatracker.ietf.org/doc/html/rfc1912>`_, underscore characters (``_``) in hostnames are not valid.  In reality some local DNS/DHCP setups will be ok with underscores and some will not.  If connecting via a static IP address, there will probably be no issues.  In some cases, initial setup using an underscore works, but then the connection fails when Home Assistant restarts.  Recommendation: use hyphen (``-``) instead of underscore if you can.  
+
+Important: follow these `instructions <https://esphome.io/components/esphome.html#changing-esphome-node-name>`_ to use the ``use_address`` parameter when renaming a live device, as the connection to an existing device will only work with the old name until the name change is complete.
+
 See Also
 --------
 
--- a/guides/faq.rst
+++ b/guides/faq.rst
@@ -347,7 +347,7 @@ Always back up all your files!
 Why shouldn't I use underscores in my device name?
 --------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


guides/faq.rst
```diff
@@ -343,6 +343,14 @@ with a lot of work).
 
 Always back up all your files!
 
+
+Why shouldn't I use underscores in my device name?
+--------------------------------------------------
+
+The top level ``name:`` field in your .yaml file defines the node name(/hostname) on the local network.  According to `RFC1912 <https://datatracker.ietf.org/doc/html/rfc1912>`_, underscore characters (``_``) in hostnames are not valid.  In reality some local DNS/DHCP setups will be ok with underscores and some will not.  If connecting via a static IP address, there will probably be no issues.  In some cases, initial setup using an underscore works, but then the connection fails when Home Assistant restarts.  Recommendation: use hyphen (``-``) instead of underscore if you can.
```
glmnet: >  but then the connection fails when Home Assistant restarts.

could this be more like: "but then the connection **might** fail when Home Assistant restarts."

guides/faq.rst
```diff
@@ -343,6 +343,14 @@ with a lot of work).
 
 Always back up all your files!
 
+
+Why shouldn't I use underscores in my device name?
+--------------------------------------------------
+
+The top level ``name:`` field in your .yaml file defines the node name(/hostname) on the local network.  According to `RFC1912 <https://datatracker.ietf.org/doc/html/rfc1912>`_, underscore characters (``_``) in hostnames are not valid.  In reality some local DNS/DHCP setups will be ok with underscores and some will not.  If connecting via a static IP address, there will probably be no issues.  In some cases, initial setup using an underscore works, but then the connection fails when Home Assistant restarts.  Recommendation: use hyphen (``-``) instead of underscore if you can.  
+
+Important: follow these `instructions </components/esphome.html#changing-esphome-node-name>`_ to use the ``use_address`` parameter when renaming a live device, as the connection to an existing device will only work with the old name until the name change is complete.
```
glmnet, line 352:
```suggestion
Follow these `instructions </components/esphome.html#changing-esphome-node-name>`_ to change the underscores in the name to hyphens.
```

No need to explain why here, it's explained in the instructions.

glmnet (commented):
I see no network gear is talk about, I mean, an advise that underscores can work with your current network router but might stop working if you upgrade those, makes sense?

guides/faq.rst
davet2001: I think it is better to give some idea why here.  Many will assume they can just rename it.

guides/faq.rst
davet2001: Agreed, done.

davet2001:
> I see no network gear is talk about, I mean, an advise that underscores can work with your current network router but might stop working if you upgrade those, makes sense?

Agreed. Updated.

guides/faq.rst
```diff
@@ -343,6 +343,14 @@ with a lot of work).
 
 Always back up all your files!
 
+
```
OttoWinter, line 346:
```suggestion
.. _faq-device_name_underscores:
```

So we can link to it better

OttoWinter: Nvm, apparently the link is not generated with that.

PR #1369, opened 2021-08-05
Update cookbook for ESP32 Water Leak Detector

ryanalden:
The current cookbook yaml no longer compiles with ESPhome >1.16.0.
The following changes rectify the situation.

1) The ST7735 display is now natively supported in ESPHome >1.16.0.  Therefore the custom component is no longer needed. Also, ESPHome color handling seems to have changed, so that has been accounted for. (Without a way to invert the display during initialization, colors had to be adjusted manually to get the proper result). Code and instructions updated accordingly.

2) Updated URL of esp32 bootloader

3) Added st7735 display docs link to footer

## Description:

## Checklist:

### DIFF
--- a/cookbook/leak-detector-m5stickC.rst
+++ b/cookbook/leak-detector-m5stickC.rst
@@ -15,11 +15,11 @@ Using the ESP32's capacitive touch GPIOs, it's relatively easy to build a water 
 Things you'll need
 ==================
 
-- M5Stick esphome components
-    https://github.com/airy10/esphome-m5stickC/issues/4
+-  `M5Stick axp192 custom component <https://github.com/airy10/esphome-m5stickC>`__
+    This is needed to power up the display.  You don't need the st7735 display driver, as it is already included with ESPHome >1.16.0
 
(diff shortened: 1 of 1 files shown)


cookbook/leak-detector-m5stickC.rst
```diff
@@ -15,11 +15,11 @@ Using the ESP32's capacitive touch GPIOs, it's relatively easy to build a water
 Things you'll need
 ==================
 
-- M5Stick esphome components
-    https://github.com/airy10/esphome-m5stickC/issues/4
+-  `M5Stick axp192 custom component <https://github.com/airy10/esphome-m5stickC>`__
+    This is needed to power up the display.  You don't need the st7735 display driver, as it is already included with ESPHome >1.16.0
```
oxan:
```suggestion
    This is needed to power up the display.  You don't need the st7735 display driver, as it is already included with ESPHome >1.16.0.
```
Would be nice if someone upstreamed that component :)

oxan (changes requested):

PR #1378, opened 2021-08-09
Update automations.rst

arallsopp:
made it clear that lambda delay is in milliseconds rather than seconds. Ref: https://community.home-assistant.io/t/lambda-value-on-delay/235794

## Description:
The documentation frequently gives delay values in seconds or minutes, but a few people are being tripped up by lambda delays as they don't realise it actually wants milliseconds.

## Checklist:


jesserockz (approved):
Thanks!

arallsopp:
Thank you @jesserockz

PR #1380, opened 2021-08-10
Update svg2png and seo.py

OttoWinter:
## Description:

`svg2png` is needed because the social headers don't support svgs.

This PR adjusts the script and runs it 

## Checklist:


jesserockz (approved):
LGTM. So how does this work exactly, do the social image requests just fallback to try find a png with the same name?

OttoWinter:
See for example this file: https://github.com/esphome/esphome-docs/blob/current/components/stepper/index.rst

at the top it says `folder-open.png` (these images are used by twitter/discord for example for the image you see as a link preview)

Note it's a png, even though the actual file is an svg. The script basically just creates a png version of every svg file (and then get copied to the images folder on build)

jesserockz:
So as well as running this script, we need to not allow people to put .svg in the seo block?

OttoWinter:
I guess so; but we can also just change seo.py to replace svg with png, which should fix it as well

OttoWinter:
Changed it so that now you _have_ to put .svg there, and it checks if the file really exists (hope it won't cause too much trouble for new integrations)

PR #1381, opened 2021-08-10
Document flash_write_interval

alexyao2015:
## Description:
Documents flash_write_interval

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2119

## Checklist:

### DIFF
--- a/components/esphome.rst
+++ b/components/esphome.rst
@@ -170,7 +170,7 @@ A safety feature has thus been implemented to mitigate issues resulting from the
 the state is first stored in memory before being flushed to flash after the ``flash_write_interval`` has passed. This
 results in fewer flash writes, preserving the flash health.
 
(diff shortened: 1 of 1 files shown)


glmnet:
I would like to read first what this setting does, then **May be** keep reading on why it works like it does. This seems to be like a history lesson followed by solution provided to a problem. I think the information is ok anyway but I usually want to read as little as possible.

OttoWinter (commented):
I guess we can keep the documentation for the feature in this file (not sure where else to put it, new index entry for just one option would be a bit useless), but needs to be updated for that

jesserockz (commented):
I think at least move the config option to its own block on this page?

jesserockz (approved):

PR #1408, opened 2021-08-19
Add switch m5stack 4relay

brotherdust:
## Description:
Add docs for M5Stack 4Relay

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2112

## Checklist:


components/switch/m5stack_4relay.rst
```diff
@@ -0,0 +1,50 @@
...
+        - platform: m5stack_4relay
+          id: relay_3
+          channel: 2
+          name: Relay 3
+        - platform: m5stack_4relay
+          id: relay_4
+          channel: 3
+          name: Relay 4
+
+
```
oxan, line 42: There should be a configuration variables section here.

components/output/mcp4728.rst
```diff
@@ -0,0 +1,64 @@
+MCP4728 Output
```
oxan, line 1: This seems to be for another PR, so the docs should also be in another PR.

PR #1437, opened 2021-08-31
Note to set ``optimistic`` when using HA to set

mnaz:
For a while I was trying to use the Number Template and I couldn't figure why changes I made from the HA UI did not apply in my ESP device.
I think adding a note here will help or revising the text that already exists.
It also might be nice to add a note to have ``restore_count`` set also.

This is what I used.

    number:
      - platform: template
        name: Trigger Count
        id: trigger_count
        icon: "mdi:sync"
        restore_value: true
        optimistic: true
        initial_value: 8
        max_value: 10
        min_value: 1
        step: 1

## Description:

## Checklist:

### DIFF
--- a/components/number/template.rst
+++ b/components/number/template.rst
@@ -34,7 +34,8 @@ Configuration variables:
 - **update_interval** (*Optional*, :ref:`config-time`): The interval to check the
   number. Defaults to ``60s``.
 - **optimistic** (*Optional*, boolean): Whether to operate in optimistic mode - when in this mode,
-  any command sent to the template number will immediately update the reported state.
+  any command sent to the template number will immediately update the reported state. 
+  This must be set to ``true`` inorder for changes from Home Assistant to be applied.
   Cannot be used with ``lambda``. Defaults to ``false``.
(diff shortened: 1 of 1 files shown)


components/number/template.rst
```diff
@@ -34,7 +34,8 @@ Configuration variables:
 - **update_interval** (*Optional*, :ref:`config-time`): The interval to check the
   number. Defaults to ``60s``.
 - **optimistic** (*Optional*, boolean): Whether to operate in optimistic mode - when in this mode,
-  any command sent to the template number will immediately update the reported state.
+  any command sent to the template number will immediately update the reported state. 
+  This must be set to ``true`` inorder for changes from Home Assistant to be applied.
```
oxan, line 38: This is not true, you could also have a `set_action` that does something based on the changed value from HA and a `lambda` that reports back what the current value is.

oxan (changes requested):

oxan:
I think #1441 adequately took care of this problem (with the related PR it's no longer even possible to create numbers with `optimistic: false` without setting a `set_action`). Feel free to reopen if you disagree.

mnaz:
perfect, I just wanted to help avoid others the confusion.

PR #1439, opened 2021-09-01
Misc documentation updates

### LINKED ISSUE #1380: Add possibility to read fan speed in lambdas e.g. id(fan.my_fan).speed

I need to be able to react on fan speed changes set by either home assistant or locally by a micro switch and automation. Currently this is impossible as there is no way to read a fan speed (PWM level).

I can't find any solution to this currently

**Additional context**

nuttytree:
Misc documentation updates

## Description:
Remove duplicate on/off conditions in the core switch component
Replace covers with switches in the core switch component lambda calls section
Add a lambda calls section in the core fan component

**Related issue (if applicable):** fixes esphome/feature-requests#1380

## Checklist:

### DIFF
--- a/components/fan/index.rst
+++ b/components/fan/index.rst
@@ -123,6 +123,77 @@ if a command to turn the fan on or off already matches the current state.
         on_turn_off:
         - logger.log: "Fan Turned Off!"
 
+lambda calls
+************
+
+From :ref:`lambdas <config-lambda>`, you can call several methods on all fans to do some
+advanced stuff (see the full API Reference for more info).
+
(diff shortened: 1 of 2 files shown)


components/fan/index.rst
```diff
@@ -123,6 +123,77 @@ if a command to turn the fan on or off already matches the current state.
         on_turn_off:
         - logger.log: "Fan Turned Off!"
 
+lambda calls
```
oxan:
```suggestion
Lambda calls
```

oxan (approved):

PR #1444, opened 2021-09-05
Template sensor publish updates

nuttytree:
## Description:
This updates the documentation for the publish interval of template sensors to include that the value is now published on this interval for sensors with a lambda function and sensors that only get values from the publish action.

**Related issue (if applicable):** fixes N/A

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2224

## Checklist:

### DIFF
--- a/components/text_sensor/template.rst
+++ b/components/text_sensor/template.rst
@@ -31,6 +31,9 @@ Configuration variables:
 - **name** (**Required**, string): The name of the text sensor.
 - **lambda** (*Optional*, :ref:`lambda <config-lambda>`):
   Lambda to be evaluated every update interval to get the new value of the text sensor
+- **update_interval** (*Optional*, :ref:`config-time`): The interval to publish the value of the
+  text sensor, either the result of the lambda function or if no lambda function the last value
+  published using the publish action. Defaults to ``60s``.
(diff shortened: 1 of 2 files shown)


components/text_sensor/template.rst
```diff
@@ -31,6 +31,9 @@ Configuration variables:
 - **name** (**Required**, string): The name of the text sensor.
 - **lambda** (*Optional*, :ref:`lambda <config-lambda>`):
   Lambda to be evaluated every update interval to get the new value of the text sensor
+- **update_interval** (*Optional*, :ref:`config-time`): The interval to publish the value of the
+  text sensor, either the result of the lambda function or if no lambda function the last value
+  published using the publish action. Defaults to ``60s``.
 - **update_interval** (*Optional*, :ref:`config-time`): The interval to check the
   text sensor. Defaults to ``60s``.
```
jesserockz: Duplicate
```suggestion
- **update_interval** (*Optional*, :ref:`config-time`): The interval to publish the value of the
  text sensor, either the result of the lambda function or if no lambda function the last value
  published using the publish action. Defaults to ``60s``.
```

jesserockz (approved):

PR #1462, opened 2021-09-11
Shelly 2.5 config for ADE7953

rfpronk:
## Description:

Add Shelly 2.5 config for ADE7953

**Related issue (if applicable):** n.a.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** n.a.

## Checklist:

  - [x] ~Link added in `/index.rst` when creating new documents for new components or cookbook.~ n.a.


components/sensor/ade7953.rst
```diff
@@ -85,39 +77,54 @@ Additionally, the device has an ::doc:`NTC temperature sensor <ntc>`.
 
     sensor:
       - platform: ade7953
-        irq_pin: GPIO16
+        irq_pin: GPIO16 # Prevent overheating by setting this
+        update_interval: 10s
```
oxan: This comment is on the wrong line:
```suggestion
        irq_pin: GPIO16
        update_interval: 10s # Prevent overheating by setting this
```

components/sensor/ade7953.rst
```diff
@@ -85,39 +77,54 @@ Additionally, the device has an ::doc:`NTC temperature sensor <ntc>`.
...
         voltage:
-          name: Shelly Voltage
-        current_a:
-          name: Shelly Current B
+          name: "${devicename} voltage"
```
oxan: You're using substitutions here that aren't defined in this code block. For simplicity I'd make it just use Shelly, users who want substitutions can put them in themselves.

components/sensor/ade7953.rst
```diff
@@ -85,39 +77,54 @@ Additionally, the device has an ::doc:`NTC temperature sensor <ntc>`.
...
+          on_value_range:
+            - above: ${max_power}
+              then:
+                - switch.turn_off: shelly_relay_1
+                - homeassistant.service:
+                    service: persistent_notification.create
+                    data:
+                      title: Message from ${devicename}
+                    data_template:
+                      message: "Switch ${channel_1} turned off because power exceeded ${max_power}W"
```
oxan: This isn't the right place to demonstrate these automations.

components/sensor/ade7953.rst
```diff
@@ -85,39 +77,54 @@ Additionally, the device has an ::doc:`NTC temperature sensor <ntc>`.
...
+          on_value_range:
+            - above: ${max_power}
+              then:
+                - switch.turn_off: shelly_relay_2
+                - homeassistant.service:
+                    service: persistent_notification.create
+                    data:
+                      title: Message from ${devicename}
+                    data_template:
+                      message: "Switch ${channel_2} turned off because power exceeded ${max_power}W"
```
oxan: Same here.

oxan (changes requested):

components/sensor/ade7953.rst
rfpronk: Actually it's not.  Setting `irq_pin` prevents the chip from overheating as described by [the ade7953 documentation](https://esphome.io/components/sensor/ade7953.html) and very much the case for my devices.
I added`update_interval` to illustrate that 10s refresh rate is fine for the sensor.

oxan: Ah, didn't know that, then this is fine. I think `update_interval` is sufficiently easy to understand and universal that it can remain.

rfpronk:
@oxan Thanks for the review. I added a commit to process your comments.

oxan:
There's still some substitutions remaining, looks good otherwise.

rfpronk:
> There's still some substitutions remaining, looks good otherwise.

Sorry about that

oxan (approved):
No worries!

oxan:
Actually, is there any reason you moved the config block for the NTC sensor?

rfpronk:
> Actually, is there any reason you moved the config block for the NTC sensor?

I've removed that as it has nothing to do with the ADE7953 sensor (which this doc is about), but if you prefer we can keep it in there.

oxan:
Fair enough, that makes sense.

PR #1467, opened 2021-09-12
Update mhz19.rst

dj-bauer:
## Description:
I just tested the MH-Z16 with the MH-Z19 component and reading the co2 concentration worked flawlessly
Therefore we could add that in the documentation


components/sensor/mhz19.rst
```diff
@@ -8,6 +8,7 @@ MH-Z19 CO_2 and Temperature Sensor
 
 The ``mhz19`` sensor platform allows you to use MH-Z19 CO_2 and temperature sensors
 (`refspace`_) with ESPHome.
+The co2-measuring also works with the MH-Z16 sensor.
```
oxan:
```suggestion
The CO_2 measurement also works with the MH-Z16 sensor.
```

oxan (changes requested):

PR #1507, opened 2021-10-05
Edited / added accuracy_decimals

genestealer:
## Description:
Current "accuracy_decimals: 0" only returns values in 1kWh amounts, so usage below 1KWh will not be logged - (0kWh vs 0.736 kWh). Similar is true for the "kW" sensor, more accuracy is being measured that what is being returned if using the example "as-is". 

## Checklist:

  - [ x] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [ x] Link added in `/index.rst` when creating new documents for new components or cookbook.

### DIFF
--- a/components/sensor/pulse_meter.rst
+++ b/components/sensor/pulse_meter.rst
@@ -64,6 +64,7 @@ measure the total consumed energy in kWh.
         pin: 12
         unit_of_measurement: 'kW'
         name: 'Electricity Usage'
+        internal_filter: 100ms
         accuracy_decimals: 5
         filters:
           - multiply: 0.06
--- a/components/sensor/pulse_meter.rst
+++ b/components/sensor/pulse_meter.rst
@@ -71,7 +71,7 @@ measure the total consumed energy in kWh.
         total:
           name: "Electricity Total"
           unit_of_measurement: "kWh"
-          accuracy_decimals: 5
+          accuracy_decimals: 3
           filters:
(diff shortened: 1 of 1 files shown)


components/sensor/pulse_meter.rst
```diff
@@ -64,12 +64,14 @@ measure the total consumed energy in kWh.
         pin: 12
         unit_of_measurement: 'kW'
         name: 'Electricity Usage'
+        internal_filter: 100ms
+        accuracy_decimals: 5
```
jesserockz: 3 makes more sense for kW as 0.001 is 1 Watt, not much point being more accurate than that

```suggestion
        accuracy_decimals: 3
```

components/sensor/pulse_meter.rst
```diff
@@ -64,12 +64,14 @@ measure the total consumed energy in kWh.
...
         filters:
           - multiply: 0.06
         total:
           name: "Electricity Total"
           unit_of_measurement: "kWh"
-          accuracy_decimals: 0
+          accuracy_decimals: 5
```
jesserockz: Same as above
```suggestion
          accuracy_decimals: 3
```

jesserockz (changes requested):

jesserockz (approved):

PR #1516, opened 2021-10-08
Create a flash guide for Sonoff 4CH R3 (as layout and method differs from R2)

arallsopp:
## Description:
The existing methods described in devices/sonoff_4ch.rst are based upon the R2, and do not apply to the current R3 version of this board. The header pins required to flash this board must be added by the end user, are in a different location, and the button used to initiate the flash at boot has a different position and label. 

This PR copies the sonoff_4ch.rst file content as sonoff_4ch_r3 and makes amends to the content to support the revised methods for accessing and flashing this device. It also adds new images to support the revisions, and adds links to the R3 variant in the sonoff and sonoff_4ch pages.

## Checklist:


arallsopp:
Any action on my part required to get this one rolling?

jesserockz:
First off, please rebase this PR from `current` as it is not linked to a specific feature that will be released with code and can go directly to the live site.

jesserockz (changes requested):
Considering the amount of actual differences between the 4CH and the 4CH R3, I would prefer that it just me merged in with the original 4CH page.

arallsopp:
Fair enough. I can do that. Cheers.

PR #1521, opened 2021-10-11
Added documentation for a throttle_average filter

rmeissn:
## Description:

Adds documentation for a throttle_average filter
Background: There is no filter which throttles sensors with variable sensor update rates without discarding any values

**Related issue (if applicable):** https://github.com/esphome/feature-requests/issues/569

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** https://github.com/esphome/esphome/pull/2485

## Checklist:

### DIFF
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -378,7 +378,9 @@ An average over the ``specified time period``, potentially throttling incoming v
 * no value received: nothing will be pushed - add the ``heartbeat`` filter if periodical pushes are required
 * one value received: the value is pushed out after the ``specified time period`` passed, without calculating an average
 
-In comparison to the ``throttle`` filter it won't discard any values. In comparison to the ``sliding_window_moving_average`` filter it supports variable sensor reporting intervals, allows to receive slow sensor update rates as they are and to throttle fast sensor update rates to the ``specified time period``.
+For example a ``throttle_average: 60s`` will push out a value every 60 seconds, in case at least one sensor value is received within these 60 seconds.
+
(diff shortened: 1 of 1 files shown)


rmeissn:
Currently the implementation uses a different name for the filter. I don't mind changing it to the best fitting one. 
What do reviewers think about the filter name?

sermayoral:
@rmeissn thanks for the documentation!! 

The implementation doesn't match exactly with the documentation. The filter publish a value on each period. For a period of 120s, a value will be publish at 120s, the second value will be publish at 240s, the third one will be publish at 360s. That value will be the average of all values received in the last 120s. 

The documentation says the trigger is a received value, and the implementation says the trigger is a time. 

In my opinion is better in that way because it guarantees a value exactly at 120s. As documentation says, If you receive a value at 100s, and you receive the second one at 200s, the average will be published at 200s, very far of the defined period (120s)

components/sensor/index.rst
```diff
@@ -369,6 +370,17 @@ If it is not older than the configured value, the value is not passed forward.
       - delta: 5.0
       - lambda: return x * (9.0/5.0) + 32.0;
 
+``throttle_average``
+********************
+
+An average over the ``specified time period``, potentially throttling incoming values. When this filter gets incoming values, it sums up all values and pushes out the average after the ``specified time period`` passed. There are two edge cases to consider within the ``specified time period``:
+
+* no value received: ``NaN`` is returned - add the ``heartbeat`` filter if periodical pushes are required and/or ``filter_out: nan`` if required
```
jesserockz:
```suggestion
* no value(s) received: ``NaN`` is returned - add the ``heartbeat`` filter if periodical pushes are required and/or ``filter_out: nan`` if required
```

jesserockz (approved):

PR #1539, opened 2021-10-18
std::to_string() triggers an error on compile

odya:
Calling ``std::to_string()`` was causing "`to_string is not part of std library`" error. Fixed by using ``esphome::to_string``

## Description:

## Checklist:

### DIFF
--- a/components/logger.rst
+++ b/components/logger.rst
@@ -188,7 +188,7 @@ using ``message`` (``const char *``), ``level`` (``int``) and ``tag`` (``const c
           - mqtt.publish:
               topic: some/topic
               payload: !lambda |-
-                return "Triggered on_message with level " + std::to_string(level) + ", tag " + tag + " and message " + message;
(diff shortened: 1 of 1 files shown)


components/logger.rst
```diff
@@ -188,7 +188,7 @@ using ``message`` (``const char *``), ``level`` (``int``) and ``tag`` (``const c
           - mqtt.publish:
               topic: some/topic
               payload: !lambda |-
-                return "Triggered on_message with level " + std::to_string(level) + ", tag " + tag + " and message " + message;
+                return "Triggered on_message with level " + esphome::to_string(level) + ", tag " + tag + " and message " + message;
```
jesserockz: The `esphome` namespace is not needed.
```suggestion
                return "Triggered on_message with level " + to_string(level) + ", tag " + tag + " and message " + message;
```

jesserockz (approved):

jesserockz:
Thanks!

PR #1552, opened 2021-10-21
Add note to release notes about GPIO0

jesserockz:
## Description:

## Checklist:


OttoWinter (changes requested):
ah note that actually was a bug. i confused things a bit. esp8266 does have toggleable pullup resistors on gpio0, but gpio16 does not.

more info here https://github.com/esphome/esphome/pull/2572

PR #1574, opened 2021-10-25
docs: add generic ESP32-CAM-MB modules' config

RubenKelevra:
## Checklist:

### DIFF
--- a/components/esp32_camera.rst
+++ b/components/esp32_camera.rst
@@ -94,6 +94,46 @@ Frame Settings:
     Camera uses PWM timer #1. If you need PWM (via the ``ledc`` platform) you need to manually specify
     a channel there (with the ``channel: 2``  parameter)
 
+Configuration for generic ESP32-CAM-MB board
+-----------------------------------
+
+.. code-block:: yaml
+
+    # Example configuration entry
+    
+    
+    esp32_camera:
+      external_clock:
+        pin: GPIO0
+        frequency: 20MHz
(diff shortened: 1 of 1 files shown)


components/esp32_camera.rst
```diff
@@ -94,6 +94,46 @@ Frame Settings:
...
+        pin: GPIO33
+        id: gpio_33
+
+    light:
+      - platform: binary
+        output: gpio_4
+        name: "$friendly_name flashlight"
+      - platform: binary
+        output: gpio_33
+        name: "$friendly_name Red-LED"
```
oxan, line 135: This uses substitutions that are specific to your configs, so using them isn't suitable for the generic documentation. However, this page about the `esp32_camera` component isn't really the place to document what other pins on this board do anyway, so this whole section should just be left out.

oxan (changes requested):

components/esp32_camera.rst
RubenKelevra: I mean it's a generic ESP32-camera module and the page lists multiple different models of such devices with the specific configurations.

Where else would you put it?

oxan: I meant specifically the config for the `output` and `light` components, for none of the other models is any config beyond that of the `esp32_camera` component listed (and that part is fine here).

A better place for the full config is probably the cookbook or a submission to esphome-devices.com.

PR #1585, opened 2021-10-31
Add M5Stack Timer Camera X/F

tannisroot:
## Description:

Pinouts taken from https://docs.m5stack.com/en/unit/timercam_f (identical to the X model) and verified on real hardware.

## Checklist:

### DIFF
--- a/components/esp32_camera.rst
+++ b/components/esp32_camera.rst
@@ -149,6 +149,28 @@ Configuration for M5Stack Camera
       name: My Camera
       # ...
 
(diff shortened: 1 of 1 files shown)


martgras (approved):
LGTM

flaviut:
LGTM, looks like a great improvement to the docs.

Should probably be targeted against `current`, not `next`, since it does not have an associated code change.

flaviut (approved):

PR #1600, opened 2021-11-10
Webserver v2

wilberforce:
## Description:
Update to the webserver with a more advanced front end. Defaults to Version 1 and works as before

- Light weight web components used focusing on minimising flash size.
- Responsive layouts mobile/tablet/desktop
- Dark/Light scheme - defaults to OS preference

https://github.com/esphome/esphome/pull/2688

## Checklist:

  - [x ] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [ ]<strike> Link added in `/index.rst` when creating new documents for new components or cookbook. </strike>

## future features
 - more components - climate, select, text
 -  document theming, colour inheritance

At present the repo to generate is here.. https://github.com/wilberforce/esphome-webserver 
Need to find a home for perhaps a  sub-repo of esp-docs so it can generate directly into `/_static/v2`


balloob:
This is really cool, great work! 

Before we merge this, let's make sure we transfer the `esphome-webserver` repository to the ESPHome organization. 

I'm also wondering if it would make sense to automatically build the frontend when the docs are build instead of including generated builds in the repository? Although there might not be a lot of changes all the time.

wilberforce:
> This is really cool, great work!
@balloob 
Thanks, I'm pretty chuffed with it ;-)

> Before we merge this, let's make sure we transfer the `esphome-webserver` repository to the ESPHome organization.

I have updated the repo with more documentation:
https://github.com/wilberforce/esphome-webserver

> I'm also wondering if it would make sense to automatically build the frontend when the docs are build instead of including generated builds in the repository? Although there might not be a lot of changes all the time.
I'm wondering if we should set a new `/esphome/esphome-webserver` up as submodule of `/esphome/esphome/docs`.

If it was at:
/esp-webserver then the build could then be setup to output to ../static/v2

It could be set up to build as part of the generated builds (I'm not clear on how to do this at this stage).

What is involved to transfer the repo to esphome?  Will I still have write access to the repo?

There are updates I need to bring this pull request and the web_server component uptodate  - I can work with @jesserockz on that.

balloob:
I am personally not a fan of submodules. I was just thinking that we shallow clone the repository during build time and create a build.

Yes, you will retain write access to the repository. We just want to have it under our org so a) we can do hot fixes if something is bad and b) if you disappear we're not in trouble.

balloob:
Actually thinking now, why do we include this in the docs? We could have this be a standalone domain to host this, automatically build by Netlify on each commit. What do you think @jesserockz ? `https://local.esphome.io` or something

jesserockz:
> Actually thinking now, why do we include this in the docs? We could have this be a standalone domain to host this, automatically build by Netlify on each commit. What do you think @jesserockz ? `https://local.esphome.io` or something

Genius!! Yeah we can just host it from the now transferred webserver repo which can build and deploy itself to its own netlify project.

wilberforce:
> > Actually thinking now, why do we include this in the docs? We could have this be a standalone domain to host this, automatically build by Netlify on each commit. What do you think @jesserockz ? `https://local.esphome.io` or something
> 
> Genius!! Yeah we can just host it from the now transferred webserver repo which can build and deploy itself to its own netlify project.

This would fit with having optional components and also i18n language support in the near future.

By the way, this release has `index.html.gz` which is completely self contained and embedding the gz/css/svg (10K). I need to add an option to the webserver.cpp to use this file as option so that as device can be deployed with this code and not require any local or internet file requests.

components/web_server.rst
```diff
@@ -78,28 +78,38 @@ Configuration variables:
...
+    .. code-block:: yaml
+
+        # Example configuration entry V2
+        web_server:
+          port: 80
+          js_include: "./v2/www.js"
+          js_url: ""
+          version: 2        
           
+        Copy https://oi.esphome.io/v2/www.js to a V2 folder in your yaml folder.
```
wilberforce: @jesserockz Wasn't sure about  how to reference copying the file to use locally?  The V1 example assumes a local install of esphome-docs.

components/web_server.rst
```diff
@@ -78,28 +78,38 @@ Configuration variables:
 
         # Example configuration entry
         web_server:
-          port: 80
+          port: 80
```
balloob: Please strip whitespace

wilberforce: Done. I was discussing with @jesserockz yesterday - what license should the esphome-webserver have - I've currently assigned MIT but you might have a better suggestion?

balloob: MIT is fine for me. I usually do Apache but it doesn't matter.

components/web_server.rst
```diff
@@ -52,6 +57,7 @@ Configuration variables:
 
 - **ota** (*Optional*, boolean): Turn on or off the OTA feature inside webserver. Strongly not suggested without enabled authentication settings. Default: `true`
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
+- **version** (*Optional*, string): 1 or 2. Version 1 displays as a table. Version 2 uses web components and has more functionality
```
balloob: Should we mention the default?

components/web_server.rst
```diff
@@ -65,22 +71,54 @@ Configuration variables:
           auth:
             username: admin
             password: !secret web_server_password
+            
+    Example web_server configuration using version 2:
```
balloob: This makes it sound like version 2 is not the default. It's the new and modern version and we should promote that. Why do we even still include the v1 ?

components/web_server.rst
```diff
@@ -65,22 +71,54 @@ Configuration variables:
...
           port: 80
           ota: false
           css_include: "../../../esphome-docs/_static/webserver-v1.min.css"
-          css_url: ""
           js_include: "../../../esphome-docs/_static/webserver-v1.min.js"
           js_url: ""
 
+    Example web_server version 2 configuration with JS included from a local file.
+    
+    CSS and JS URL's are set to empty value, so no internet access is needed for this device to show it's web interface.
```
balloob: Can we put this in a section with "here by dragons" or something. We want most users just to enable the default web server and have it work.

components/web_server.rst
wilberforce: Yes - currently 1  - see below

components/web_server.rst
wilberforce: @jesserockz  @balloob 
Hi,  Currently version 1 is the default and you get the existing behavour.  I didn't want to force V2, although perhaps this is the better choice. In breaking changes we could mention if you can select back to version 1?

I'm happy to make 2 the default if you guys are happy!

components/web_server.rst
wilberforce: Yes - it's an advanced option - you are better off using the defaults or even the `local` option that requires no external files

wilberforce: re-arranged and simplified

jesserockz (changes requested):
Can you update the screenshot based on todays changes?

jesserockz (approved):

PR #1614, opened 2021-11-12
Add new on_state trigger info

dudanov:
## Description:

Add some info about new ``on_state`` trigger for climate components.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2707

## Checklist:

### DIFF
--- a/components/climate/index.rst
+++ b/components/climate/index.rst
@@ -184,6 +184,20 @@ advanced stuff.
       // etc. see API reference
       call.perform();
 
+.. _climate-on_state_trigger:
+
+``climate.on_state`` Trigger
+******************************************************
+
+This trigger is activated each time the climate device is updated (using publish_state).
+
+.. code-block:: yaml
+
+    climate:
+      - platform: midea  # or any other platform
+        # ...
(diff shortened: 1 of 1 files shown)


components/climate/index.rst
```diff
@@ -180,6 +180,20 @@ advanced stuff.
       // etc. see API reference
       call.perform();
 
+.. _climate-on_state_trigger:
+
+``climate.on_state`` Trigger
+******************************************************
+
+This trigger is activated each time the climate device is updated (using publish_state).
```
oxan: `publish_state` is an internal implementation detail that doesn't mean much to our users.
```suggestion
This trigger is activated each time the state of the climate device is updated 
(for example, if the current temperature measurement or the mode set by the users changes).
```

dudanov: Done!

oxan:
Oh, this also should've targetted `next` instead of `current` (as it's documentation for a new PR). I've fixed it for now as I already merged the associated code PR, but please use the correct target next time :)

PR #1640, opened 2021-11-19
Update uart.rst

peterforeman:
Docs mention "sequence" but it should be "delimiter".

## Description:

## Checklist:


mmakaay:
Ah yes, that option was renamed during development to not result in confusion with the action sequence propery.
Thanks for spotting and fixing it!

PR #1645, opened 2021-11-21
Set name parameter as optional

davet2001:
## Description:

## Checklist:

### DIFF
--- a/components/binary_sensor/homeassistant.rst
+++ b/components/binary_sensor/homeassistant.rst
@@ -30,7 +30,7 @@ With Home Assistant 2021.6 or newer, entity state attributes can also be importe
 Configuration variables:
 ------------------------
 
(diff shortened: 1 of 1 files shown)


oxan:
If we do this, we shouldn't do it only on this page, but everywhere (there's hundreds more of these). I'm not sure whether that's something we want to do though, not supplying the name makes `id` required (which is also marked as optional), and the entity becomes internal, which is fairly subtle behaviour and not something we want to happen unconsciously.

glmnet:
Here are other ideas https://github.com/esphome/feature-requests/issues/600

davet2001:
Ok, I was not aware that this was a wider issue. I noticed that I had not assigned `name`, yet my yaml  was still accepted by the validator and it worked (so I assumed just a typo in the doc).

I will add my suggestion to [#600](https://github.com/esphome/feature-requests/issues/600).

PR #1648, opened 2021-11-23
Add output pin definition to hbridge fan example configuration

Lerosen:
To better understand how to use the hbridge fan component, the output pin definitions should be added to the example configuration to make it complete. Otherwise if the user just copy and pasted the example and fills in the pins directly (D1, D2, etc.), he will get an error and not directly notice what he is missing.

## Description:

## Checklist:

  - [ X] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/components/fan/hbridge.rst
+++ b/components/fan/hbridge.rst
@@ -23,6 +23,14 @@ The `'hbridge`' fan platform allows you to use a compatible `h-bridge` (L298N, D
 .. code-block:: yaml
 
     # Example configuration entry
+    output:
+      - platform: esp8266_pwm
+        id: motor_forward_pin
+        pin: D1
+      - platform: esp8266_pwm
+        id: motor_reverse_pin
+        pin: D2
+    
     fan:
       - platform: hbridge
         id: my_fan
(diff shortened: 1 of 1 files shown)


components/fan/hbridge.rst
```diff
@@ -23,6 +23,14 @@ The `'hbridge`' fan platform allows you to use a compatible `h-bridge` (L298N, D
...
 
     # Example configuration entry
+    output:
+      - platform: esp8266_pwm
+        id: motor_forward_pin
+        pin: D1
+      - platform: esp8266_pwm
+        id: motor_reverse_pin
+        pin: D2
+
```
jesserockz: Would prefer not to mention the specific platform and use GPIO pins not board aliases

```suggestion
    output:
      - platform: ...
        id: motor_forward_pin
        pin: GPIO5
      - platform: ...
        id: motor_reverse_pin
        pin: GPIO4
    
```

Lerosen:
Thanks for your feedback @jesserockz , I agree with your suggestion. Is there any documentation for such best practices when documenting a component?

jesserockz:
> Thanks for your feedback @jesserockz , I agree with your suggestion. Is there any documentation for such best practices when documenting a component?

No documentation for the documentation practices. Or even code contributing really. ESPHome really needs a developer docs site

PR #1670, opened 2021-11-26
Add note about hlw8012 BL0937 SEL pin

michaelmeller:
According to the datasheet of BL0937 (https://www.belling.com.cn/media/file_object/bel_product/BL0937/datasheet/BL0937_V1.02_en.pdf) if SEL=0 current is measured, otherwise voltage. This has to be accommodated by inverting the SEL pin function. Hint in documentation makes this clear.

## Description:

## Checklist:


jesserockz (approved):
Thanks

PR #1671, opened 2021-11-26
add  I2S PCM5102A & dlna(or another network audio player) for esp32

ananyevgv:
Hi. I would like to get support for audio devices over the i2c bus, with the ability to create a network audio player


oxan:
You can file feature requests with an issue in the [esphome/feature-requests](https://github.com/esphome/feature-requests/issues) repository, this isn't the correct place

PR #1676, opened 2021-11-28
Enhance Template Number example

ferbar:
## Description:

Just enhancing the example a bit.

## Checklist:

### DIFF
--- a/components/number/template.rst
+++ b/components/number/template.rst
@@ -13,11 +13,19 @@ using :ref:`lambdas <config-lambda>`.
     # Example configuration entry
     number:
       - platform: template
-        name: "Template number"
-        optimistic: true
+        name: "Changeable Global Variable"
         min_value: 0
         max_value: 100
         step: 1
+        lambda: |-
+          float value = id(sensor).getValue();
+          ESP_LOGD("volume", "returning value %f, value);
+          return value;
+        set_action:
+          lambda: |-
+            ESP_LOGD("sensor", "setting value %f", x);
+            id(sensor).setValue(x);
+
 
 Configuration variables:
 ------------------------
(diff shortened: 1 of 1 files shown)


oxan:
I would prefer it if the default example keeps using optimistic mode, as that's (in my experience) what most people actually want.

ferbar:
@oxan: If I'm right you are either using the template number with optimistic + initial_value + set_action, when you won't read a value or you are using it with lambda + set_action when you want feedback. If so, I would suggest 2 examples since they give a better clue what can be done with the template number.

```
# Example configuration entry for a write - only number:
number:
  - platform: template
    name: "Template number"
    optimistic: true
    initial_value: 50
    min_value: 0
    max_value: 100
    step: 1
    set_action:
      [something]

# Example configuration entry, currend value read by a lambda function
  - platform: template
    name: "Changeable Global Variable"
    min_value: 0
    max_value: 100
    step: 1
    lambda: |-
      [something]
    set_action:
      [something]
```

oxan:
Yeah, two examples seems like a good idea.

ferbar:
@oxan I found out that the best usage for the Template Number is probably the add_on_state_callback function in the C++ code which I noted in the example. Have a look at it please. Thanks!

components/number/template.rst
```diff
@@ -10,14 +10,32 @@ using :ref:`lambdas <config-lambda>`.
 
 .. code-block:: yaml
 
-    # Example configuration entry
+    # Example configuration entry for setting a value, either in combination with the set_action configuration or the add_on_state_callback() method in a custom component.
```
jesserockz, line 13: I don't like this addition exposing information for custom components here.
Anyone developing custom components should be looking at the api docs themselves, not the configuration docs.

PR #1701, opened 2021-12-05
Clarification of the behavior of calibraty_linear filter

swifty99:
## Description:
I was misunderstood, how calibrate linear works. I found others having the same trouble. Here is my first contribution to the documentation. So, please be gentle. ;-)

**Related issue (if applicable):** fixes (https://github.com/esphome/issues/issues/2768)

## Checklist:

### DIFF
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -191,7 +191,10 @@ the value the sensor shows.
 
 The arguments are a list of data points, each in the form ``MEASURED -> TRUTH``. ESPHome will
 then fit a linear equation to the values (using least squares). So you need to supply at least
-two values.
+two values. If more than two points are specified, they may not lie on a straight line. In this 
+case, the calculated results often do not correspond to the specified values. Especially for 
+non-linear relationships, the result can deviate strongly from the expected value. It is 
+recommended to use the filter calibrate linear only if two given data points are sufficient.
 
 .. _sensor-calibrate_polynomial:
 
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -191,10 +191,9 @@ the value the sensor shows.
 
 The arguments are a list of data points, each in the form ``MEASURED -> TRUTH``. ESPHome will
 then fit a linear equation to the values (using least squares). So you need to supply at least
-two values. If more than two points are specified, they may not lie on a straight line. In this 
-case, the calculated results often do not correspond to the specified values. Especially for 
-non-linear relationships, the result can deviate strongly from the expected value. It is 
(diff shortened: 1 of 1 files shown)


swifty99:
I was misunderstood, how calibrate linear works. I found others having the same trouble. Here is my first contribution to the documentation. So, please be gentle. ;-)

flaviut (changes requested):
I'm not sure what this is saying, but I believe it should:

- either be shortened to "So you must supply at exactly two values."
- describe the actual behavior when more than two values are passed

swifty99:
It is allowed to use more than two values. Which was confusing me, that could be changed on code side.

If that change is unwanted how about this:
"if more than two values are given an exact representation on each value will not be found unless they form a line."
or
"if more than two values are given a linear solution will be calculated and will not represent each value exactly"

components/sensor/index.rst
```diff
@@ -191,7 +191,10 @@ the value the sensor shows.
 
 The arguments are a list of data points, each in the form ``MEASURED -> TRUTH``. ESPHome will
 then fit a linear equation to the values (using least squares). So you need to supply at least
-two values.
+two values. If more than two points are specified, they may not lie on a straight line. In this 
+case, the calculated results often do not correspond to the specified values. Especially for 
+non-linear relationships, the result can deviate strongly from the expected value. It is 
+recommended to use the filter calibrate linear only if two given data points are sufficient.
```
flaviut:
```suggestion
two values. If more than two points are specified, the line of best fit may not pass through any of
the specified values, and the result may be very different from the expected value. Therefore, it is
recommended to use the filter calibrate linear only if two given data points are sufficient.
```

What do you think about this phrasing?

flaviut: Oh! I like the "if more than two values are given a linear solution will be calculated and will not represent each value exactly", I didn't re-read it when I first came back to this.

How about

```suggestion
two values. If more than two values are given a linear solution will be calculated and may not
represent each value exactly. It is recommended to use the filter calibrate linear only if two data
points are sufficient.
```

oxan: > It is recommended to use the filter calibrate linear only if two data points are sufficient.

Maybe it's my physics background, but I don't fully agree with this. You should only use `calibrate_linear` if the data you're trying to calibrate (and thus the data points you've sampled) is linear, but it's absolutely fine to use more than two data points.

flaviut: Removed that sentence here: https://github.com/esphome/esphome-docs/pull/1830

The sentence explaining that it won't always pass through both points has been left in.

oxan: Excellent.

PR #1724, opened 2021-12-15
Clarify Wireless Install requirements

robclark56:
## Description:
Update to clarify that Wireless Install requires the Device to be on same subnet as the Home Assistant computer. (and device must already have ESPHome intsalled)

## Checklist:


oxan:
It doesn't have to be on the same subnet, it must be reachable (and possibly mDNS-resolvable) from the PC running ESPHome.

jesserockz:
Yeah as above, mine are in two different vlans =)

If you have different subnets then you should know what you are doing network wise and we do not need to mention it.

flaviut:
Closing for the above reasons.

Thanks for taking the time to improve this though!

PR #1749, opened 2021-12-25
Add documentation for enable_ipv6 flag.

HeMan:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#2953

## Checklist:

### DIFF
--- a/components/network.rst
+++ b/components/network.rst
@@ -18,5 +18,5 @@ networks (WiFi, Ethernet).
 Configuration variables:
 ------------------------
 
-- **enable_ipv6** (*Optional*, boolean): Enables IPv6 support. Defaults to ``false``. Only available on ESP32 with esp-idf platform.
+- **enable_ipv6** (*Optional*, boolean): Enables IPv6 support. Defaults to ``false``. Only available on ESP32 with ESP-IDF framework.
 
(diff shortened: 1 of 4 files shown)


components/network.rst
```diff
@@ -0,0 +1,22 @@
...
+.. code-block:: yaml
+
+    # Example configuration
+    network:
+        enable_ipv6: true
+        
+Configuration variables:
+------------------------
+
+- **enable_ipv6** (*Optional*, boolean): Enables IPv6 support. Defaults to ``false``.
```
jesserockz: This should say it requires the ESP-IDF framework.

HeMan: Good catch! Updated it now.

components/network.rst
```diff
@@ -0,0 +1,22 @@
...
+.. code-block:: yaml
+
+    # Example configuration
+    network:
+        enable_ipv6: true
+        
+Configuration variables:
+------------------------
+
+- **enable_ipv6** (*Optional*, boolean): Enables IPv6 support. Defaults to ``false``. Only available on ESP32 with esp-idf platform.
```
oxan:
```suggestion
- **enable_ipv6** (*Optional*, boolean): Enables IPv6 support. Defaults to ``false``. Only available on ESP32 with ESP-IDF framework.
```

oxan (changes requested):

PR #1751, opened 2021-12-26
Update rotary encoder doc

aquaticus:
Added extra options example for encoder pin configuration.


flaviut (approved):

flaviut:
Looks very helpful, the composability of various options isn't always obvious.

PR #1758, opened 2021-12-29
Update index.rst to add instructions on rotating text.

granvillebarker:
Added instructions for using set_rotation to print rotated text.

## Description:

## Checklist:

### DIFF
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -372,6 +372,27 @@ Configuration variables:
 
 RGB displays use red, green, and blue, while grayscale displays may use white.
 
+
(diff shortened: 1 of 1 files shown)


oxan:
The `set_rotation` method is intended to be used by the `rotation` configuration option, and not directly from lambdas (it's also marked as an internal method). I'm not sure we want to document using it like this, since that might break at some point, and it's not a nice API anyway.

flaviut:
Thanks for taking a look into this! Going to go ahead and close this, since it doesn't sound like we really want folks to use this API until it is improved.

PR #1763, opened 2021-12-30
Update mirabella-genio-bulb.rst

skandiah94:
The cwww2 custom component is no longer required to control the Mirabella Genio CWWW downlights, I managed to get the downlights working using the standard cwww component in ESPHome 2021.12.1

## Description:

## Checklist:


flaviut:
I think you've made the changes to the wrong configuration, but I'm not convinced they'd be correct. It looks like the cwww2 component allows you to control a brightness & color temperature output separately, which is different than what the cwww component requires.

Could this be a new hardware revision?

PR #1765, opened 2021-12-31
Add SH1107 128x64 to SSD1306 Display documentation

arunderwood:
## Description:

* Add documentation for the `SH1107_128x64` model on the `ssd1306_i2c` platform https://github.com/esphome/esphome/pull/2967
* Correct an error in the documentation, the [default polling interval is currently `1 second`](https://github.com/esphome/esphome/blob/e712e87d61adae8561f12a20f90526469078f2e2/esphome/components/ssd1306_base/__init__.py#L69)

https://github.com/esphome/esphome/pull/2967

## Checklist:


components/display/ssd1306.rst
```diff
@@ -56,6 +56,7 @@ Configuration variables:
   - ``SH1106 128x64``
   - ``SH1106 96x16``
   - ``SH1106 64x48``
+  - ``SH1107 128x64``
```
jesserockz, line 59: Want to add this to the SPI list below too?

arunderwood: I went back and forth on this.  Based on how the ssd1306 component is written, theoretically the changes I have made to enable the use of a SH1107 over I2C also enabled the use over SPI, however [the screen](https://www.adafruit.com/product/4650) I'm testing with doesn't have the SPI interface populated, so the changes are untested.

I'm fine with including this but I'll leave this note here for future people letting them know that I wasn't able to test `SH1107 128x64` over SPI on real hardware.

jesserockz: Thats fine, the screen driver (according to its datasheet) does support SPI so might as well just have it there just in case there is a module out there.

jesserockz (approved):

PR #1767, opened 2021-12-31
Sensor for ltr390 is uv, not uvi

arunderwood:
## Description:

The `uvi` sensor in the example is invalid, the correct key is `uv`.

[Definition](https://github.com/esphome/esphome/blob/b20760c93c3f8874e0d73d515e0e88d04a7c87aa/esphome/components/ltr390/sensor.py#L25)
[Passing test](https://github.com/esphome/esphome/blob/dev/tests/test2.yaml#L258-L260)

## Checklist:


oxan (approved):
Thanks!

PR #1768, opened 2022-01-03
Fix documentation to use correct values for the 'inverted' restore modes.

MrHacky:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#1810
--> Was the original PR that introduced the 'inverted' restore modes, but somehow the docs got added with mismatching names it seems.

## Checklist:

    --> 'current' as it is a fix
    --> N/A as nothing new is added


oxan (approved):
Thanks!

PR #1769, opened 2022-01-04
Document sensor id and how it relates to the name

Cat-Ion:
## Description:

Document the existance of `id` and that `name` is not required if `id` exists

**Related issue (if applicable):** None

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** None

## Checklist:

### DIFF
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -38,7 +38,8 @@ override them if you want to.
 
 Configuration variables:
 
-- **name** (**Required**, string): The name for the sensor.
+- **id** (*Optional*, string): Manually specify the ID for code generation. At least one of **id** and **name** must be specified.
+- **name** (*Optional*, string): The name for the sensor. At least one of **id** and **name** must be specified.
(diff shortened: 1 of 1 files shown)


mbo18:
If `id` is added here, should it be removed from all sensors documentation where it is written `All other options from Sensor. `?
Like here: https://esphome.io/components/sensor/sgp30.html
Or here: https://esphome.io/components/sensor/homeassistant.html

Cat-Ion:
Probably, yes. At least that's certainly less work than making everything have both options, and making sure that they're documented the same way.

flaviut:
@Cat-Ion Would you be willing to make that change?

Cat-Ion:
Yes, but I'm going to need a bit because I'm absolutely swamped at work right now.

OttoWinter:
No, I don't think it should be removed elsewhere. We have the most important options in each sensor block, and ID is something more users may want to use to chain together things etc.

OttoWinter (approved):
Thanks!

PR #1772, opened 2022-01-05
Add docs to pzemac on how to change its address

abmantis:
## Description:

Most tutorials or suggestions online point to using either the official PZEM tool (windows only) or Tasmota. 
With the suggested snipped it is possible to do it using ESPHome.

## Checklist:

### DIFF
--- a/components/sensor/pzemac.rst
+++ b/components/sensor/pzemac.rst
@@ -83,6 +83,63 @@ Configuration variables:
   the same UART bus. You will need to set the address of each device manually. Defaults to ``1``.
 - **modbus_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the Modbus hub.
 
+
+Changing the address of a PZEM-004T:
+------------------------
+
+You can use the following configuration to change the address of a sensor.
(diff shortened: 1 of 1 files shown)


flaviut (approved):
This is cool, great idea figuring this out!

glmnet:
I've got it done with this snippet, did it some time ago. Sorry for not sharing before

```yaml

modbus_controller:
  address: 248 # broadcast address, you can communicate with any connected device regardless of it address

number:
  - platform: modbus_controller
    id: pzem_addr
    name: "pzem_addr"
    address: 0x2 # this is the address registry number, and it is not related to device address, do not change
    value_type: U_WORD


api:
  services:
    service: set_addr
    variables:
      address: int
    then:
      number.set:
        id: pzem_addr
        value: !lambda "return address;"

```

I'm not completely sure if the service is required (again did this a month ago). Had  to do some "fixing" in the modbus number, but worked ok.

BTW, this also queries the address from the device.

flaviut:
Hey @glmnet, I'm not sure I understand your snippet, and I'm not familiar with this protocol.

If it's a shorter & simpler way to set this the address, a new PR that updates the docs with it & a brief explanation would be greatly appreciated!

PR #1778, opened 2022-01-06
Use invert_colors and use_bgr in leak detector recipe

shish:
## Description:

No need for a "wonky color fix, in lieu of finding a way to invert the display" - the st7735 display component includes native BGR and Invert modes

(Every time I start a new m5stickC project I come back to copy-pasting from this cookbook, so it seems nice to have it using current best-practices :) )

## Checklist:


flaviut (approved):

oxan (approved):

oxan:
Thanks!

PR #1784, opened 2022-01-08
Update esp32_camera.rst

pixeldoc2000:
Troubleshooting Section added.

## Description:

## Checklist:


flaviut:
I don't feel like this fits in the documentation unless a root cause is found. It feels too much like flinging stuff at the wall to see what sticks.

Thank you for taking the time to submit this change though, it's always appreciated!

PR #1786, opened 2022-01-08
Add live reloading

flaviut:
## Description:

Replace the existing webserver target with a live-reloading target.

Adds a dependency on https://github.com/executablebooks/sphinx-autobuild

I considered using something like modd/devd, but that would have caused difficulty when working with the various architectures the docker container targets.

## Checklist:


oxan (approved):
Nice!

PR #1791, opened 2022-01-12
Update dashboard screenshots

jesserockz:
## Description:

- Update dashboard screenshots
- Update getting started text for dashboard

## Checklist:


guides/getting_started_hassio.rst
```diff
@@ -30,55 +28,43 @@ After that, wait a bit until it is installed (this can take a while). Click "Sta
...
 creating your first configuration.
 
-.. figure:: images/hassio_start.png
+.. figure:: images/dashboard_empty.png
     :align: center
     :width: 95.0%
 
 The wizard will guide you through creating your first configuration and, depending on your browser, install it on your ESP device. You will need to name your configuration and enter your wireless network settings so that the ESP device can come online and can communicate with Home Assistant.
 
-.. raw:: html
```
flaviut, line 39: This seems helpful, why remove it? There's also a couple external users: https://www.google.com/search?q=getting_started_hassio.html%23webserial

guides/contributing.rst
```diff
@@ -123,13 +123,13 @@ documents establish the following character order for better consistency.
 
   .. code-block:: rst
 
-      .. figure:: images/dashboard.png
```
flaviut, line 126: the new image is wayyy better.

guides/getting_started_hassio.rst
```diff
@@ -30,55 +28,43 @@ After that, wait a bit until it is installed (this can take a while). Click "Sta
...
 
-- **SHOW LOGS**: With this command you can view all the logs the node is outputting. If a USB device is
-  connected, it will attempt to use the serial connection. Otherwise it will use the built-in OTA logs.
+- **Overflow menu**: This is a dropdown menu which allows you to perform more actions.
 
-- **COMPILE**: This command compiles the firmware and gives you the option of downloading the generated
-  binary so that you can install it yourself from your computer using :ref:`ESPHome-flasher <esphome-flasher>`.
+  - **Validate**: This will validate the configuration file.
+  - **Install**: Opens the Install dialog.
+  - **Clean Build Files**: This will delete all of the generated build files and is a safe operation ot perform. Sometimes old files are not removed and can cause compile issues later on.
```
flaviut: I think pointing out that this is a safe operation is a good idea. I would be good to point out Delete below is not a safe operation. Unless there's a confirmation dialog in the UI that already does that.

guides/getting_started_hassio.rst
jesserockz: Because if you are following the getting started and using the dashboard, then basically all the same note text is now contained inside there.

guides/getting_started_hassio.rst
jesserockz: Found a typo =)
```suggestion
  - **Clean Build Files**: This will delete all of the generated build files and is a safe operation to perform. Sometimes old files are not removed and can cause compile issues later on.
```

jesserockz: ~~Good point about the delete~~

There is a confirmation dialog when deleting

flaviut (approved):

jesserockz:
Will merge this alongside the 2022.1 release

PR #1792, opened 2022-01-12
Fix MAX31856 image

DAVe3283:
## Description:

The wrong image was selected for the MAX31856 (it was using the MAX31865 image), likely due to a simple typo. This switches to the correct image.

## Checklist:


flaviut:
I wouldn't have seen that, good catch. Thanks!

PR #1793, opened 2022-01-12
Add ektf2232 touchscreen controller

jesserockz:
## Description:

A touchscreen controller used in the inkplate 6 plus device

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3027

## Checklist:


mkocus:
@jesserockz Found an error in the documentation - ektf2232 component has variable `on_touch`, not `on_press` - according to the PR.

Also, it would be good to use the sample code as in the test5.yaml:
```
ektf2232:
  interrupt_pin: GPIO36
  rts_pin: GPIO5
  rotation: 90
  on_touch:
    - logger.log:
        format: Touch at (%d, %d)
        args: ["touch.x", "touch.y"]
```

Btw, I cannot figure which is rts_pin according to the schematic. Is it 5 (test) or 16 (doc)?

jesserockz:
@mkocus I have fixed that, thanks for bringing it up.

The rts pin is actually neither for the inkplate. It is a pin on the mcp23017 expander

```yaml
  rts_pin:
    mcp23xxx: mcp23017_hub
    number: 10
```

I just didn't want to include an MCP23xxx example in this component docs, but when I update the 6 plus docs I will add it there.

mkocus:
Ok, now i'm nearly certain that `rts_pin` should use the value from TS_RTS (`Touch.cpp`), which is 10.

PR #1798, opened 2022-01-13
Enhance documentation

Micha031982:
## Description:
Enhance Deep sleep Documentation 
I searched for implemtations of deep sleep and found this Solution 

I found the linked Artikel and from my point of view this is a good Solution 
So maybe it could be added as a Solution 

https://tatham.blog/2021/02/06/esphome-batteries-deep-sleep-and-over-the-air-updates/#comments

Sorry if this is not structured or at the wrong Place this is my First pull 

**Pull request in [esphome](https://github.com/esphome/esphome) 

## Checklist:

  - [ next]
  - [ https://tatham.blog/2021/02/06/esphome-batteries-deep-sleep-and-over-the-air-updates/#comments]


flaviut:
I'm assuming this is for https://github.com/esphome/esphome-docs/pull/1798/commits/52fa637f09e0b1f379affb99886bddd64f2465b4.

Looks like it wasn't included in the last release by mistake, but there's a new release soon and it'll show up then. Thanks for pointing this out, but let's just sit tight for a couple weeks and it'll all work itself out!

PR #1809, opened 2022-01-17
Specified how the sensor values get populated

spacemanspiff2007:
## Description:

Specified how the sensor gets its value since the lack of interval was really confusing for me

## Checklist:


flaviut:
I've reworded this to not mention the main loop (which is a bit of an implementation detail), and to avoid being redundant with the previous sentence. Thanks for your contribution!

PR #1812, opened 2022-01-18
Update tuya.rst

sbur83:
Zemismart YH002 confirmed working with the same settings as ZM79E-DT
Added model alongside the other zemismart chain motor in Tuya Component document


jesserockz (approved):
Thanks

PR #1825, opened 2022-01-20
Update xiaomi_ble.rst

JeeCee1:
Added ´´´The easiest method to retrieve the bindkey from the cloud is to use the Cloud Tokens Extractor, written by one of Home Assistant users. If you prefer to not use the executable, read here.´´ and a link under "See Also"

## Description:

## Checklist:


flaviut:
Thanks, this seems like a great alternative solution!

PR #1826, opened 2022-01-20
Update mqtt to reflect changes in 2022.1

leakypixel:
`root` is no longer passed by reference; update the documentation to reflect this.

Related issue: https://github.com/esphome/issues/issues/2960

### DIFF
--- a/components/mqtt.rst
+++ b/components/mqtt.rst
@@ -416,7 +416,7 @@ Configuration variables:
 
     .. code-block:: cpp
 
-        id(mqtt_client).subscribe_json("the/topic", [=](JsonObject root) {
+        id(mqtt_client).subscribe_json("the/topic", [=](const std::string &topic, JsonObject root) {
             // do something with JSON-decoded value root
         });
 
--- a/components/mqtt.rst
+++ b/components/mqtt.rst
@@ -416,7 +416,7 @@ Configuration variables:
 
     .. code-block:: cpp
 
-        id(mqtt_client).subscribe_json("the/topic", [=](JsonObject &root) {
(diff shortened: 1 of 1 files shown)


components/mqtt.rst
```diff
@@ -416,7 +416,7 @@ Configuration variables:
 
     .. code-block:: cpp
 
-        id(mqtt_client).subscribe_json("the/topic", [=](JsonObject &root) {
+        id(mqtt_client).subscribe_json("the/topic", [=](JsonObject root) {
```
flaviut: I'm no C++ expert, but it looks like this lambda also ought to take a `const std::string &` parameter, and that's been the case since forever: https://github.com/esphome/esphome/commit/33f17f75a0d03c6a550597fd902479779757346e#diff-bf3517861bff7a37f472c325cd64d7c771ddd8e47d61496e70d8de55c3d790efL23

Am I missing something?

oxan: Looks like you're right to me.

leakypixel: Sorry, I don't understand - I ran into the same problem as in the linked issue and made the change recommended there, which solved the problem for me. I'd been using the example in the docs, so after seeing that the change was needed to make things work I thought it best to update the docs - but I don't really understand beyond that it's a change from pointer to value as described in the linked issue and changelog.

The related change in the changelog: https://esphome.io/changelog/2022.1.0.html#json

flaviut:
Looks like there's been a long-standing issue with `.subscribe_json` in the docs. I've fixed it & merged, thanks for taking the time to submit this PR!

PR #1872, opened 2022-02-01
fixed example lambda key

depuits:
## Description:

The example uses the `lambda` key for the lambda which is not supported in the output component. This key is changed to `write_lambda`. The example itself should be checked it it still makes sense.

## Checklist:


OttoWinter (approved):
Thanks!

PR #1883, opened 2022-02-06
Switch modbus controller register_type

depuits:
Switch modbus controller replaced `modbus_functioncode` with `register_type` as in the example code. Added explanation for write lambda with no return value.

## Description:

## Checklist:

### DIFF
--- a/components/switch/modbus_controller.rst
+++ b/components/switch/modbus_controller.rst
@@ -13,7 +13,7 @@ Configuration variables:
 
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
 - **name** (**Required**, string): The name of the sensor.
-- **modbus_functioncode** (**Required**): type of the modbus register.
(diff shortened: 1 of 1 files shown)


components/switch/modbus_controller.rst
```diff
@@ -41,7 +41,8 @@ Configuration variables:
 Possible return values for the lambda:
 
  - ``return <true / false>;`` the new value for the sensor.
- - ``return <anything>; and fill payload with data`` if the payload is added from the lambda then these bytes will be sent
+ - ``return <anything>; and fill payload with data`` if the payload is added from the lambda then these bytes will be sent.
+ - *No return value* in the case the lambda handles the sending of the value itself.
```
jesserockz: This is actually what it should be
```suggestion
 - ``return {};`` in the case the lambda handles the sending of the value itself.
```

jesserockz (approved):
Thanks

PR #1914, opened 2022-02-16
Example to parse values from JSON response to HTTP Request

nagyrobi:
## Description:

## Checklist:

### DIFF
--- a/components/http_request.rst
+++ b/components/http_request.rst
@@ -234,13 +234,9 @@ If you want to retrieve the value for the ``vol`` key and assign it to a templat
         on_response:
           then:
             - lambda: |-
-                DynamicJsonDocument doc(1024);
-                DeserializationError error = deserializeJson(doc, id(http_request_data).get_string());
-                if (error)
-                  return;
-                float vol = doc["vol"];
-                id(number_player_volume).publish_state(vol);
-
+                json::parse_json(id(http_request_data).get_string(), [](JsonObject root) {
+                    id(number_player_volume).publish_state(root["vol"]);
+                });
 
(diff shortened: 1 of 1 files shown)


components/http_request.rst
```diff
@@ -14,6 +14,7 @@ The ``http_request`` component lets you make HTTP/HTTPS requests. First, you nee
     # Example configuration entry
     http_request:
       useragent: esphome/device
+      id: http_request_data
```
jesserockz: I don't like setting the `id` like this on the main example, especially when it is only used way down below in the new example...

nagyrobi: Removed

components/http_request.rst
```diff
@@ -217,6 +217,29 @@ as seen below.
         # Will send:
         # {"key": 42.0, "greeting": "Hello World"}
 
+GET values from a JSON body response
+************************************
+
+The server returns a response in a JSON object over HTTP like this:
+``{"type":"0","ch":"0","mode":"40","loop":"0","eq":"0","status":"play","curpos":"7133","offset_pts":"7133","totlen":"0","Title":"556E6B6E6F776E","Artist":"556E6B6E6F776E","Album":"556E6B6E6F776E","alarmflag":"0","plicount":"0","plicurr":"0","vol":"42","mute":"0"}``
```
flaviut: Can you please simplify this example to match those above? esphome.io domain, a very short data object, etc?

nagyrobi: I thought a real world working example is always better.

jesserockz: While that can be true, a short, concise, easy to understand example is always better for users.

nagyrobi: Done.

flaviut:
Thank you :)

PR #1918, opened 2022-02-17
Add bitmasks description and example to extract bits

nagyrobi:
## Description:

## Checklist:

### DIFF
--- a/components/modbus_controller.rst
+++ b/components/modbus_controller.rst
@@ -117,6 +117,90 @@ Technically there is no difference between the "inline" and the standard definit
         value_type: U_WORD
 
 
+Bitmasks
+--------
+
+Some devices use decimal values in read registers to show multiple binary states occupying only one register address. To decode them, you can use bitmasks according to the table below. The decimal value corresponding to a bit is always double of the previous one in the row. Multiple bits can be represented in a single register by making a sum of all the values corresponding to the bits.
+
++------------+------------------+-----------+-----------+ 
+| Alarm  bit | Description      | DEC value | HEX value |
++============+==================+===========+===========+ 
+| bit 0      | Binary Sensor 0  | 1         | 1         |
++------------+------------------+-----------+-----------+ 
+| bit 1      | Binary Sensor 1  | 2         | 2         |
++------------+------------------+-----------+-----------+ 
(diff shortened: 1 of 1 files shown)


components/modbus_controller.rst
```diff
@@ -117,6 +117,90 @@ Technically there is no difference between the "inline" and the standard definit
...
+| bit 12     | Binary Sensor 12 | 4096      | 1000      |
++------------+------------------+-----------+-----------+ 
+| bit 13     | Binary Sensor 13 | 8192      | 2000      |
++------------+------------------+-----------+-----------+ 
+| bit 14     | Binary Sensor 14 | 16384     | 4000      |
++------------+------------------+-----------+-----------+ 
+| bit 15     | Binary Sensor 15 | 32768     | 8000      |
++------------+------------------+-----------+-----------+ 
+
+For example, at read register ``15``, a decimal value of ``12288`` is the sum of ``4096`` + ``8192``, meaning the corresponding bits ``12` and ``13`` are ``1``, the rest a re ``0``.
```
flaviut: I'm not sure how this is relevant--if a bitmask is used to extract a single bit, why do we need to add them?

flaviut: poked a bit more at it, and it seems like `0b1000` works fine :)

do you think that'd be easier to explain? And if so, would you like to update this guide?

nagyrobi: Not to extract a single bit. One can extract multiple bits from the same register - and that's the trickiest part.

sure, but what more to explain?

the goal is to let the user know what a bitmask is and how to use it.
this kind of bitmask usage dates back to the '90s when memory and communication possibilities were tighter, nowdays it's unuisual and is getting less and less implemented as there's plenty of resources.
But since Modbus is a very ancient protocol this has to be explained to today's users in an extensive manner.

flaviut: I misunderstood what you were explaining here. Thanks for clarifying!

flaviut:
Thank you for taking the time to write this and send it in!

nagyrobi:
> Thank you for taking the time to write this and send it in!

I do my best to improve what I can.
See the openhasp project's examples.

PR #1922, opened 2022-02-18
Canbus ESP32 - Schemtics added

danielschramm:
Schematic images.

## Description:

The great new ESP32 CanBus component has complete documentation for professionals.
Most beginners might need some schematics.

I changed the code example to match the Schematics.
Adjusting the schematics would not have been as nice in the images.

Used current branch as it is an improvement for an existing component and doc.

## Checklist:


components/canbus.rst
```diff
@@ -150,20 +150,39 @@ ESP32 CAN Component
...
+**************
+
+5V Can transceivers are cheap and generate complient levels.
+If you power your board with 5V this is the preferred option.
+R501 is important to reduce the 5V Logic level down to 3.3V for the ESP32 otherwise the ESP32 gets damaged.
+(Alternative option: Voltage divider 330Ohm / 680Ohm.)
+
+.. figure:: images/canbus_esp32_5v.png
+    :align: center
+    :target: ../_images/canbus_esp32_5v.png
```
OttoWinter, line 177: What is this `target` for?

danielschramm: The target should make the Image clickable to show it unscaled. The normal view in the page ist scaled down and some texts are hard to read then.

components/canbus.rst
```diff
@@ -150,20 +150,39 @@ ESP32 CAN Component
...
         can_id: 4
         bit_rate: 50kbps
         on_frame:
           ...
 
+Wiring options
+**************
+
+5V CAN transceivers are cheap and generate compliant levels. If you power your
+board with 5V this is the preferred option. R501 is important to reduce the 5V
```
flaviut, line 171: I'm not super thrilled with the idea of abusing the ESP32's internal ESD diodes like this (with R501), it's using ESP32 out of spec even if it works fine and is likely to be reliable.

If you'd like to change this, let me know, otherwise I'll merge it as-is.

danielschramm: Just have a look in the line below there I mention the Voltage divider instead of just one resistor for more advanced Users.

Normally you can find the rating for the internal diodes between 2mA and 40mA in most Microcontroller manuals. Espressive ist known for not providing many data so I did not explicitly checked their datasheet. 
For SPI I use this schematic for years with the ESP.
I also gave the hint for more robust designs with transceivers with integrated level shifters, etc...

So for absolute beginners (who need this schematics) this is the easiest solution to get a reliable setup without damaging their devices or having to debug rise and fall times behind a Voltage divider. Having these beginners setting up System on a breadboard without Issues is my goal here. 
If you think of a low resistance voltage divider for 500kBit/s on a breadboard and the R to GND has not perfect contact the ESP exposes his magic smoke.

flaviut:
Thank you! This is a great improvement. And thanks for taking the time to explain your reasoning behind it.

PR #1960, opened 2022-03-12
Add simpler Home Assistant configuration for servo control

demichiel:
## Description:

@ssieb mentioned in the Esphome discord that the current way of controlling servos was a bit convoluted. Since Home Assistant 2021.7 we can use the number component in ESPHome to get and set the servo angle in Home Assistant. This has made any config in Home Assistant redundant as the number gets discovered automatically. This PR removes the Home Assistant config and replaces the section with a minimal ESPHome config.

## Checklist:


demichiel:
I think the same changes can be made for the stepper motor docs as well, but I don't have a stepper motor to try it out. Should be the same though.

ssieb (approved):

PR #1971, opened 2022-03-16
Updated to reflect changes in the sm2135 code

matika77:
## Description:

Updated to describe new options rgb_current and cw_current.

## Checklist:

### DIFF
--- a/components/output/sm2135.rst
+++ b/components/output/sm2135.rst
@@ -18,6 +18,7 @@ It is used in some smart light bulbs:
 
 - Calex Smart RGB Reflector LED lamp (GU10)
 - LSC Smart GU10
+- LSC Smart Ceiling Light RGB/CW
 
 To use the channels of this components, you first need to setup the
 global ``sm2135`` hub and give it an id, and then define the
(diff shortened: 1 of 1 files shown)


components/output/sm2135.rst
```diff
@@ -59,6 +62,8 @@ Configuration variables:
 -  **data_pin** (**Required**, :ref:`Pin Schema <config-pin_schema>`): The pin used for MOSI.
 -  **clock_pin** (**Required**, :ref:`Pin Schema <config-pin_schema>`): The pin which SCLK is
    connected to.
+-  **rgb_current** (**Required**): Sets the current for the RGB LEDs (in milliamperes)
```
BoukeHaarsma23: Could we make this optional and set a sane default?

matika77: the only sane default would be the minimum 10mA, i guess. and then the user could be wondering why the lights are so dim or not working at all. so i suggest to push the user to enter the correct values.

jesserockz:
As this is a feature matched with a PR in `esphome/esphome`, please target your PR to the `next` branch and rebase.

Thanks
Jesse

PR #1989, opened 2022-03-26
Refine sensor lambda filter docs

micw:
## Description:

Adds an undocumented features of sonesor lambda filters (return {} to skip a value) + example.

## Checklist:


micw:
Not stale!

nagyrobi (approved):

PR #1994, opened 2022-03-30
Add border-radius to images + dropshadow to tables

spacegaier:
## Description:

The tables always feel "hacky" too me and I realized big portion of that is the harsh inlayed pictures of boards, sensors, ...

I think we can somewhat mitigate that with rounded borders plus giving the table a more visible dropshadow (in fact I copied the one over that is used on the guide cards on the initial page).

image

Giving the tables itself also slightly rounded corners might help even further, but I did not want to change too much in one PR.

## Checklist:

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -20,17 +20,17 @@ div.sphinxsidebar {
 
 img.component-image {
   border: none;
-  vertical-align: middle;
   display: block;
   margin-left: auto;
   margin-right: auto;
-  width: 100%;
   height: 85px;
   object-fit: contain;
+  border-radius: 10px;
 }
 
 table.docutils {
   width: 100%;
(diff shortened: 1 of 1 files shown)


jesserockz:
The rounded borders are ok, but not sure I like the shadow on the tables.

balloob (approved):
Agree, looks better 👍

balloob:
I feel like it can use even more of a face lift, but we can iterate.

balloob:
Oh yeah, I guess the shadow is not perse an improvement. Let's revert the shadows 😅

spacegaier:
I personally feel that the shadows are a rather big improvement, since they result in visually clearer (more visually distinct) blocks which prevents the current "wall of text feel" by providing more visually distinct elements. But might just be a peorsnal feeling...

But overall I agree the page in general needs more changes.

PR #1998, opened 2022-04-03
Added note on WAK signal for CCS811

sumirati:
## Description:

**Related issue (if applicable):** fixes #2331 - update to CCS811 documentation to hint on the usual need to ground the WAK pin

## Checklist:


jesserockz (approved):
Thanks

PR #2002, opened 2022-04-07
Update automations.rst

sevorl:
Include the timeout option in the code example for wait_until.

## Description:

## Checklist:


jesserockz (changes requested):
Could you make this a second block of yaml, or just a separate action so people can see both versions? Thanks

sevorl:
added a 2nd block as suggested

jesserockz (approved):

PR #2004, opened 2022-04-09
Add newer HomeAssistant actions to list of all actions

jburns20:
## Description:
The `homeassistant.event` and `homeassistant.tag_scanned` actions are missing from this list of all actions that are available. See https://esphome.io/components/api.html for more info about these actions.

## Checklist:


jesserockz (approved):
Thanks

PR #2005, opened 2022-04-09
Fixing subnet mask

wilcomir:
The subnet mask provided in the example is not compatible with the IPs provided.

## Description:

## Checklist:

### DIFF
--- a/components/wifi.rst
+++ b/components/wifi.rst
@@ -24,7 +24,7 @@ dramatically improve connection times.
       manual_ip:
         static_ip: 10.0.0.42
         gateway: 10.0.0.1
-        subnet: 255.255.255.0
+        subnet: 255.0.0.0
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


wilcomir:
To clarify, the documentation was - at least for me - confusing. I use the whole 10.x.x.x/8 at home, and I glanced over the mask since I saw the 10.x address.
Perhaps my edit was a bit too quick - maybe I could add a couple of sections, one for the more traditional 192.168.x.x address, and one for the full 10.x.x.x subnet.
This is based on the assumption that most people who use 10.x.x.x addresses use a /8 mask, which is at least debatable of course :)

I did have to pull my sonoff out of the wall to fix this so I think that while the edit can be improved, we should change something.

jesserockz (changes requested):
This is not really correct to change as it is up to the user what their actual subnet is. I personally use 4 vlans
`10.1.1.0/24` `10.2.2.0/24` `10.5.5.0/24` and `10.6.6.0/24`

As mentioned in your comment, maybe better to change the entire example to use `192.168.1.0/24`

wilcomir:
Hey @jesserockz , I modified as you suggested. Cheers!

jesserockz (approved):
Thanks

PR #2018, opened 2022-04-13
Updating the ST7789 docs page to show limitation

cbhiii:
## Description:

Added text to let users know that this platform only supports 135x240 pixel displays correctly. It does not work for other dimension displays because it is hardcoded. A pull exists to fix this, but has not yet been included. https://github.com/fraxinas/esphome/tree/fix_st7789v

Looking to prevent other users from assuming and purchasing modules that will not work correctly with ESPHome.

**More info here:** [st7789v Display Resolution Hardcoded To 135x240](https://github.com/esphome/issues/issues/1335)

## Checklist:


cbhiii:
Holy linting batman I got it to pass!

jesserockz (approved):

PR #2029, opened 2022-04-19
Implement a simple LCD menu

numo68:
## Description:

The integration implements a hierarchical menu on character-based LCD displays, primarily meant to be used together with a rotary encoder with a single push button. It is possible to navigate the menu, changing enum values (named integer) and float ones, with the values themselves being maintained in globals (that allows to use them freely in other places, persisting them etc). Automations are added to perform user-defined functionality on entering or leaving the menu level and changing the value or triggering a command. The menu can be hidden and shown again, enabling alternating between the menu and other usage for the display.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3406

## Checklist:


components/display-menu.rst
```diff
@@ -0,0 +1,791 @@
...
+          id(my_lcd_menu).draw();
+          if (!id(my_lcd_menu).is_active())
+            it.print("Menu is not active");
+
+    # Declare a LCD menu
+    lcd_menu:
+      id: my_lcd_menu
+      display_id: my_lcd
+      dimensions: 20x4
+      active: True
```
jesserockz: Please change all `True` and `False` to be `true` and `false`
```suggestion
      active: true
```

jesserockz (changes requested):
This should start out as either `lcd_menu.rst` or split into `lcd_menu.rst` for specific `lcd_menu` stuff and a `display_menu.rst` for the common stuff. (Also underscore not hypen please in the filename)

components/display-menu.rst
numo68: I have split the description to the common and `lcd_menu` specific part; I created a `display_menu` subdirectory as it is expected that a pixel-oriented display arrives here at some day. True/False changed to lowerspace (frankly the existing code has them half/half so it is hard to tell what is expected). Names have underscores. Thanks.

nagyrobi:
Not stale

jesserockz (approved):

PR #2041, opened 2022-04-25
Update thermostat.rst

gvdhoven:
Explained "default_mode" in more detail.

## Description:

## Checklist:

### DIFF
--- a/components/climate/thermostat.rst
+++ b/components/climate/thermostat.rst
@@ -177,9 +177,11 @@ These configuration items determine default values the thermostat controller sho
   - ``fan_only``
   - ``auto``
   
-  This value is used if you flash your device for the first time. This means that if you want
-  your thermostat configuration to start in a certain state, you should configure this. If
-  not configured, you have to change the mode (once) via e.g. the UI of Home Assistant.
+This value is used the first time your device starts after ESPHome is initially installed onto it. Add
+this option into your configuration if you want your thermostat component to start in a climate mode other
+than ``off``.  If this option is not configured, you'll need to manually change the climate mode later via
+the front end (Home Assistant), an ESPHome action, automation, or from within a lambda elsewhere in your
+device's configuration.
 
 - **default_target_temperature_low** (*Optional*, float): The default low target
   temperature for the control algorithm. This can be dynamically set in the frontend later.
--- a/components/climate/thermostat.rst
+++ b/components/climate/thermostat.rst
@@ -52,6 +52,7 @@ modes that Home Assistant offers.
       - platform: thermostat
         name: "Thermostat Climate Controller"
         sensor: my_temperature_sensor
+        default_mode: heat_cool
         default_target_temperature_low: 20 °C
         default_target_temperature_high: 22 °C
         min_cooling_off_time: 300s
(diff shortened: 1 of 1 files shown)


kbx81:
The examples we provide at the top are usually intended to illustrate the minimum configuration to get the component working -- while I don't feel terribly strongly about this, I think we should keep them that way in the interest of streamlining the documentation. If we added a line there every time someone didn't understand one of the options, it would end up looking like what I have in my [config](https://github.com/kbx81/esphome-configs/blob/62ad8ff9d2ad4e69b03b16aabeda80ac4ae8baa2/shared/esp_thermostat_nextion_common_config.yaml#L223-L317); I think it best we avoid that. 🙂 Also, based on what I know about it, I think it's also a safe assumption that most folks will not want the default mode to be a mode where the system is active.

I think the call-out you added is good and adds some clarity; I've got a couple suggestions in terms of wording to make it more consistent with the language/terms we use in the community. I'll leave an in-line comment or two in a moment.

components/climate/thermostat.rst
```diff
@@ -52,6 +52,7 @@ modes that Home Assistant offers.
       - platform: thermostat
         name: "Thermostat Climate Controller"
         sensor: my_temperature_sensor
+        default_mode: heat_cool
```
kbx81: I think we should strike this and keep the examples as simple/minimal as possible.

components/climate/thermostat.rst
```diff
@@ -74,6 +75,7 @@ modes that Home Assistant offers.
       - platform: thermostat
         name: "Thermostat Climate Controller"
         sensor: my_temperature_sensor
+        default_mode: heat
```
kbx81: Same as above

components/climate/thermostat.rst
```diff
@@ -90,6 +92,7 @@ modes that Home Assistant offers.
       - platform: thermostat
         name: "Thermostat Climate Controller"
         sensor: my_temperature_sensor
+        default_mode: cool
```
kbx81: Same as above

components/climate/thermostat.rst
```diff
@@ -173,6 +176,10 @@ These configuration items determine default values the thermostat controller sho
   - ``dry``
   - ``fan_only``
   - ``auto``
+  
+  This value is used if you flash your device for the first time. This means that if you want
+  your thermostat configuration to start in a certain state, you should configure this. If
+  not configured, you have to change the mode (once) via e.g. the UI of Home Assistant.
```
kbx81: Let's re-word this so it's more consistent with the terms/phrasing we use in the community. Here's a suggestion:

```suggestion
This value is used the first time your device starts after ESPHome is initially installed onto it. Add
this option into your configuration if you want your thermostat component to start in a climate mode other
than ``off``.  If this option is not configured, you'll need to manually change the climate mode later via
the front end (Home Assistant), an ESPHome action, automation, or from within a lambda elsewhere in your
device's configuration.
```

kbx81 (changes requested):
Some suggestions, as promised. 😄

gvdhoven:
All comments are resolved.

kbx81 (approved):
LGTM -- thanks!

PR #2073, opened 2022-05-12
Update Inkbird TH1 docs to show support for TH2 sensors

ciband:
## Description:
The current Inkbird TH1 component appears to support TH2 sensors as well.  This should be reflected in the documentation.  I am running this on my own setup with success.

## Checklist:


ciband:
@jesserockz This patch is ready for review.  I didn't want to add too much to the existing docs but can add pictures or something else if you think it is valuable.

jesserockz (approved):

jesserockz:
Thanks @ciband

PR #2134, opened 2022-06-14
Add the possibility to transform strings for LCD displays

numo68:
## Description:

The LCD displays are produced with several ROM variants defining different character sets. On the other hand the native character encoding of the yaml configuration and most of code editors is UTF-8. An `it.printf("Temp %4.1f °C", id(temp).state);` lambda cannot be used as is.

The `transform` lambda allows the user to define a transformation function run immediately before the string is output to the LCD.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3563

## Checklist:

### DIFF
--- a/components/display/lcd_display.rst
+++ b/components/display/lcd_display.rst
@@ -243,15 +243,15 @@ defines a dot at the upper left and lower right of the character.
           it.print("Hello, world \x08 \x07!");
 
 
-Transform Lambda
-----------------
+Transformation Lambda
+---------------------
 
 The LCD displays are produced with several ROM variants defining different character sets.
 On the other hand the native character encoding of the yaml configuration and most of code
-editors is UTF-8. An `it.printf("Temp %4.1f °C", id(temp).state);` lambda cannot be used
+editors is UTF-8. An ``it.printf("Temp %4.1f °C", id(temp).state);`` lambda cannot be used
 as is.
 
-The `transform` lambda allows the user to define a transformation function run immediately
+The ``transform`` lambda allows the user to define a transformation function run immediately
 before the string is output to the LCD. The input string is supplied in the argument `x`
 and the lambda needs to return the result in the LDC's character set. The following snippet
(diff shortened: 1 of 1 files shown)


glmnet:
You should add the `transform` to the formal parameter list. And add a link to the explanation.

Also, did you try that the code (with the sample) fixes the original issue? I mean, since the issue is specifically with the character given in the issue: https://github.com/esphome/feature-requests/issues/1766#issuecomment-1154191530 providing an example for those will be good (I had issues as some of those characters are multi byte so could not be stored in char or something like that and had to use a string), may be you can have @nagyrobi test it for you.

numo68:
> You should add the `transform` to the formal parameter list. And add a link to the explanation.

Done, also added one for `user_characters` that was missed as well.

> Also, did you try that the code (with the sample) fixes the original issue? I mean, since the issue is specifically with the character given in the issue: [esphome/feature-requests#1766 (comment)](https://github.com/esphome/feature-requests/issues/1766#issuecomment-1154191530) providing an example for those will be good (I had issues as some of those characters are multi byte so could not be stored in char or something like that and had to use a string), may be you can have @nagyrobi test it for you.

Here I would like to ask @nagyrobi to test as he has a real project for this. The code I have tested and is provided in the documentation example at https://deploy-preview-2134--esphome.netlify.app/components/display/lcd_display#display-transformation-lambda is however IMO close enough to what was done in that comment. The degree character used is a multi-byte one (`\xc2\xb0`) so that part is tested.

Thanks

numo68:
@glmnet Well.. there is an effect identified by @nagyrobi's testing limiting the usability for the intended purpose somehow and that is that one cannot easily calculate the actual length of the string anymore, meaning any naive attempt to computationally right-align things fails if there is a multi-byte character involved. I added a note of this to the documentation.

I still find this PR usable, YMMV.

PR #2156, opened 2022-06-27
add min_value and max_value filter sensors in docs

### LINKED ISSUE #1581: Add support for max_value/min_value in filters for sensors

Ability to limit the value of a sensor to a min/max value

```
  - lambda: |-
      if (x >= 150.0) { return 150.0;}
      return x
```

**Additional context**

I know it's possible with a lambda but it feels like this should be supported out of the box with ``max_value`` and ``min_value`` in the filters.

euEmica:
## Description:

Added in the documentation the new filters for threshold value sensors for maximum and minimum values.

**Related issue (if applicable):** fixes [#1581](https://github.com/esphome/feature-requests/issues/1581)

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3605

## Checklist:

### DIFF
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -121,6 +121,8 @@ Filters are processed in the order they are defined in your configuration.
     filters:
       - offset: 2.0
       - multiply: 1.2
+      - min_value: 0
+      - max_value: 50
       - calibrate_linear:
           - 0.0 -> 0.0
           - 40.0 -> 45.0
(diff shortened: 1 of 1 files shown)


components/sensor/index.rst
```diff
@@ -170,6 +172,28 @@ Adds a constant value to each sensor value.
 
 Multiplies each value by a constant value.
 
+``min_value``
+*************
+
+Minimum value limit value for the ``min`` value.
```
OttoWinter, line 178: This will need more text to explain what it does exactly (something like if sensor value is less than X, output X, otherwise sensor value). Important because it's hard to distinguish from the `min` filter otherwise.

PR #2163, opened 2022-06-30
wifi: document enable_btm and enable_rrm

stintel:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3600

## Checklist:

### DIFF
--- a/components/wifi.rst
+++ b/components/wifi.rst
@@ -85,6 +85,9 @@ Configuration variables:
   The downside is that this option connects to the first network the ESP sees, even if that network is very far away and
   better ones are available.
 
+- **enable_btm** (*Optional*, bool): Only on ``esp32`` with ``esp-idf``. Enable 802.11v BSS Transition Management support.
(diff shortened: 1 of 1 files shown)


components/wifi.rst
```diff
@@ -85,6 +85,9 @@ Configuration variables:
   The downside is that this option connects to the first network the ESP sees, even if that network is very far away and
   better ones are available.
 
+- **enable_btm** (*Optional*, bool): Only on ``esp32`` with ``esp-idf``. Enable 802.11v BSS Transition Management support.
+- **enable_rrm** (*Optional*, bool): Only on ``esp32`` with ``esp-idf``. Enable 802.11k Radio Resource Management support.
```
OttoWinter, line 89: Quick check, does enabling these have any downsides? If not, can we enable them per default? Would also save adding a new config option.

stintel: Good question ... I've seen too many WiFi compatibility issues to feel good with forcing something onto users, so I prefer having a config option.

jesserockz (approved):

PR #2165, opened 2022-07-01
add example to add support for HA Energy Dashboard

bvansambeek:
This adds an example to add device in the individual devices drop down list in the HA Energy Dashboard.

## Description:

## Checklist:

### DIFF
--- a/components/sensor/total_daily_energy.rst
+++ b/components/sensor/total_daily_energy.rst
@@ -66,7 +66,7 @@ Some sensors such as the :doc:`HLW8012 <hlw8012>` expose their power sensor with
           unit_of_measurement: kW
 
 Add as Individual device to the Home Assistant Energy Dashboard
------------------------
+---------------------------------------------------------------
 Add device_class ``energy`` and add unit of measuremment ``kWh`` to make it appear in the dropdown list of the Individual devices in the Home Assistant Energy Dashboard.
 
 .. code-block:: yaml
--- a/components/sensor/total_daily_energy.rst
+++ b/components/sensor/total_daily_energy.rst
@@ -65,6 +65,34 @@ Some sensors such as the :doc:`HLW8012 <hlw8012>` expose their power sensor with
             - multiply: 0.001
           unit_of_measurement: kW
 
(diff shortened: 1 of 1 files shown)


components/sensor/total_daily_energy.rst
```diff
@@ -65,6 +65,34 @@ Some sensors such as the :doc:`HLW8012 <hlw8012>` expose their power sensor with
...
+
+    # Example configuration entry
+    sensor:
+      - platform: total_daily_energy
+        name: "Total Daily Energy"
+        power_id: my_power
+        filters:
+            - multiply: 0.001
+        unit_of_measurement: kWh
+        device_class: energy
```
frenck, line 82: This is already its default device class, setting it, is, therefore, not needed.

See: <https://github.com/esphome/esphome/blob/fe535939a3b151d1280d2eddf09213f5eb956153/esphome/components/total_daily_energy/sensor.py#L43>

components/sensor/total_daily_energy.rst
```diff
@@ -65,6 +65,34 @@ Some sensors such as the :doc:`HLW8012 <hlw8012>` expose their power sensor with
...
+.. code-block:: yaml
+
+    # Example configuration entry
+    sensor:
+      - platform: total_daily_energy
+        name: "Total Daily Energy"
+        power_id: my_power
+        filters:
+            - multiply: 0.001
+        unit_of_measurement: kWh
```
frenck, line 81: There is no need to add filters or multiple to convert from `Wh` to `kWh`, Home Assistant will accept both and handle the conversion automatically.

frenck (commented):
I wonder how this example is different compared from the original example on the top of the page?

The default example shown in the introduction of this component is already valid and complete for use with Home Assistant's energy dashboard.

bvansambeek:
Those changes came from a different project with the Integration - Riemann sum integral from Home Assistant and I thought it was also required here, but I didn't test it or thought it through... Sorry. Thanks Frenck for pointing this out and keeping the docs clean.

PR #2168, opened 2022-07-02
Add examples to pca9685

gsieben:
The entry of the ID at the hub is missing. Therefore it is very confusing for a beginner.


components/output/pca9685.rst
```diff
@@ -81,6 +82,27 @@ Configuration variables:
   Use this if you have multiple PCA9685s you want to use at the same time.
 - All other options from :ref:`Output <config-output>`.
 
+
+Note
```
jesserockz: I don't like the title here being `Note`...
But cannot think of what it should be

jesserockz (changes requested):

components/output/pca9685.rst
gsieben: Yes. Good question. I tried to stick to the existing structure. Here are a few suggestions:

Usage
Example
Use case

jesserockz:
```suggestion
Examples
```

components/output/pca9685.rst
```diff
@@ -81,6 +82,27 @@ Configuration variables:
   Use this if you have multiple PCA9685s you want to use at the same time.
 - All other options from :ref:`Output <config-output>`.
 
+
+Examples
+----
```
jesserockz:
```suggestion
--------
```

jesserockz (dismissed):

PR #2169, opened 2022-07-02
Suggested Note

gsieben:
I would be happy if I had seen this entry. It makes it easier for the beginner to understand. I wasted a lot of time trying to understand why I can't get a slider. Now it is clear to me. But many others were looking for the same thing.


gsieben:
The preview somehow does not work for me. Maybe I didn't recognize the syntax correctly.

PR #2180, opened 2022-07-12
lambda example in the script section

gsieben:
## Description:
I have included examples of 'lambda' in the "script" section. I played around for ages until I noticed that the round clamps at the end were missing. In the forum also searched someone who did not get an answer to his question. It will help others.

**Related issue (if applicable):** fixes none

## Checklist:


gsieben:
I have no idea why the check fails. The alleged missing line is there. Something I do not see.

PR #2202, opened 2022-07-29
Include Note about TMC2208 Modules

vBJohannes:
## Description:
As TMC2208 modules are mostly compatible with A4988 modules, there is a way to use this library to also drive an TMC2208. Only slight changes according to the different pinout need to be accounted for.

**Related issue (if applicable):** There is no particular issue I had in mind.

## Checklist:


nagyrobi (commented):
I think you should be more specific rather than just saying _check online_...

vBJohannes:
I don't get the error in the Lint build, tbh.

nagyrobi (dismissed):

vBJohannes:
Apparently the lint check is happy once I indented the table. Merge, please :)

ssieb:
I don't understand how this would be configured with the two pins.  Can you expand the example to be a full config?

vBJohannes:
> I don't understand how this would be configured with the two pins. Can you expand the example to be a full config?

How about now?

nagyrobi:
This should be resubmitted as a PR against `current` branch, please.

PR #2208, opened 2022-08-03
Update index.rst

JeeCee1:
Added required 'then:' to 'switch.is_on / switch.is_off Condition'. See also https://esphome.io/guides/automations.html#if-action

## Description:

## Checklist:

### DIFF
--- a/components/switch/index.rst
+++ b/components/switch/index.rst
@@ -98,10 +98,11 @@ This :ref:`Condition <config-condition>` checks if the given switch is ON (or OF
 
     # In some trigger:
     on_...:
-      if:
-        condition:
-          # Same syntax for is_off
-          switch.is_on: my_switch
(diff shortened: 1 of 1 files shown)


jesserockz (commented):
`then` is not actually **required** for most triggers, but it is accepted by all triggers.

JeeCee1:
Sorry, I added it at the wrong position, meant to do this:

```
on_...:
  if:
    condition:
      # Same syntax for is_off
      switch.is_on: my_switch
    then:
      ...
```
This matches with the `switch.turn_off Action` example.

Edit: I pushed a new commit. I'm quite new to GH, hope this is the right way.

jesserockz:
@JeeCee1 Ah I see. BTW you didnt actually push the commit, it must be sitting local only.

PR #2210, opened 2022-08-06
Document existing feature for "on_value_range"

### LINKED ISSUE #1130: Allow multiple ranges for the sensor.on_value_range trigger

Allow multiple ranges for the on_range_value automation trigger.

As it stands, on_range_value can only be included once per sensor. This is because the sensor value is only compared to one range of values. If the value crosses into the range of values, the trigger fires. But since some sensor ranges call for different actions along its range of possible values, this diminishes the on_range_value trigger's usefulness.

For instance, the humidity in a bathroom. If it's 0-33%, turn fan off, if it's 33-67% turn fan to low, if it's 67-100% turn fan to high.

It's currently probably easiest to accomplish this using templates, but templates are not user friendly.

Something like this is far more parseable and easier for the user to maintain:

```
on_value_range:
  below: 33%
  then:
    - homeassistant.service:
        service: fan.turn_off
        data:
          entity_id: fan.bathroom
  above: 33%
  below: 67%
  then:
    - homeassistant.service:
        service: fan.set_speed
        data:
          entity_id: fan.bathroom
          speed: low
  above: 67%
  then:
    - homeassistant.service:
        service: fan.set_speed
        data:
          entity_id: fan.bathroom
          speed: high
```

**Additional context**

CarlosGS:
## Description:
Updates the `on_value_range` example to reflect already existing feature (multiple ranges).

**Related issue (if applicable):** closes https://github.com/esphome/feature-requests/issues/1130

### DIFF
--- a/components/sensor/index.rst
+++ b/components/sensor/index.rst
@@ -558,10 +558,16 @@ So for example ``above: 5`` with no below would mean the range from 5 to positiv
       - platform: dallas
         # ...
         on_value_range:
-          above: 5
-          below: 10
-          then:
-            - switch.turn_on: relay_1
+          - below: 5.0
+            then:
+              - switch.turn_on: relay_1
+          - above: 5.0
+            below: 10.0
+            then:
+              - switch.turn_on: relay_2
+          - above: 10.0
+            then:
+              - switch.turn_on: relay_3
 
 Configuration variables:
 


nagyrobi (approved):
Useful, thanx

PR #2234, opened 2022-08-19
bme680_bsec: Add information and example about multiple devices support

bisbastuner:
## Description:
PR 3550 (https://github.com/esphome/esphome/pull/3550) adds support for multiple devices, so explicitly state that, and add an example configuration that uses multiple devices

## Checklist:


trvrnrth:
I've done some very minor nit-picking in https://github.com/bisbastuner/esphome-docs/pull/1 but this looks good to me 👍

trvrnrth (dismissed):

trvrnrth (approved):

ssieb:
This needs to target `next`, not `current`.
I don't see the point of including that huge config.  It's not any different than the other components that can have multiple devices.  What needs to be documented is the `bme680_bsec_id:` parameter.

bisbastuner:
> This needs to target `next`, not `current`. I don't see the point of including that huge config. It's not any different than the other components that can have multiple devices. What needs to be documented is the `bme680_bsec_id:` parameter.

Hmm.. I think this means that it would be better to document that by extending the example configuration above, what do you think?

And about the target branch, should I open a new PR and close this or should I rebase this PR?

ssieb:
> I think this means that it would be better to document that by extending the example configuration above, what do you think?

I don't think you need to change any examples, but you could if you want.  Just document the `bme680_bsec_id:` parameter in both sections.  See the Dallas component for example.

ssieb:
> And about the target branch, should I open a new PR and close this or should I rebase this PR?

Rebase it and then use the edit button at the top of this page to change the target branch.

bisbastuner:
@ssieb I think I did it. Please tell me if I did something wrong, these are my first contributions!

components/sensor/bme680_bsec.rst
```diff
@@ -163,6 +169,13 @@ For each sensor all other options from :ref:`Sensor <config-sensor>` and :ref:`T
 .. code-block:: yaml
 
     bme680_bsec:
+        # id
+        # -----------
+        # Identifier for this component, useful when working with multiple devices.
+        # Muse be unique, and can be used in the sensor sections to refer to the correct device.
```
trvrnrth:
```suggestion
        # Must be unique, and can be used in the sensor sections to refer to the correct device.
```

trvrnrth:
This isn't stale. Just needs final maintainer review.

jesserockz (approved):

PR #2238, opened 2022-08-21
Add Prometheus metrics relabeling

jangrewe:
## Description:

Instructions how to use Prometheus metric relabeling.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3734

## Checklist:


glmnet (dismissed):
I'm trying to test this and I'm getting a compilation issue, not sure if it's my setup as I'm not doing much core ESPHome lately. In any case add this feature to any of the 'testNN.yaml' files so the CI builds it, if it goes well I'll check again.

glmnet (approved):

PR #2263, opened 2022-09-04
Delayed Startup Code

evilsidekick293:
Adding Note regarding the need of a delayed startup of the communication
between ESP and AM2320 sensor due to the sensor being slow on a cold startup.

## Description:

**Related issue (if applicable):** this issue was supposed to be fixed with https://github.com/esphome/esphome/pull/554 but appeared at my setup and this workaround seems to fix it

## Checklist:

### DIFF
--- a/components/sensor/am2320.rst
+++ b/components/sensor/am2320.rst
@@ -24,6 +24,13 @@ The ``am2320`` Temperature+Humidity sensor allows you to use your AM2320
     Logs might include some warnings about receiving a NACK from the sensor.
     This is due to a wake call to the sensor which the sensor never acknowledges by design.
 
+.. note::
+
+    If there seems to be an issue with the ESP setting up a connection, the
+    sensor might not be fully started up. This can occur on a cold start of
+    your node. Check if a delay helps you by adding ``setup_priority: -1000``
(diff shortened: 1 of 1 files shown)


sidekick293:
My full code looks like this:

.. code-block:: yaml
    esphome:
      name: trh-sensor-4
      platform: ESP8266
      board: esp01_1m
    
    # Enable logging
    logger:
    
    # Enable Home Assistant API
    api:
    
    ota:
      password: "<some pw>"
    
    wifi:
      ssid: "<wifi ssid>"
      password: "<wifi pw>"
    
      # Enable fallback hotspot (captive portal) in case wifi connection fails
      ap:
        ssid: "Trh-Sensor-4 Fallback Hotspot"
        password: "<some other pw>"
    
    captive_portal:
    
    i2c:
      sda: GPIO4
      scl: GPIO5
      scan: True
      frequency: 200000
    
    sensor:
      - platform: am2320
        setup_priority: -1000
        temperature:
          name: "T"
        humidity:
          name: "RH"
        update_interval: 60s

Without the ``setup_priority: -1000`` the sensor is not responding quick enough and the whole node cannot start properly.

PR #2288, opened 2022-09-12
Add magnetometer calibration reference

nliaudat:
## Description:
Add magnetometer calibration reference with esphome code

**Related issue (if applicable): nothing

## Checklist:

  - [ X] Branch: `next` is for changes and new documentation that will go public with the next ESPHome release. Fixes, changes and adjustments for the current release should be created against `current`.
  - [X ] Link added in `/index.rst` when creating new documents for new components or cookbook. - Not applicable


jesserockz:
Hello, 
Thanks for contributing.

I feel like a small link in the see also section is not really helpful.

What would be better is an actual section of text with a header and explaining how this is helpful and link to the repository there.

Same goes for #2289 =)

Thanks,
Jesse

nliaudat:
Hi, 
I mean magnetometer calibration is intended only for advanced users and the small link in "See also" is enough. 
Regards

PR #2291, opened 2022-09-13
"name" not required

scaarup:
Does not seem like "name" is required and it is not mentioned in example 2.

## Description:
The documentation for a home assistant text sensor, mentions that "name" is a required configuration parameter. On the same page, there is an example given without "name". Also when I test a configuration without "name", it still works fine. So I remove "required" from "name" on this doc.

## Checklist:


glmnet:
You should mark it optional, like attribute is.

Now that we are here, shouldn't this page indicate that someone should give this sensor an id so it can be used in other automations / conditions? 

name is only useful if you want to forward this thing to say web server or mqtt I guess... never tried though.

and of course the same applies to other Home Asisstant **  things so this PR should keep consistency in case a change is applied

scaarup:
> You should mark it optional, like attribute is.
> 
> Now that we are here, shouldn't this page indicate that someone should give this sensor an id so it can be used in other automations / conditions?
> 
> name is only useful if you want to forward this thing to say web server or mqtt I guess... never tried though.
> 
> and of course the same applies to other Home Asisstant ** things so this PR should keep consistency in case a change is applied

Sorry forgot about that. Name is now marked as optional.

nielsnl68:
From what i learned the name is being used for identify the component in Home Assistant. @scaarup, Are you saying that without setting the name property you still use the component?

scaarup:
> From what i learned the name is being used for identify the component in Home Assistant. @scaarup, Are you saying that without setting the name property you still use the component?

Yeah it works fine without a "name". And I am only using the text_sensor's internally on the esp, so referring to them by id.

PR #2292, opened 2022-09-13
Remove deprecated settings and document new default_preset

MrMDavidson:
## Description:

Removes deprecated settings from the documetnation and documents new default_preset related settings

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3643

## Checklist:

  - ~~[ ] Link added in `/index.rst` when creating new documents for new components or cookbook~~

### DIFF
--- a/components/climate/thermostat.rst
+++ b/components/climate/thermostat.rst
@@ -392,7 +392,7 @@ These configuration items determine default values the thermostat controller sho
     - ``default_preset``: The thermostat will always switch to the presets specified by **default_preset**
 
 .. note::
-  
+
   You can specify a ``default_preset`` and set ``on_boot_restore_from`` to ``memory``. In this mode when
   the settings from last boot cannot be retrieved, for any reason, then the specified ``default_preset``
   will be applied.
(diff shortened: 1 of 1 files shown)


components/climate/thermostat.rst
```diff
@@ -422,6 +378,67 @@ experience and automation.
...
+
+These configuration items determine default values the thermostat controller should use when it starts.
+
+- **default_preset** (*Optional*, string): The name of the preset to use by default. Must match a preset
+  as per :ref:`preset <thermostat-preset>`.
+- **on_boot_restore_from**: (*Optional*, on_boot_restore_from): Controls what the thermostat will do when
+  it first boots. One of:
+
+    - ``memory`` (default): The thermostat will restore any settings from last time it was running.
+    - ``default_preset``: The thermostat will always switch to the presets specified by **default_preset**
```
kbx81: Total nitpick here, but this should probably say "preset" (singular) and not "presets". Could be confusing. 😇 You'd never switch to more than one preset (unless there is something going on that I don't know about...lol).

kbx81 (changes requested):
One tiny, tiny nitpick, mostly just to avoid confusion. Otherwise, consider this approved!

components/climate/thermostat.rst
MrMDavidson: Good catch, fixed!

kbx81 (approved):

PR #2304, opened 2022-09-17
Document steps for adding a password to OTA

ccdunder:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** none

## Checklist:

### DIFF
--- a/components/ota.rst
+++ b/components/ota.rst
@@ -197,8 +197,8 @@ through an ``on_boot`` trigger:
 Adding a password:
 ******************
 
-If OTA is already enabled without a password, simply add a `password:` line to the existing
-`ota:` config block.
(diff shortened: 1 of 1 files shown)


components/ota.rst
```diff
@@ -191,6 +194,12 @@ through an ``on_boot`` trigger:
       password: "Old password"
       id: my_ota
 
+Adding a password:
+******************
+
+If OTA is already enabled without a password, simply add a `password:` line to the existing
+`ota:` config block.
```
ssieb:
```suggestion
If OTA is already enabled without a password, simply add a ``password:`` line to the existing
``ota:`` config block.
```

PR #2306, opened 2022-09-18
Update Mirabella-genio-bulb.rst

Drew765:
Adding note (at line 255) that custom component for Mirabella CCW downlights is no longer needed. It was merged into ESPhome. Link explains solution.

## Description:

## Checklist:


jesserockz (changes requested):
Instead of adding a note, you should change the configuration to match what should be used.

PR #2320, opened 2022-09-23
Max6956 support added

looping40:
## Description:
Documentation for Max6956

Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable): https://github.com/esphome/esphome/pull/3570

## Checklist:


jesserockz (changes requested):
The multiple large blocks of config examples needs to be changed.
There should be a `Component/Hub` heading which has an example with minimum required config for the main component (no fields if they have defaults) then list all config variables for the main component.

Then a new heading for float output with a single small example for one output and lis thte variables.

Then a new heading for the gpio pin configuration.

Let me know if you ahve any questions and ill check back again

jesserockz (approved):

PR #2347, opened 2022-10-05
Pcf85063 rtc

brogon:
## Description:

Adds documentation about NXP PCF85063 time source

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3873

## Checklist:


jesserockz (changes requested):
I personally think this actually makes it harder to read. It would be better to keep the two components in their own headers, even if the content is mostly the same.

PR #2363, opened 2022-10-13
Updates for xl_maxsonar_wr

### LINKED ISSUE #1913: add support for XL-MaxSonar-WR/WRC MaxBotix sensors
Hello Community =) 

ESPHome already supports the high resolution (hr) versions of the maxbotix distance sensors, but it would be nice to also support other models. Specifically, the XL-MaxSonar-WR series with model numbers MB70XX and MB71XX as given in this datasheet: https://www.maxbotix.com/documents/XL-MaxSonar-WR_Datasheet.pdf

Got my hands on a MB7052 and using a MAX232 converter to convert RS232 to TTL. It all works, the esp8266 receives bytes without a problem, but the provided sensor platform for hrxl_maxsonar_wr will only log "Invalid data read from sensor: %s"

I found a workaround using a text sensor to read the bytes and convert them to int in HA downstream, but this solution lacks nice features like filters and throttling (which are very much desired at a publishing frequency of 6.6 Hz). 

**Additional context**

Looking at the source code, I found that the control flow requires a buffer length of 6 for values to get published. However, the lower precision versions will only send bytes with a total length of 5, e.g. 'R012\r' and not 'R0123\r'. As a result, the values are never published because the if condition never evaluates to true. 

I tried to tweak the hrxl_maxbotix_wr.cpp file, but couldn't find it. Simply replacing the value in line 17 would suffice, I guess. 
 `static const int MAX_DATA_LENGTH_BYTES = 5; // priviously 6`
Since this would also help my situation, I'd be grateful for advice on how to locate esphome source files in my HA installation.

Worschi:
Give some more context on different models and hint at new platform name to use

## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1913

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#3903

## Checklist:

### DIFF
--- a/components/sensor/hrxl_maxsonar_wr.rst
+++ b/components/sensor/hrxl_maxsonar_wr.rst
@@ -2,14 +2,15 @@ HRXL MaxSonar WR Series
 =======================
 
 .. seo::
-    :description: Instructions for setting up MaxBotix HRXL MaxSonar WR ultrasonic distance measurement sensors in ESPHome.
+    :description: Instructions for setting up MaxBotix (HR)XL MaxSonar WR ultrasonic distance measurement sensors in ESPHome.
     :image: hrxl_maxsonar_wr.jpg
     :keywords: ultrasonic, maxbotix, maxsonar
 
-This sensor allows you to use HRXL MaxSonar WR series ultrasonic sensors by MaxBotix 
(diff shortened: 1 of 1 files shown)


Worschi:
Closing this, since related pull request https://github.com/esphome/esphome/pull/3903 has been rejected

PR #2371, opened 2022-10-19
Add example workaround for static uniqueid

rlowens:
Since the sensor's uniqueid is static you have to use the copy sensor instead. Here's an example of doing that.

## Description:

## Checklist:


rlowens:
Added sensor-copy anchor

nagyrobi:
retest

PR #2377, opened 2022-10-21
Revert "Fix color temp units for strobe effect"

nagyrobi:
Reverts esphome/esphome-docs#2375


nagyrobi:
Reverted because I merged it by mistake. Sorry.

PR #2402, opened 2022-10-30
Added note about decimal accuracy of other sensors

dirrgang:
## Description:
Any sensors other than the DHT11 has at least one digit of accuracy. The default is 0, so this accuracy is lost and it's not obvious that the default masks the possibly higher resolution.

## Checklist:


nagyrobi:
Thanks

PR #2406, opened 2022-11-02
SDM230M important notice to user to pay attention when configuring the energy monitor

optimusprimespace:
## Description:
The SDM230M is a useful power meter, but it doesn't come with the 9600bps default stated in the previous documentation.

I changed the doc to include the important notice, so if someone makes the same mistake to not change the default baud rate the energy monitor came with to the value that is defined in the example yaml file it won't work.

## Checklist:


nagyrobi:
Is this related to a new component?
If not, then the branch you need to send your PR against is `current`.

optimusprimespace:
Did i change it successfully? I'm new to github.

nagyrobi:
Much better, thanks.
Why changing SDM220M everywhere to SDM230M?

optimusprimespace:
The sdm230m is a newer model, but i could change it back to sdm220m if everyone is happy with that. :)

nagyrobi:
Did you test the new model yourself?

optimusprimespace:
I've been running it for months without problems.

PR #2416, opened 2022-11-08
Update sonoff.rst

lakshanthad:
Not all SONOFF devices are supported by ESPHome. Only the ones with ESP based chips inside are supported


nagyrobi:
Thank you

PR #2435, opened 2022-11-13
Update description on temperature offset

Duckle29:
This clarifies the "polarity" of the temperature offset setting


Duckle29:
I am also curious, does this setting have any effect on the reported relative humidity? If the temperature is too high, it'll also report a too low relative humidity.

I've done some RH compensation myself before, but I've never actually verified if that works:
https://github.com/Duckle29/wifir_firmware/blob/main/lib/sensor_wrapper/sensor_wrapper.cpp#L345-L360
I could try and verify that now tho, as I have some Aqara sensors that seem pretty solid that I could compare to

Duckle29:
Questions answered. BSEC does correct humidity reading too :) 
image

nagyrobi:
Thanks

PR #2468, opened 2022-11-22
Waveshare full update every

Nizzle:
## Description:
The waveshare documentation lists some models that support the full_update_every variable, but this list is not complete. This change does not list them explicitly anymore.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** https://github.com/esphome/esphome/pull/4066

## Checklist:


Nizzle:
I am not quite sure if I should use current or next branch. Please let me know if current is not the right one.

PR #2469, opened 2022-11-22
Update bluetooth proxy docs with latest information on active bluetooth connections

bdraco:
## Description:

- Add warning about web server component causing failures with active bluetooth connections
- Add notes about using esp-idf
- Add notes about flashing via serial to change partitions
- Add notes about recommended versions
- Add sample minimal configuration

## Checklist:


nagyrobi:
This is not exclusive to webserver vs. active connections of proxy. Can be anything vs proxy, for example I tried 20 timed covers and the proxy and got the same result as you.

It'a a generic thing. Many other component combinations can lead to this.

Bluetooth Proxy and Webserver are both memory intensive so yeah, when used together it's more likely that users can run into this problem.

bdraco:
It should be a lot better in 2022.12.x with the new idf.   The web server is just the biggest offender I keep getting in sample config files when people complain about proxies not working

nagyrobi:
I'd put a more generic warning at both proxy and web server, that the component is memory intensive and using together with other (memory intensive) components or configurations can quickly lead to unexpected behavior.

bdraco:
> I'd put a more generic warning at both proxy and web server, that the component is memory intensive and using together with other (memory intensive) components or configurations can quickly lead to unexpected behavior.

There is already a warning on the web server page.  I really want something to point users to since they keep questioning why the web server needs to be disabled when troubleshooting

nagyrobi:
> There is already a warning on the web server page.

That warning is weak, it needs to be emphasized (eg. as a separate note or warning paragraph). That should be scary enough for the people to take it seriously.

components/bluetooth_proxy.rst
```diff
@@ -22,6 +22,9 @@ our `Bluetooth Proxy installer <https://esphome.github.io/bluetooth-proxies/>`__
     Passively broadcasted sensor data (that is advertised by certain devices without active connections) is received
     separately from these, and is not limited to a specific number.
     
+    When using active connections, the :doc:`web_server` component should be disabled as the device is likely
+    to run of memory and malfunction when both components are enabled simultaneously.
```
balloob:
```suggestion
    to run out of memory and will malfunction when both components are enabled simultaneously.
```

balloob:
I wonder if the Bluetooth proxy docs should have a minimal sample config and tell people that if they deviate and run into issues they should disable their extra things one by one until issue resolves, starting with webserver

nagyrobi:
That should be more apropriate, yes!

components/bluetooth_proxy.rst
```diff
@@ -22,6 +22,9 @@ our `Bluetooth Proxy installer <https://esphome.github.io/bluetooth-proxies/>`__
...
+
```
jesserockz: I would even go as far as to make this whole thing its own warning box

```suggestion

  .. warning::

    
    When using active connections, the :doc:`web_server` component should not be configured as the device is likely
    to run out of memory and malfunction when both components are enabled simultaneously.
    This can also be the case when trying to add ``bluetooth_proxy`` to a device that already has a number of other 
    components configured. It is best to make the device as dedicated to ``bluetooth_proxy`` as you can to 
    have the best results.
    
```

bdraco:
I'll come back to this once I'm finished with all the core side

bdraco:
I will rework this once 2022.12.x is released with a sample config

2022.12.x should allow us to switch to esp-idf which will save 50k of RAM which makes everything much more stable

related https://github.com/esphome/bluetooth-proxies/pull/52

components/bluetooth_proxy.rst
bdraco: Adjusted

bdraco:
> I wonder if the Bluetooth proxy docs should have a minimal sample config and tell people that if they deviate and run into issues they should disable their extra things one by one until issue resolves, starting with webserver

I added a sample config with guidance: `If you experience issues with your proxy, try reducing your configuration to be as similar to this as possible.`

components/bluetooth_proxy.rst
```diff
@@ -56,6 +67,58 @@ To maximize the chances of catching advertisements of the sensors, you can set `
...
+Complete sample recommended configuration
+-----------------------------------------
+
+Below is a complete sample recommended configuration for a Bluetooth proxy.
+
+This configuration is for a Olimex ESP32-PoE-ISO board with an Ethernet connection to the network.
+
+If you are using a different board, you will need to change the ``board`` substitution to match your board.
+
+If you experience issues your proxy, try reducing your configuration to be as similar to this as possible.
```
bdraco:
```suggestion
If you experience issues with your proxy, try reducing your configuration to be as similar to this as possible.
```

components/bluetooth_proxy.rst
```diff
@@ -56,6 +67,58 @@ To maximize the chances of catching advertisements of the sensors, you can set `
...
 Avoid placing the ESP node in racks, close to routers/switches or other network equipment as EMI interference will degrade Bluetooth signal reception. For best results put as far away as possible, at least 3 meters distance from any other such equipment. Place your ESPHome devices close to the Bluetooth devices that you want to interact with for the best experience.
 
+Complete sample recommended configuration
+-----------------------------------------
+
+Below is a complete sample recommended configuration for a Bluetooth proxy.
+
+This configuration is for a Olimex ESP32-PoE-ISO board with an Ethernet connection to the network.
+
+If you are using a different board, you will need to change the ``board`` substitution to match your board.
```
bdraco:
```suggestion
If you use a different board, you must change the ``board`` substitution to match your board.
```

components/bluetooth_proxy.rst
```diff
@@ -56,6 +67,58 @@ To maximize the chances of catching advertisements of the sensors, you can set `
 
 Avoid placing the ESP node in racks, close to routers/switches or other network equipment as EMI interference will degrade Bluetooth signal reception. For best results put as far away as possible, at least 3 meters distance from any other such equipment. Place your ESPHome devices close to the Bluetooth devices that you want to interact with for the best experience.
 
+Complete sample recommended configuration
+-----------------------------------------
+
+Below is a complete sample recommended configuration for a Bluetooth proxy.
+
+This configuration is for a Olimex ESP32-PoE-ISO board with an Ethernet connection to the network.
```
bdraco:
```suggestion
This configuration is for an Olimex ESP32-PoE-ISO board with an Ethernet connection to the network.
```

components/bluetooth_proxy.rst
```diff
@@ -40,6 +50,7 @@ The Bluetooth proxy depends on :doc:`esp32_ble_tracker` so make sure to add that
 .. note::
 
     Bluetooth proxy requires Home Assistant 2022.9 or later.
+
```
bdraco:
```suggestion
```

components/bluetooth_proxy.rst
```diff
@@ -14,13 +14,21 @@ and Home Assistant.
...
 
     The Bluetooth proxy of ESPHome provides Home Assistant with a maximum number of 3 simultaneous active connections.
     Devices which maintain a *continuous active* connection will consume one of these constantly, whilst devices which
     do *periodic disconnections and reconnections* will permit using more than 3 of them (on a statistical basis).
     Passively broadcasted sensor data (that is advertised by certain devices without active connections) is received
     separately from these, and is not limited to a specific number.
+
+    The :doc:`esp32` component should be configured to use the ``esp-idf`` framework, as the ``arduino`` framework
+    uses significantly more memory and performs poorly with the Bluetooth proxy enabled. When switching from
+    ``arduino`` to ``esp-idf``, make sure to update the device with a serial cable as the partition table is
```
nagyrobi, line 27: I'd be a bit more permissive an exact:

When switching from  ``arduino`` to ``esp-idf``, and using active connections, it is recommended to to update the device with a serial cable

because this is not something mandatory when somebody is using the device passively (which is default setting).

nagyrobi:
To be more understandable: about this cache_services functionality:
- _services_ mean _Bluetooth services_?
- only affects use cases with _active: true_?

components/bluetooth_proxy.rst
bdraco: I think this ok as-is given its in the `Active connections` block

bdraco:
> To be more understandable: about this cache_services functionality:
> 
> * _services_ mean _Bluetooth services_?

I clarified they are GATT services

> * only affects use cases with _active: true_?

I think `which significantly speeds up active connections` covers this

bdraco:
Thanks

PR #2477, opened 2022-11-25
Disable logging to serial

darth-hp:
## Description:
Disable logging to serial in example config ... Took me ages to figure this one out...


nagyrobi (changes requested):
Instead of changing the default config example, I'd suggest the standard warning:

```
.. warning::

    If you are using the :doc:`logger` make sure you are not using the same pins for it or otherwise disable the UART 
    logging with the ``baud_rate: 0`` option.
```

darth-hp:
Thing is, the example uses the default UART, thus this is required.

PR #2488, opened 2022-11-28
Note about S26

pedjas:
## Description:

S26 has different board layout. I entered short notice for people who are looking what to buy to know. I have board photos too but was unable to upload them (uploading is disabled). I will try to flash ESPHome firmware to this model and I will post update about it.

sonoff_s20_pcb_back_01
sonoff_s20_pcb_back_02
sonoff_s20_pcb_esp8285
sonoff_s20_pcb_front_01
sonoff_s20_pcb_front_02

### DIFF
--- a/devices/sonoff_s20.rst
+++ b/devices/sonoff_s20.rst
@@ -320,6 +320,12 @@ plug it into the socket.
 
 Happy hacking!
 
+Additional notes
+--------------------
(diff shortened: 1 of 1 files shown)


pedjas:
It seems there is not much point contributing when contributions are just ignored and not included.

PR #2525, opened 2022-12-13
Note ID is not used by Home Assistant

gb0101010101:
## Description:

Add note to ID variables to make it clear that it is not used by Home Assistant.

## Checklist:
  - [x ] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.


nagyrobi:
Don't know why this is needed. There was nowhere any kind of suggestion that would make anyone ever believe that this has anything to do with HA's `entity_id`. HA generates `entity_id` mostly on its own.

gb0101010101:
> Don't know why this is needed.  There was nowhere any kind of suggestion....

The problem is that its difficult to find info in HA or ESPHome docs about this. I still cannot find a HA doc page that says `Entity ID` is generated from `Name`. It's mentioned in the forums but you have to know what to search for. It was certainly a bad assumption by me that it would work this way. 

I know I am not alone in this as there are two feature requests where others thought the same and wanted it to work that way in the future.

https://github.com/home-assistant/core/issues/35683
https://github.com/home-assistant/core/issues/47546

I was planning on adding more ESPHome doc info on using good ESPHome `Name` so that it translates into meaningful HA `Entity ID`  e.g. Use device name prefix in sensor name, only use alpha numeric, no symbols. 

This would stop issues like:

ESPHome Button Name: "TV Volume+" -> HA Entity ID: "button.tv_volume"
ESPHome Button Name: "TV Volume-" -> HA Entity ID: "button.tv_volume_2"

If you have multiple ESPHome devices that do the same:

ESPHome Name: familyroom, Button Name: "TV Volume+" -> HA Entity ID: "button.tv_volume_3"
ESPHome Name: bedroompink, Button Name: "TV Volume+" -> HA Entity ID: "button.tv_volume_4"

Recommend to use:
ESPHome Button Name: "Family Room TV Volume Plus" -> HA Entity ID: "button.family_room_tv_volume_plus"

Let me know if you still don't see value in this.

fxstein:
First of all, thank you @gb0101010101! 

I spent the past several days debugging a pretty complex ESPHome library because, like you, I just assumed that the ids become entity ids in HA.

The naming situation is less than perfect for larger, more complex setups. The amount of manual overrides needed quickly becomes a runaway maintenance issue. In reality, every UI needs short names wherever possible but distinct unique ids to manage all its relationships.

PR #2533, opened 2022-12-14
addded ethernet pinouts for SLZB-06

sergge1:
## Description:
Added ethernet pinouts for SMLIGHT SLZB-06

## Checklist:

### DIFF
--- a/components/ethernet.rst
+++ b/components/ethernet.rst
@@ -76,6 +76,18 @@ Configuration variables:
 Configuration examples
 ----------------------
 
+**SMLIGHT SLZB-06 POE ESP32 Zigbee USB adapter**
+
+.. code-block:: yaml
+
+    ethernet:
+      type: LAN8720
+      mdc_pin: GPIO23
+      mdio_pin: GPIO18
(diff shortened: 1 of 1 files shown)


nagyrobi:
Already in the doc ;)

sergge1:
> Already in the doc ;)

Just found it.  Did not see it before PR.
All the devices are one by one, and only SLZB-06 is grouped with some others. 
Hard to find. 
So I will make a new PR as we are listing all the devices but not the Pinouts group (otherwise, we have to group some other devices with the same pinouts in order to be in the same logic)

nagyrobi:
The exact same pinout configuration don't have to be listed twice, it makes the page long and difficult to see.

Currently these two devices share the exact same config, as far as I see, all the others differ at least in one parameter. But as soon as there will appear more devices matching an existing config, it's sure that we won't start duplicating the same code example for each.

This is not a database, nor an advertisment site. Users who search for the information, can find it, that minimum effort is something we expect the users arriving here to be able to do by themselves.

nagyrobi:
Done brusing.

PR #2542, opened 2022-12-15
Add pixel_mapper option for Addressable Light

nejc-cc:
## Description:
Someone else may find it helpful that it's possible to connect 2x WS2812B flexible LED matrix modules in series and place them one above/below the other.

This is possible by setting pixel_mapper value to:
```c++
      int iMatrixOffset = y >= 8 ? 256 : 0;
      if (x % 2 == 0) {
          return (x * 8) + (y % 8) + iMatrixOffset;
        }
        return (x * 8) + iMatrixOffset + (7 - (y % 8));
```
This makes the whole display become a 16x32 pixel display.

## Checklist:

    or


jesserockz:
This is not a real PR

PR #2571, opened 2023-01-01
feat: Add manual control config variable doc

X-Ryl669:
## Description:
This allows to use a time based cover with a manual control switch. The cover state is therefore only a guess so when calling any command (open or close), the action is performed completely nevertheless. At the end of the action, the assumed state is re-synchronized with the cover position.

**Related issue (if applicable):** Fix https://github.com/esphome/feature-requests/issues/2045

https://github.com/esphome/esphome/pull/4249

## Checklist:

    or


nagyrobi:
How is this related to https://github.com/esphome/esphome-docs/pull/2045 ?

X-Ryl669:
Not this one. [This one](https://github.com/esphome/feature-requests/issues/2045).

nagyrobi:
Link is wrong in the opening post.

X-Ryl669:
It's not a link. It's Github "auto" feature that create a link when you type a hash followed by a number. Since esphome is split in multiple repositories (features requests / docs / issues) for dependent features, this "auto" stuff that works everywhere else on github doesn't work here. The solution is to merge all repositories in a single one if you intend to fix it with the auto stuff or change the "issue" template here so it doesn't ask for the issue number prefixed with a hash.

nagyrobi:
You can just paste the link instead of the number, and it will be correct instantly ;)

X-Ryl669:
Done.

glmnet:
I'll rewrite this as:


For covers with manual external controls. With this configuration if the cover is commanded to open 
or close the corresponding actions will be performed even if assumed state **fully open** or 
**fully closed** matches desired state, then `stop_action` will be called after the full duration of the
action elapses.  Defaults to `false`.

X-Ryl669:
I'm not fond of using the "assumed state" in the description since there's a  "assumed_state" already in the documentation and it's causing confusion. I'm not sure I've understood what the "assumed_state" option actually refers to, it's not used in the time_based component at all, but more likely in the cover component (shouldn't it be there instead?). I've played with it and it didn't do anything on my setup.

So I'd like to find a way not to link/refer with the other parameters name to avoid messing up user's mind. 

It seems the component maintains an internal state (let's call it "computed state", or "expected state") which is not related to the position (there are multiple possible state for the same position, open vs closing...)
Here, the PR doesn't change the state, but deal with the position. When an action is triggered, it switches the position so that the shortcut check for "expected position == current state" doesn't trigger and the action is completed instead of been skipped.

This is, IMHO too complex to write and explain in the documentation, so as you did, maybe it's better to describe the effects, what about:

For covers with manual external controls. With this configuration if the cover is commanded to open
or close the corresponding actions will be triggered in all cases including any declared stop_action. Defaults to false.

or "whatever the cover position" instead of "in all cases", not sure which is clearer.

nagyrobi:
It's all very confusing now...

glmnet:
@nagyrobi sorry yes

@X-Ryl669 it should be made clear that in this case the whole time will have to elapse again before calling stop. 

If it is fully closed by ESPHome then 10% opened by hand, then commanded to close, ESPHome will close it by 100%. But if it was 10% opened by ESPHome then it will later close by 10% only, that's timed based cover expected. 

I guess it is ok to give a detailed explanation of how this works. 

I'll check the assumed state implementation, it should allow you the up down buttons in HA ui even when state is "assumed" fully open or closed. If configured false up or down button can be grayed 
See here when configured false:

image

nagyrobi:
> when state is "assumed" fully open or closed.  If configured false up or down button can be grayed

Please, don't. That's a regression if you change the way assumed state works.

I use RF-based covers, and it needs assumed state with up and down buttons still enabled even if at end positions, because RF commands are missed in some cases (433MHz, when there's interference etc.) Since there's no feedback of any kind from the cover, nor physical buttons used, it needs to always have the up and down buttons available to restore sync.

nagyrobi:
Perhaps this logic could be implemented: https://github.com/nagyrobi/home-assistant-custom-components-cover-rf-time-based see _Services to set position or action without triggering cover movement_ `set_known_position` and `set_known_action` service calls which seem more understandable.

It would be more universally usable with binary sensors sensing the manual switches (or receiving the RF codes) and puiblish the state towards HA with the new real situation.

X-Ryl669:
I think we're trying to fit many different scenario in a single component. I'm listing here the different scenario I'm thinking of depending on the current code:

## Cover without manual switch control
### Cover without end stops detector
In that case, the current default for the time_based cover is good to use. At startup, the cover position isn't known until it's operated. After first operation, the state tracked by HA/ESPHome is the real state since there isn't any other way to change it. 
**The `assumed_state` is correct and can be used for graying useless buttons**
The motor's phases aren't energized continuously, so a stop action **must** be triggered to stop the relay and prevent energizing 2 opposite phase of the motor. 
*Note: I'm not speaking of interlock group, it's possible to wire the relays to avoid any interlock here*

### Cover with (integrated) end stops detectors
In that case, it's not a problem to keep a phase energized since the end stop detector will prevent mechanical failure. 
So the `has_built_in_endstop` is set to true, and ESPHome allow to re-trigger the close action even if the `assumed_state` is already closed, because it doesn't risk any damage. However, the stop action is never triggered, so at least one phase of the motor is always energized.
**The `assumed_state` is correct and can be used for graying useless buttons**

## Cover with a manual switch control (like a mechanical switch or remote RF trigger)
###  Cover without end stop detector
In that case, the `assumed_state` that's computed by ESPHome/HA isn't valid since there isn't any way for ESPHome to detect when the cover is operated manually. The `manual_control` option simply enable a different logic in ESPHome's timebased_cover to allow operating the switch even if the desired state is already the same as the assumed state. The stop action is triggered anyway to still avoid energizing the motor phase when not required. 
After HA/ESPHome operates the cover, the state is right, so any automation is still applied appropriately.
**The `assumed_state` is incorrect and can not be used for graying useless buttons**

### Cover with end stop detector
Currently, there isn't any mechanism in ESPHome / timebased_cover to query/monitor some external input and understand them as end stop detector (or position detector). Here the solution is probably to use another component, like the rf_timebased_cover that have some service that can be called to re-synchronize the assumed state of the cover in HA with the actual state of the cover.

### Cover with end stop detector but a motor logic wired in
Those are very common in EU. The motor will work as usual with 2 phases. However, when the 2 phases are energized, the motor will learn its current position as the end (or the start) of the travel. If the 2 phases are energized any unexpected time, this is very painful, since then the motor will refuse to go further (it can be half closed and never close until it's reset). 
The `has_built_in_endstop` option is thus useless for these motors, since because one phase is left energized, when operated manually, it "learn" the current position (which is typically the opposite of what you want) as the "end" position.
In that case, `manual_control` is the only way to use the system safely.
**The `assumed_state` is incorrect and can not be used for graying useless buttons**

nagyrobi:
I think you're fundamentally wrong about the original purpose of `assumed_state`. De [docs](https://developers.home-assistant.io/docs/core/entity/#generic-properties) say:
kép
That's it and nothing more. 
`assumed_state` is by design intended to let the user interface handle all the buttons. You're likely confusing it with the `optimistic` setting which can be seen in the [Template Cover](https://www.home-assistant.io/integrations/cover.template/#optimistic-mode), for example.

Also, you're misunderstanding the problem with the RF cover - where not only we're missing any state feedback, but there's also the chance that the RF command didn't succeed to start anything - the worse of all - thus the real state of the cover can be really an assumption, nothing more. Your sentence 

> The `assumed_state` is incorrect and can not be used for graying useless buttons

is invalid, because `assumed_state` _has to be_ set to `true` for this kind of usage, since there are no useless buttons here. All buttons are always useful, because you never know when you're gonna run into an unsynced state. You always want to have all the buttons usable for RF covers.

X-Ryl669:
You're probably right. Like I said above, I couldn't figure out the meaning of the `assumed_state` variable. The [documentation in timebased_cover](https://github.com/esphome/esphome-docs/blob/c33b6978847e94caabb3aec0600909d15a27d443/components/cover/time_based.rst) is:

```
assumed_state (Optional, boolean): Whether the true state of the cover is not known. 
This will make the Home Assistant frontend show buttons for both OPEN and CLOSE actions, instead of hiding 
or disabling one of them. Defaults to true.
```
 which is the opposite of what you've linked with.

Since there is no code in timebased_cover component using it (there's only code to set it), I couldn't find out what it was used for and I didn't look into the base class. 

I'm not saying that you should disable `assumed_state` variable in my previous post. In fact, the variable name is confusing, since the timebased cover maintains a state (it's **the** *assumed state* for the Cover base class, but not the configuration's `assumed_state` that should be better called `use_tracked_state` IMHO) and that's what I was referring to about the correctness of this variable. 

Anyway, I invite you to try the `manual_control` PR I've submitted. It's doing exactly what you want. It allows to click "close" on the interface and have the cover to close. If the cover is closed, it does nothing (but energizes the motor). If the cover has been opened via RF or a manual switch, it closes. After it has operated, it stops energizing the motor unlike `has_built_in_endstop`, so it can be manually or RF operated again. Also, by default (without touching the `assumed_state` configuration variable), the buttons in the UI aren't disabled so you can use them.
Also, you don't need to use any trick to capture the RF signal in HA to set the cover state. With that change, HA displayed state might be wrong, but as soon as it's being operated, it'll fix itself to the right state, so that's a good compromise without painful setup and/or hardware.

X-Ryl669:
Not stale: waiting for consensus on documentation.

glmnet:
Let's forget about assumed state, it is ok as it is:

> which is the opposite of what you've linked with.

No, it's the same, just wording is a bit confusing:
`assumed_state` is a HA setting (please don't even try to change the name) it is just forwarded to HA so UI behaves like it's documented...
When it is set to true, "state is based on assumption" and in this case up and down button will be always enabled.
When is set to false, we tell HA we are not assuming the state, so it is allowed to gray / hide the buttons on the UI based on the position feedback. See my screenshot.

To not confuse ourselves, in ESPHome we have the "optimistic" keyword which will tell us if the device (e.g. a cover) is tracking the state in the ESP memory or is it grabbing it from the actual hardware via sensors to forward the actual state to HA. We don't use _assumed state_ for this.

So a time based cover is optimistic by definition as we use time to "assume" the position (we don't have a sensor to tell us the real position), yet we don't hard code the assumed state and let the user decide.

I see how you are uncomfortable with my rewrite of the documentation by using "assumed state", I invite you to try to rewrite the doc without this keyword but make sure you communicate what the component is doing and what is going to happen. (I'm ok with the code side of this PR)

1. It must explain that "if the cover is commanded to open or close the corresponding actions will be performed without checking current state." (same wording from `has_built_in_endstop`) 
2. It should be made clear that in this case **the whole time will have to elapse again** before the component will call stop. Imagine a motor gets stuck because it lacks an auto stop feature. i.e. the system should be tolerant to set the motors to open/close the cover even when the cover is fully open/closed.
3. May be this should suggest configuring `assumed_state` to true (even though it's the default) as this new option won't be able to be used from HA if the button would be disabled?

X-Ryl669:
What about:


- **manual_control** (*Optional*, boolean): If your cover has a manual control switch, it's not 
  possible for the system to be sure of its current position. If the cover is commanded to open or close, the command will be performed completely (including any defined ``stop_action``) without checking the current state. The command duration will follow the programmed duration. The current state will then be relearned upon completion.
It's recommended to  set ``assumed_state`` to true so the cover control button aren't disabled in the interface.
  Defaults to ``false``.

glmnet:
To my understanding this is not 100% correct, by reading your PR I understand the duration of the open could last 50% if the cover internal state is 0.5?

By reading this explanation I could pick that an open / close command will always run for the programmed duration.

Or I misunderstood the code?

X-Ryl669:
> To my understanding this is not 100% correct, by reading your PR I understand the duration of the open could last 50% if the cover internal state is 0.5?

If the internal state is 0.5, then the test that read `if (this->manual_control_ && (pos == COVER_OPEN || pos == COVER_CLOSED)) {` shouldn't pass, since `pos` isn't open or closed but in between (IIUC). So it'll not trigger anything, usual behavior with OPEN or CLOSE will happen, as expected. 

It would only trigger if ESPHome thinks the cover is resting at a final position (opened or closed) and asked to go in the same direction (resp. open or close), and in that case, it'll run the asked action for the defined/programmed **complete** duration despite ESPHome thinking it's already at the required position.

I agree that this sentence is confusing, so I'll remove it, it's useless, since the previous sentence says "the command will be performed completely" which is more accurate, whatever the real state of the cover.

glmnet:
I'm sorry to go back and forth and not finishing this thread.

This is software documentation, it is a property description. You should focus on what this does, so avoid trying to sell something to the user, focus on what the configuration does and avoid trying to guess what the user problem is and give a magic solution to it.

You should give a short concise and complete description of the property, not why somebody would need it, if you want you can add a link on it and explain in a small heading below why the property can be useful and give more information about it.

> **manual_control** (*Optional*, boolean): If your cover has a manual control switch, it's not 
  possible for the system to be sure of its current position. If the cover is commanded to open 
  or close, the command will be performed completely (including any defined ``stop_action``) 
  without checking the current state.
  The current state will then be relearned upon completion.
  It's recommended to  set ``assumed_state`` to true so the cover control button aren't disabled 
  in the interface. Defaults to ``false``. 
  
It is still not clear from this description what the duration of the open or close action will be if it is starting from 0.5 or from 1.0 or from 0.0, this is really what the code companion PR is doing and this doc is not explaining that. Please read my [proposed version](https://github.com/esphome/esphome-docs/pull/2571#issuecomment-1374030070) of the documentation again and let's see if we can get it towards that detail (ignoring the assumed state thing)

> It would only trigger if ESPHome thinks the cover is resting at a final position (opened or closed) and asked to go in the same direction (resp. open or close),

I believe your current documentation is not explaining this.

X-Ryl669:
Can you review the updated trial ? It should match your expectations. Thanks!

jesserockz (approved):

PR #2573, opened 2023-01-04
Update rotary_encoder.rst

lillborje71:
changed # Templated to match #in some trigger
 old yaml
  # Templated
  - sensor.my_rotary_encoder.publish: id: my_rotary_encoder value: !lambda 'return -1;'

correct yaml

      # Templated
      - sensor.rotary_encoder.set_value:
          id: my_rotary_encoder
          value: !lambda 'return -1;'

## Description:

## Checklist:

    or

### DIFF
--- a/components/sensor/rotary_encoder.rst
+++ b/components/sensor/rotary_encoder.rst
@@ -112,7 +112,7 @@ to the newly set internal value.
           value: 10
 
       # Templated
-      - sensor.my_rotary_encoder.publish:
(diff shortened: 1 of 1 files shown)


nagyrobi:
Is `.publish` deprecated?

ssieb:
> Is `.publish` deprecated?

The original example was just wrong, but in general, you shouldn't publish to a sensor that's not a template.  In this case, it would be a temporary effect, because the encoder still has its internal state and would continue from there.

PR #2588, opened 2023-01-11
Clarify bang bang sensor needs °C

poldim:
This isn't stated elsewhere in the docs and I spent a while figuring out why it wasn't working when everything else was set to `°F`

## Description:

## Checklist:

    or

### DIFF
--- a/components/climate/bang_bang.rst
+++ b/components/climate/bang_bang.rst
@@ -69,7 +69,7 @@ Do note that the actions are only called when the current temperature leaves the
 Configuration variables:
 ------------------------
 
-- **sensor** (**Required**, :ref:`config-id`): The sensor that is used to measure the current temperature.
+- **sensor** (**Required**, :ref:`config-id`): The sensor that is used to measure the current temperature (must be in `°C`).
 - **default_target_temperature_low** (**Required**, float): The default low target temperature for
   the control algorithm. This can be dynamically set in the frontend later.
 - **default_target_temperature_high** (**Required**, float): The default high target temperature for


nagyrobi:
> spent a while figuring out why it wasn't working

What was the symptom?

poldim:
> > spent a while figuring out why it wasn't working
> 
> What was the symptom?

sensor value is too high to ever trigger the climate module for turning anything on

nagyrobi:
How did you fix it?

ssieb:
If you're going to put it in a specific component, you need to put it in all of them.
Also see https://esphome.io/components/climate/tuya.html for the `reports_fahrenheit` option.

poldim:
> How did you fix it?

```yaml
- platform: homeassistant
  id: ha_temp_front
  entity_id: sensor.living_room_temperature
  internal: true
  filters:
  - lambda: return (x - 32.0) * (5.0/9.0);
```

poldim:
> If you're going to put it in a specific component, you need to put it in all of them. 

I fix what I can.  I really hope that's not the perspective on an FOSS project.  

> Also see https://esphome.io/components/climate/tuya.html for the `reports_fahrenheit` option.

I don't follow, that's a different component.  Are you saying the `reports_fahrenheit` option also exists for the bang bang controller and that it's just not documented?

ssieb:
> I don't follow, that's a different component. Are you saying the `reports_fahrenheit` option also exists for the bang bang controller and that it's just not documented?

No, that was just a verification that you're correct.  °C is assumed.

> > If you're going to put it in a specific component, you need to put it in all of them.
> 
> I fix what I can. I really hope that's not the perspective on an FOSS project.

I suppose "need" is a little strong, but there's not much point in changing it in only the one place.

nagyrobi:
> °C is assumed

Should we put a note in the [Base (Core) Climate](https://esphome.io/components/climate/index.html) doc instead? That would then apply to all.

PR #2595, opened 2023-01-14
Added power negative filter example - very useful

clowrey:
## Description:

## Checklist:

    or


clowrey:
This is my first time trying to improve the documentation so maybe I didn't do it absolutely correct :) But having the negative filter example on main page would be really helpful as it was assumed behavior for me that the value would go negative but it does not of course.. So just having another example there would have saved me lots of time and I am sure others as the issue I eventually found the solution in shows: https://github.com/esphome/issues/issues/2949

PR #2652, opened 2023-02-08
Update hdc1080.rst

meijerwynand:
HDC1080 library compatible with CHT8305

## Description:

## Checklist:

    or
  - [ x ] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.

### DIFF
--- a/components/sensor/hdc1080.rst
+++ b/components/sensor/hdc1080.rst
@@ -58,6 +58,24 @@ Currently, the platform doesn't support activating the built-in heater,
 as it seems to only be rarely of use. If you need it, please open an
 issue.
 
+Notes
+--------
+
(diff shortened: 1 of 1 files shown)


nagyrobi:
I'm a bit reluctant about "appears to be" phrasing in a product documentation. Is it 100% or not?

meijerwynand:
I do not have another wired temperature sensor to compare it, only bluetooth sensors, I am also unsure which of these 2 are "more accurate" as there does appear to be a ~1 degree difference.

The blue line is the bluetooth temperature sensor and the purple the wired.

image

The interruption ~5:30pm was due to a router reboot on my side and not missing sensor data.

image

To your question:
Is it functioning with the HDC1080 library - Yes.
Is it accurate - I do not know.
Is it working - It appears to be :)

PR #2723, opened 2023-03-06
Add absolute humidity component

DAVe3283:
## Description:

Adds a component to calculate the absolute humidity provided air temperature and relative humidity from other sensors.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4519

## Checklist:

    or


components/sensor/absolute_humidity.rst
```diff
@@ -0,0 +1,60 @@
...
+      # Use any temperature and relative humidity source, e.g. a BME280:
+      - platform: bme280
+        temperature:
+          name: Temperature
+          id: air_temperature
+        pressure:
+          name: Air Pressure
+        humidity:
+          name: Relative Humidity
+          id: relative_humidity
```
jesserockz: Lets just leave this as a blank template as people will copy paste this and it wont match their setup
```suggestion
      - platform: ...
        temperature:
          name: Temperature
          id: air_temperature
        humidity:
          name: Relative Humidity
          id: relative_humidity
```

jesserockz (changes requested):
Im fine with you adding the link to the different sensors, but not sure about deleting all the references between sensors. Lets leave them there in this PR and we can visit removing/tidying in a new PR if need be...
Same with altering the scd4x example yaml.
I am fine with updating the bme280 example to use this sensor.

DAVe3283:
I had noticed it was very inconsistent what sensors linked to what other sensors, and it seemed that was all a holdover from when there were only a handful of sensors supported. I agree, that is out of scope for this PR though. Once this is all merged in, I can create separate PRs to clean up the issues I noticed in the sensor components.

jesserockz:
The reason it is inconsistent is because people don't go back and add them after new ones are added. It would actually be better to link to the homepage category for the similar sensor type tbh.

jesserockz (approved):

PR #2763, opened 2023-03-18
Add clarity for use of multiplier

rcblackwell:
The use of multiplier wasn't really clear to me until explained as described below. I suggest, and have added to the document the following description for consideration;

## Description:

For seasonal changes, it’s easier to use the multiplier to adjust the watering time instead of adjusting the run durations directly. Set your run duration to the time you want for ‘mild’ weather (spring/autumn) and then use the multiplier to increase (summer) or decrease (winter) the calculated run times.


nagyrobi:
@kbx81

kbx81 (approved):

PR #2783, opened 2023-03-25
Update doc to add multiple dallas hubs

lboue:
## Description:
Update doc to explain how to add multiple dallas hubs.


lboue (commented):
Fix Title underline too short

PR #2797, opened 2023-04-01
Remote receiver improvements

Mat931:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4642

## Checklist:

    or


nagyrobi:
What are the minimum and maximum possible/allowed values for `clock_divider`?

Mat931:
@nagyrobi 1 and 255 according to https://docs.espressif.com/projects/esp-idf/en/v4.3/esp32/api-reference/peripherals/rmt.html#common-parameters. I added a config validation for this.

nagyrobi:
I think it would be appropriate to mention in the doc the min/max possible values for all the new parameters.

nagyrobi:
@Mat931 wonderful, thanks.

components/remote_receiver.rst
```diff
@@ -55,18 +55,22 @@ Configuration variables:
   - **sony**: Decode and dump Sony infrared codes.
   - **toshiba_ac**: Decode and dump Toshiba AC infrared codes.
 
-- **tolerance** (*Optional*, int): The percentage that the remote signal lengths can deviate in the
-  decoding process. Defaults to ``25%``.
+- **tolerance** (*Optional*, int or :ref:`config-time`): The percentage or time that the remote signal lengths can
+  deviate in the decoding process. Allowed values are in range ``0..4294967295%`` or ``0..4294967295us``. Defaults to
```
kbx81: When I first read this I thought the double-decimal thing was a typo. Let's just write it out.
```suggestion
  deviate in the decoding process. Allowed values are in range ``0`` to ``4294967295%`` or ``0`` to ``4294967295us``. Defaults to
```

components/remote_receiver.rst
```diff
@@ -55,18 +55,22 @@ Configuration variables:
...
+  ``25%``.
 - **buffer_size** (*Optional*, int): The size of the internal buffer for storing the remote codes. Defaults to ``10kB``
   on the ESP32 and ``1kB`` on the ESP8266.
-- **memory_blocks** (*Optional*, int): The number of RMT memory blocks used. Only used on ESP32 platform. Defaults to
-  ``3``.
+- **memory_blocks** (*Optional*, int): The number of RMT memory blocks used. Only used on ESP32 platform. The maximum
+  number of blocks shared by all receivers and transmitters is ``8``. Defaults to ``3``.
 - **filter** (*Optional*, :ref:`config-time`): Filter any pulses that are shorter than this. Useful for removing
-  glitches from noisy signals. Defaults to ``50us``.
+  glitches from noisy signals. Allowed values are in range ``0..4294967295us``. Defaults to ``50us``.
```
kbx81: Same as above.
```suggestion
  glitches from noisy signals. Allowed values are in range ``0`` to ``4294967295us``. Defaults to ``50us``.
```

components/remote_receiver.rst
```diff
@@ -55,18 +55,22 @@ Configuration variables:
...
 - **idle** (*Optional*, :ref:`config-time`): The amount of time that a signal should remain stable (i.e. not
-  change) for it to be considered complete. Defaults to ``10ms``.
+  change) for it to be considered complete. Allowed values are in range ``0..4294967295us``. Defaults to ``10ms``.
```
kbx81: Same as above.
```suggestion
  change) for it to be considered complete. Allowed values are in range ``0`` to ``4294967295us``. Defaults to ``10ms``.
```

components/remote_receiver.rst
```diff
@@ -55,18 +55,22 @@ Configuration variables:
...
 - **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation. Use this if you have
   multiple remote receivers.
+- **clock_divider** (*Optional*, int): The clock divider used by the RMT peripheral. A clock divider of ``80`` leads to
+  a resolution of 1 µs per tick, ``160`` leads to 2 µs. Allowed values are in range ``1..255``. Only used on ESP32
```
kbx81:
```suggestion
  a resolution of 1 µs per tick, ``160`` leads to 2 µs. Allowed values are in range ``1`` to ``255``. Only used on ESP32
```

kbx81 (dismissed):

components/remote_receiver.rst
```diff
@@ -56,18 +56,22 @@ Configuration variables:
   - **sony**: Decode and dump Sony infrared codes.
   - **toshiba_ac**: Decode and dump Toshiba AC infrared codes.
 
-- **tolerance** (*Optional*, int): The percentage that the remote signal lengths can deviate in the
-  decoding process. Defaults to ``25%``.
+- **tolerance** (*Optional*, int or :ref:`config-time`): The percentage or time that the remote signal lengths can
+  deviate in the decoding process. Allowed values are in range ``0`` to ``4294967295%`` or ``0`` to ``4294967295us``. Defaults to
```
jesserockz: Does that percentage make sense?
```suggestion
  deviate in the decoding process. Allowed values are in range ``0`` to ``100%`` or ``0`` to ``4294967295us``. Defaults to
```

Mat931: Sure, we can limit the value to 100% but in my opinion the percentage based tolerance doesn't make much sense at all since every protocol has different needs. Ideally the acceptable timing range should be defined by each protocol and not globally by the receiver.

jesserockz (approved):

PR #2801, opened 2023-04-03
D1 D2 is wrong

hiteshkher:
need code run font part

## Description:

## Checklist:

    or

### DIFF
--- a/components/display/ssd1306.rst
+++ b/components/display/ssd1306.rst
@@ -30,15 +30,15 @@ improve reliability.
 
 .. code-block:: yaml
 
-        # Example configuration entry for ESP8266
-        i2c:
-          sda: GPIO4 # GPIO21 for ESP32
-          scl: GPIO5 # GPIO22 For ESP32
-       font:
-        # gfonts://family[@weight]
-       - file: "gfonts://Roboto"
-           id: my_font
-         size: 20
+    # Example configuration entry for ESP8266
+    i2c:
+      sda: GPIO4 # GPIO21 for ESP32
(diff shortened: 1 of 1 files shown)


hiteshkher (commented):
D1 and D2 is wrong also code run need font data and id

nagyrobi:
Why was it wrong?

ssieb:
The examples are not intended to necessarily be fully complete.  They only demonstrate how to setup the specific component.  So you are expected to look at the general display docs to find out how to use fonts.
Also, `D1` and `D2` are not "wrong".  They might not work for your particular board, but again, you're expected to understand about pin definitions and put the ones there that you need.

hiteshkher:
Wrong like sda is not D1 it's D2
                    Scl is not D2 it's D1
So everyone is not cross verify with pin layout and his project is not run
and lot of confusion also not mantion font requirment is not mantion in
this page so biggner have so much time consuming example i spend around 8
to 10 hour  for just display run with basic code so i think if every
example have basic
Code lot of help for biggner.

ssieb:
We're not going to put a full display example on every display component.  You are going to have to look at the display docs to find out how to draw things on the screen anyway.  I would even consider removing the `i2c:` part out of the example as well since it's not really relevant and should just say that you need an `i2c` component configured as well.

ssieb:
Btw, the `SDA` and `SCL` assignments to `D1` and `D2` are arbitrary and only suggestions.  i2c on ESP8266 is software, so it can use most pins.

hiteshkher:
Ok sir
Thank you

PR #2817, opened 2023-04-11
Update modbus_controller.rst

victorclaessen:
## Description:

I think at some time in the past the 'command_throttle' parameter changed from a unit of 'milliseconds' to the unit of 'time'.

If you try to compile with a value of 1000 (supposedly milliseconds), compilation fails with the following error:
```
    Don't know what '1000' means as it has no time *unit*! Did you mean '1000s'?.
    command_throttle: 1000
```
If I put a value of '1s' then compilation works and also this is the time I see that is being used.

**Related issue (if applicable):** no issue has been raised yet

## Checklist:

    or


nagyrobi:
Thank you

PR #2819, opened 2023-04-12
Add some missing binary sensors to the index

kbx81:
## Description:

Following up on https://github.com/esphome/esphome-docs/pull/2802 -- add a few binary sensors to the index so they may be more easily found.

## Checklist:

    or


nagyrobi:
Thank you

PR #2837, opened 2023-04-13
Add mdi image documentation

guillempages:
## Description:
Add support for automatic download of MDI images

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4654

## Checklist:

    or

### DIFF
--- a/components/display/index.rst
+++ b/components/display/index.rst
@@ -660,14 +660,24 @@ Use this component to store graphical images on the device, you can then draw th
         id: my_image
         resize: 100x100
 
+.. code-block:: yaml
+
+    image:
+      - file: mdi:alert-outline
+        id: alert
+        resize: 80x80
+
 Configuration variables:
 
-- **file** (**Required**, string): The path (relative to where the .yaml file is) of the image file.
+- **file** (**Required**, string):
+
+  - **Local files**: The path (relative to where the .yaml file is) of the image file.
+  - **Material Design Icons**: Specify the `Material Design Icon <https://pictogrammers.com/library/mdi/>`_ id in the format ``mdi:icon-name``, and that icon will automatically be downloaded and added to the configuration.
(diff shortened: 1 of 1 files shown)


components/display/index.rst
```diff
@@ -660,14 +660,21 @@ Use this component to store graphical images on the device, you can then draw th
...
+
+    image:
+      - file: mdi:alert-outline
+        id: alert
+        resize: 80x80
+
 Configuration variables:
 
-- **file** (**Required**, string): The path (relative to where the .yaml file is) of the image file.
+- **file** (**Required**, string): The path (relative to where the .yaml file is) of the image file. Instead of a local file, you can specify a *Material Design Icon* id in the format ``mdi:icon-name``, and that icon will automatically be downloaded and added to the configuration.
```
Unfocused: Suggest adding link to either the MDI homepage or index of icons, since some people either won't know what MDI is or won't know how to find the list of supported icons (and even if they do, it's good to reduce friction/be helpful).

jesserockz: I think this text is too long and could be broken up into a list something like this:
```suggestion
- **file** (**Required**, string): 
  
  - **Local files**: The path (relative to where the .yaml file is) of the image file. 
  - **Material Design Icons**: Specify the *Material Design Icon* id in the format ``mdi:icon-name``, and that icon will automatically be downloaded and added to the configuration.
```

jesserockz (changes requested):

components/display/index.rst
guillempages: @Unfocused I've added the link to the same page that HomeAssistant is referencing; I think it's good to keep that consistent.
@jesserockz Your suggestion makes sense. I've taken it in.

jesserockz (approved):

PR #2844, opened 2023-04-15
Add ABB-Welcome / Busch-Welcome Door Intercom Protocol

Mat931:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4689

## Checklist:

    or


nagyrobi:
Which modules from ABB use this? I've looked at their site but only found video-over IP stuff...

Mat931:
@nagyrobi Look in the category "ABB-Welcome" (the one without IP) [here](https://new.abb.com/low-voltage/products/building-automation/product-range/abb-welcome) or "Busch-Welcome" [here](https://www.busch-jaeger.de/en/smarter-home/systems/busch-welcomer)
Example Model Number: 83222 U-625

They are quite common in Europe. More info: https://developer.nuki.io/t/abb-or-busch-welcome-series/3026

nagyrobi:
How do these (non-IP) transmit audio/video to each other? Also RF? Scrolled through the manual, only need a wire pair for power?

Mat931:
No, there's no RF. Everything is sent over the two wires. Not sure about the audio but the video signal is analog. They are using an ADV7180 for decoding. As you can imagine, the video quality is not the best and they only let you save 480x360 jpegs, no videos. Can't recommend haha

nagyrobi:
> No, there's no RF.

What is this component for, if that's the case?

Mat931:
@nagyrobi For sending and receiving messages on that two wire bus. You need a custom circuit: https://github.com/esphome/esphome/pull/4689
Where would be the best place in the docs to put information like this?

nagyrobi:
Oops... That needs a very clear note in the doc.
I suggest to definitely state in the description that the signal is intended to be transmitted or received over copper wires, with an adaptor circuit referred at the end of the page. You should make a note box at the bottom, describing the circuit and noting what devices it is compatible with.

nagyrobi:
Do you have any info about how's audio transmitted over the wires? Since we're starting to support mic/line input on ESPHome, this could be a nice extension to a future smart intercom through HA...

Mat931:
No, sorry. I would need to do some more reverse engineering of the PCB and it would certainly require extra hardware.
Is there any support in HA besides giving my delivery guy access to my voice assistant?
AFAIK the audio does not pass through the CPU of the indoor station. There are some audio ICs but I don't remember the details.

jesserockz (approved):

PR #2873, opened 2023-04-24
Add new esp32 bt classic & bt classic presence components.

RoboMagus:
## Description:

Documentation for new components: ESP32 BT Classic & BT Classic Presence.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4736

## Checklist:

    or

### DIFF
--- a/components/binary_sensor/bt_classic_presence.rst
+++ b/components/binary_sensor/bt_classic_presence.rst
@@ -1,0 +1,40 @@
+ESP32 Bluetooth Classic Presence
+================================
+
+.. seo::
+    :description: Instructions for setting up BT Classic binary sensors for the ESP32.
+    :image: bluetooth.svg
+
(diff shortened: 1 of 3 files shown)


nagyrobi:
Are some specific devices needed for this?

RoboMagus:
Nope, this should work just fine with most android phones. Anything that has regular non BLE Bluetooth.

(also no need for pairing, knowing the MAC is sufficient)

nagyrobi:
So what it does is to just monitor classic bletooth packets in the air, right?

RoboMagus:
Not quite. Unlike ble, BT Classic does not advertise itself like that (and in some cases if it is discoverable it may show randomized Mac addresses).
The functionality is described in the main PR. It requests registered BT devices for their names using their known (real) MAC address. If they respond, they are present.

PR #2883, opened 2023-05-01
display font: document using gfonts:// for icons

peralta:
Added example showing how to use gfonts to download icons. It seems it is not widely known.

## Description:

Using glyphs from Google's material design library is very common but examples and real usage shows that people are manually downloading the font rather than using esphome's already existing gfonts great helper.

This PR adds an example and explicitly mentions that this is possible.

## Checklist:


nagyrobi:
I suggest to also add a `note` about how you calculated the "\U0000e425" codepoint, as it's not straightforward from the MDI site.

peralta:
> I suggest to also add a `note` about how you calculated the "\U0000e425" codepoint, as it's not straightforward from the MDI site.

That is a great suggestion. I found the amount of external links in esphome-docs to be relatively low, so I wonder if such *ad-hoc* instructions are welcome. Happy to provide them.

nagyrobi:
Put it in a `note` box.

nagyrobi (changes requested):
Please add a `note` box with a small explanation on is calculated the "\U0000e425" codepoint, as it's not straightforward from the MDI site.

nagyrobi (approved):

PR #2906, opened 2023-05-09
grove TB6612FNG

max246:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome/pull/4797) with YAML changes (if applicable):** esphome/esphome#<esphome PR number goes here>

## Checklist:

  - [ x ] I am merging into `next` because this is new documentation that has a matching pull-request in [esphome](https://github.com/esphome/esphome) as linked above.  
    or

  - [ x] Link added in `/index.rst` when creating new documents for new components or cookbook.


max246:
Need a reviewer :/

jesserockz (approved):

glmnet:
This is not good. Check https://next.esphome.io/#motor-components no table format here?? May be this should go into misc components?

Also the doc file is `grove-tb6612fng` which looks ok to me as this implements the chip with that name, but the esphome component is `grove_i2c_motor`. This is just wrong; no other components mismatch like this. Easiest will be to rename the doc file, but I believe naming the component as the chip is much better.

Also same image grove_tb6612fng.jpg is in two folders.

max246:
> This is not good. Check https://next.esphome.io/#motor-components no table format here?? May be this should go into misc components?
> 
> Also the doc file is `grove-tb6612fng` which looks ok to me as this implements the chip with that name, but the esphome component is `grove_i2c_motor`. This is just wrong; no other components mismatch like this. Easiest will be to rename the doc file, but I believe naming the component as the chip is much better.
> 
> Also same image grove_tb6612fng.jpg is in two folders.

I will open a PR to amend the naming in the main core and another one to fix the table in the doc. thank you for pointing it out.

PR #2928, opened 2023-05-17
Add mute pin to i2s_audio/media_player

rpatel3001:
## Description:

Document the mute pin parameter for i2s_audio/media_player.

Remove the Pin Schema link for the i2s_dout_pin parameter, replaced with simple Pin.

## Checklist:

    or


kroimon:
> As an aside: The i2s_dout_pin parameter is not treated as a pin schema in the code, either the code or the docs should be updated to match.

`i2s_dout_pin` is copied over to [`i2s_pin_config_t`](https://docs.espressif.com/projects/esp-idf/en/v3.3/api-reference/peripherals/i2s.html#structi2s__pin__config__t) which is only a single pin number, so the docs need to be updated.

rpatel3001:
Updated to also fix the Pin Schema -> Pin issue on i2s_dout_pin

kroimon (approved):

jesserockz (approved):
Not sure how I missed this. Thanks

PR #2953, opened 2023-05-29
Update esp32_dac.rst

madsdyd:
Add example of changing value of the dac using a custom service callable from Home Assistant.

## Description:

## Checklist:

    or
  - [ X] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.


jesserockz (changes requested):
This is not something we put in a main config example. It would go at the bottom of the page.

And also we should encourage the use of a template number instead of an API service for something like this. 
Feel free to reopen if you change to have an example using a number.

PR #2970, opened 2023-06-01
Update gps.rst

ohkaja:
## Description:

## Checklist:

    or
  - [ x] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.


jesserockz:
This is specifically talking about the UART bus config, which means the rx pin. Changing it would just be confusing

ohkaja:
As noob i was confused. So just to help others out it would be right to write that tx from the gps board must be connectet to the rx on the esp board.

Saw much threads on the net wich came up with the same problem.

But when you mean this would be more confusing. Ok.

PR #2999, opened 2023-06-17
bluetooth_proxy: Clarify that only BLE devices are supported.

kpfleming:
## Description:

The `bluetooth_proxy` component docs never mention BLE, which could lead users to believe that it supports non-BLE Bluetooth devices... but it does not.

## Checklist:

    or


nagyrobi (approved):
@jesserockz really? Only BLE supported? Wasn't aware...

jesserockz (approved):

PR #3000, opened 2023-06-17
Added hints about repetition of signals

DominikBitzer:
## Description:

While working on setting up me 433-power-outlets, I was easily getting them to work with plain Arduino-studio code using rc-switch library and examples. I was happy to see ESPHome uses the same library, but had problems getting this to work, since the standard RCSwitch.send doesn't match remote_transmitter behavior.
Finally this was due to the repetitions. This PR builds on an earlier PR, that was not merged. I tried to include OttoWinters original suggestions while keeping the documentation short and clear.
I hope this PR would help future users, since I saw the issue with the repetitions and wait_time on several forum posts.

Let me know what you think and if there is anything I could improve :)

**Related issue (if applicable):** fixes or updates PR https://github.com/esphome/esphome-docs/pull/290

## Checklist:

    or
  - [ x ] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.

### DIFF
--- a/components/remote_transmitter.rst
+++ b/components/remote_transmitter.rst
@@ -898,6 +898,10 @@ earlier, create a new template switch that sends the RF code when triggered:
           - remote_transmitter.transmit_rc_switch_raw:
               code: '100010000000000010111110'
               protocol: 2
+              repeat: 
+                times: 10
+                wait_time: 0s
+
 
     # Or for raw code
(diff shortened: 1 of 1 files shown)


nagyrobi (approved):
Agreed, thanks.

PR #3005, opened 2023-06-19
Add sun_gtil2 component

Mat931:
## Description:
Documentation for the sun_gtil2 component

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#4958

## Checklist:

    or


components/sun_gtil2.rst
```diff
@@ -0,0 +1,123 @@
...
+need to set the ``baud_rate`` to 9600.
+
+.. code-block:: yaml
+
+    # Example configuration entry
+    uart:
+      - id: control_to_display
+        rx_pin: GPIO16
+        baud_rate: 9600
+
```
jesserockz: Dependent components are not to be included in examples
```suggestion
```

components/sun_gtil2.rst
```diff
@@ -0,0 +1,123 @@
...
+        state:
+          id: gtil_state
+          name: "State"
+        serial_number:
+          id: gtil_serial_number
+          internal: True
+
+The inverter will report new measurements approximately every 1.3 seconds.
+
+Configuration variables:
```
jesserockz, line 92: Please see `pylontech.rst` for an example on splitting the configuration variables per platform type.

jesserockz (changes requested):

components/sun_gtil2.rst
Mat931: Should I put that somewhere else or just leave it out completely?

jesserockz: Just leave it out, its up to the user to read the docs on uart and configure it according to their physical board and pins.

jesserockz (approved):

PR #3025, opened 2023-06-27
Dark mode

grahambrown11:
## Description:

Implement Dark Mode for Docs 😎 

Added an optional column on the end of the `imgtable::` for invert the image color - add `dark-invert`

## Checklist:

    or


_static/custom.css
```diff
@@ -133,11 +141,10 @@ div.index-shields {
 }
 
 .guide-card {
-  color: #004b6b;
   display: block;
   height: auto;
   width: 300px;
-  background-color: #f3f6f6;
+  background-color: #d5f7f7;
```
jesserockz: Not sure I like this new colour on these cards

_static/custom.css
```diff
@@ -170,23 +181,141 @@ div.index-shields {
...
+    color: #ececec;
+    border-bottom: 1px dotted #ececec;
+  }
+
+  div.sphinxsidebar a:hover, .breadcrumbs a:hover {
+    border-bottom: 1px solid #ececec;
+  }
+
+  .guide-card {
+    background-color: #14504f;
```
jesserockz: I think keeping the guide just a lighter grey in dark mode fits better
```suggestion
    background-color: #393939;
```

conf.py
```diff
@@ -100,7 +100,7 @@
 # default_role = 'cpp:any'
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = "xcode"
+pygments_style = "solarized-dark"
```
jesserockz: I think this makes the text too light in light mode

jesserockz (changes requested):

_static/custom.css
grahambrown11: OK, I thought the green from the chevrons in the picture was the inspiration for the original greyish green... to make it tie in with the dark mode I made it more green, but happy to leave it more grey

conf.py
grahambrown11: That was the only one I found that more or less works for both light and dark... I see Sphinx supports `pygments_dark_style` but not as a user setting, so I'll look at creating a custom theme that extends `alabaster` so we can use `xcode` for light and `solarized-dark` (or I prefer github-dark)

jesserockz (approved):

PR #3042, opened 2023-07-04
Clarify usage of MAC suffix

bhanafee:
## Description:

Clarify that the MAC suffix is applied to both device and friendly name, and describe potential impact on Home Assistant ESPHome Add-on users.

## Checklist:

    or


components/esphome.rst
```diff
@@ -333,6 +335,11 @@ have unique identification for customer installs.
     upstream.
 
 
+Some tools, such as the ESPHome Add-On to Home Assistant, expect the name of the device configuration
+file to match exactly the device name, ``<name>.yaml``. Adding the MAC address suffix may prevent these tools 
+correlating a discovered device to a matching configuration file.
+
+
```
jesserockz, line 342: I don't think we need this section.

```suggestion
```

The whole point of the suffix is for device distribution, not for flashing a bunch of devices then using them personally without their own yaml files on your own dashboard.

When people adopt a device that has a suffix, it is written into both names and then `name_add_mac_suffix` is set to false, giving you a device yaml that matches your new device.

jesserockz (changes requested):

components/esphome.rst
bhanafee: >The whole point of the suffix is for device distribution

The reason I opened the PR in the first place is that the docs lack context for users who *aren't* doing device distribution. This PR isn't suggesting the design be changed. It's to make the documentation clearer. If a user isn't reading it with the thought of "how can I distribute lots of devices" but rather with the thought of "how can I make it easier to manage a bunch of devices that might have similar names" then the side effect might not be obvious.

The specific way it burned me was I noticed the flag exists and it would have been helpful to have the MAC in the friendly name. I wasn't expecting that flag to change the name, nor to cause the HA Add-On to lose track of all my devices. Adding documentation that it also changes both name and friendly_name helps, but I don't think it's nice to expect every new user to realize that the side effect will hose their HA configuration.

PR #3057, opened 2023-07-12
Update sps30.rst to include i2c sample

codyc1515:
## Description:
Cannot use the example as it was as you must include i2c too.

## Checklist:


nagyrobi:
It's clearly written a few sentences above: _The I²C Bus is required to be set up in your configuration for this sensor to work._

The goal is also to make users read, learn and think, not just copy and paste every day.

codyc1515:
This doesn't really make sense for me.

PR #3087, opened 2023-07-24
CAN bus BMS

clydebarrow:
## Description:

Add a new component that captures BMS data via CAN bus. Docs in ESPHome-docs.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#5146

## Checklist:

    or


nagyrobi:
It would be worth adding this to the `canbus` page See Also section...

PR #3177, opened 2023-09-11
Add a doc for PSRAM config

kbx81:
## Description:

Adds a doc for PSRAM configuration.

## Checklist:

    or


components/esp32_psram.rst
jesserockz, line 1: filename should match the actual component name in the code. `psram`

components/psram.rst
```diff
@@ -0,0 +1,31 @@
...
+=====
+
+.. seo::
+    :description: Configuration for the ESP32 PSRAM platform for ESPHome.
+    :image: psram.svg
+
+This component enables and configures PSRAM if/when available on ESP32 modules/boards.
+It is automatically loaded and enabled by components that require it.
+
+PSRAM is not available on the ESP8266.
```
jesserockz: Since there are more than just esp32/esp8266 we should probably flip this
```suggestion
PSRAM is only available on the ESP32.
```

kbx81: Ah yes, good call!

kbx81: ...should we also add "and variants"?

jesserockz (approved):

components/psram.rst
jesserockz: Nah I think its fine

PR #3184, opened 2023-09-13
Updated documentation related to power meter integration

dirrgang:
## Description:
This has gotten a bit larger than expected. As I've added my parent's solar panels to home assistant using ESPHome, I've looked up a lot of documentation and gained some insight in the subject. I figured I'd like to update the existing ESPHome documentation to make it easier for people who want to do the same in future.

For this I've updated the components most often used in such a project, as well as the corresponding cookbook:
- pulse_meter
- pulse_counter
- power_meter
- total_daily_energy

I've added most of the information that I found to be useful while setting up and configuring the device myself. Mostly pertaining to which component is preferable and why, and how to set them up properly.

## Checklist:

    or


dirrgang:
Someone can perhaps double-check my math regarding the minimum pulse width, but it seems to work out.

components/sensor/pulse_meter.rst
```diff
@@ -24,85 +32,80 @@ Configuration variables:
...
 
 - **internal_filter_mode** (*Optional*, string): Determines how the internal filter is applied.
-  One of ``EDGE`` and ``PULSE``. Defaults to ``EDGE``. In ``EDGE`` mode subsequent rising edges are compared and if they fall into an interval lesser than the internal filter value, the last one is discarded. In ``PULSE`` mode the rising edge is discarded if any further interrupts are detected before the internal_filter time has passed. In other words, a high pulse must be at least internal_filter long to be counted. This is useful if you are detecting long pulses that may bounces before and/or after the main pulse.  
+  One of ``EDGE`` and ``PULSE``. Defaults to ``EDGE``. In ``EDGE`` mode subsequent rising edges are compared and if they fall into an interval lesser than the internal filter value, the last one is discarded. In ``PULSE`` mode the rising edge is discarded if any further interrupts are detected before the ``internal_filter`` time has passed. In other words, a high pulse must be at least ``internal_filter`` long to be counted. This is useful if you are detecting long pulses that may bounces before and/or after the main pulse.
```
clydebarrow:
```suggestion
  One of ``EDGE`` and ``PULSE``. Defaults to ``EDGE``. In ``EDGE`` mode rising edges only are detected and pulse width calculated as the time between successive edges.  In ``PULSE`` mode both rising and falling edges are detected and pulse width calculated as the time between a rising and a falling edge. In either case a pulse with width less than ``interval_timer`` is ignored.
```

components/sensor/pulse_meter.rst
```diff
@@ -24,85 +32,80 @@ Configuration variables:
...
 - All other options from :ref:`Sensor <config-sensor>`.
 
 Converting units
 ----------------
 
-The sensor defaults to units of “pulses/min”. You can change this by using :ref:`sensor-filters`.
-For example, if you’re using the pulse meter with a photodiode to
-count the light pulses on a power meter that outputs 1000 pulses per kWh,
-you can use the following to output instantaneous usage in kW:
+The sensor defaults to units of ``pulses/min``. You can change this by using :ref:`sensor-filters`.
```
clydebarrow: 10000 pulses/kWh? AFAIK most meters are 1000.

components/sensor/pulse_meter.rst
```diff
@@ -24,85 +32,80 @@ Configuration variables:
...
 
-    on_...:
-      then:
-        - pulse_meter.set_total_pulses:
-            id: pulse_meter_id
-            value: 12345
-
-.. note::
-
-    This value is the raw count of pulses, and not the value you see after the filters
```
clydebarrow, line 98: Is this note wrong? If not, why remove it?

cookbook/power_meter.rst
```diff
@@ -5,29 +5,19 @@ Non-Invasive Power Meter
...
-power meter which can often cost a bunch of money. However, quite a few power meters
-have a red LED on the front that blinks every time that one Wh has been used.
+So an essential part of making your home smart is knowing how much power it uses over the day. Tracking this can be difficult, often you need to install a completely new power meter which can often cost a bunch of money. However, quite a few power meters have a red LED on the front that blinks every time that one Wh has been used.
 
-The simple idea therefore is: Why don't we just abuse that functionality to make the power-meter
-IoT enabled? We just have to hook up a simple photoresistor in front of that aforementioned
-LED and track the amount of pulses we receive. Then using ESPHome we can instantly have
-the power meter show up in Home Assistant 🎉
+The simple idea therefore is: Why don't we just abuse that functionality to make the power-meter IoT enabled? We just have to hook up a simple photoresistor in front of that aforementioned LED and track the amount of pulses we receive. Then using ESPHome we can instantly have the power meter show up in Home Assistant 🎉
```
clydebarrow:
```suggestion
The simple idea therefore is: Why don't we just abuse that functionality to make the power-meter IoT enabled? We just have to hook up a simple photoresistor in front of that aforementioned LED and track the number of pulses we receive. Then using ESPHome we can instantly have the power meter show up in Home Assistant 🎉
```

cookbook/power_meter.rst
```diff
@@ -36,45 +26,140 @@ And... that should already be it :)
...
+
+    The minimum pulse width threshold is determined by considering the upper limit of the *load* the meter is designed to handle, as well as the meter's *impulse constant* (``x pulses / kWh``). Here's the calculation involved:
+
+    - **Load Limit in Watts**: Establish the *upper load limit* that the meter is designed to measure. For example, if the limit is 16 kW (16,000 Watts), this becomes a reference point.
+
+    - **Pulse Rate Calculation**: Determine the pulse rate corresponding to this load limit. For this we need to know the impulse constant. In our example, the power meter has an impulse constant of ``10000 pulses/kWh``. Dividing both sides by 60s gives us ``166.67 pulses/s = 60 kWs``. This means that ``60 kWs`` corresponds to 1,000 impulses per 6 seconds, or 166.67 impulses per second. Scaling this down to 16 kW, we get ``16 kWs = 44.44 Pulses/s``  (or 400 pulses per 9 seconds). This is the upper bound of the pulse rate we expect to see.
```
clydebarrow:
```suggestion
    - **Pulse Rate Calculation**: Determine the pulse rate corresponding to this load limit. For this we need to know the impulse constant. In our example, the power meter has an impulse constant of ``10000 pulses/kWh``, resulting in 160000 pulses per hour at the maximum 16kW load, i.e. 44.4 pulses/second.
```

cookbook/power_meter.rst
```diff
@@ -36,45 +26,140 @@ And... that should already be it :)
...
+
+    - **Minimum Pulse Width Calculation**: Use the pulse rate to calculate the minimum pulse width threshold. Employ the formula: ``Minimum Pulse Width (seconds) = Time Period / Number of Pulses``. In our example, with a time period of 9 seconds and 400 pulses, the minimum pulse width we expect to see is approximately 22.5 milliseconds. This means you don't want to increase the internal filter time above ``22.5ms``, or you will start to miss pulses within the expected load range.
```
clydebarrow:
```suggestion
    - **Minimum Pulse Width Calculation**: Use the pulse rate to calculate the minimum pulse width threshold. In our example, with a maximum pulse rate of 44.4/sec, the minimum pulse width we expect to see is approximately 22.5 milliseconds. Choose a slightly smaller value than this to avoid missing pulses.
```

cookbook/power_meter.rst
```diff
@@ -36,45 +26,140 @@ And... that should already be it :)
...
 
     .. code-block:: yaml
 
         sensor:
-          - platform: pulse_counter
+          - platform: pulse_meter
             # ...
-            internal_filter: 10us
+            internal_filter: 22.5ms
```
clydebarrow:
```suggestion
            internal_filter: 20ms
```

components/sensor/pulse_meter.rst
```diff
@@ -24,85 +32,80 @@ Configuration variables:
 - **name** (**Required**, string): The name of the sensor.
 
 - **internal_filter** (*Optional*, :ref:`config-time`): If a pulse shorter than this
-  time is detected, it’s discarded and no pulse is counted. Defaults to ``13us``. For S0 pulse meters that are used to meter power consumption 50-100 ms is a reasonable value.
+  time is detected, it’s discarded and no pulse is counted. Defaults to ``13us``.
```
clydebarrow:
```suggestion
  time is detected, it is discarded. Defaults to ``13us``. 
```

components/sensor/pulse_meter.rst
```diff
@@ -24,85 +32,80 @@ Configuration variables:
 - **name** (**Required**, string): The name of the sensor.
 
 - **internal_filter** (*Optional*, :ref:`config-time`): If a pulse shorter than this
-  time is detected, it’s discarded and no pulse is counted. Defaults to ``13us``. For S0 pulse meters that are used to meter power consumption 50-100 ms is a reasonable value.
+  time is detected, it’s discarded and no pulse is counted. Defaults to ``13us``. 
+  
+  To calculate a meaningful value, you need to know the shortest pulse you expect to see, as not to filter out valid pulses.
```
clydebarrow:
```suggestion
  This acts as a debounce filter to eliminate input noise, so choose a value a little less than your expected minimum pulse width.
```

components/sensor/pulse_meter.rst
```diff
@@ -6,8 +6,16 @@ Pulse Meter Sensor
     :image: pulse.svg
 
 The pulse meter sensor allows you to count the number and frequency of pulses on any pin. It is intended to be a drop-in replacement
-for :doc:`integration sensor </components/sensor/pulse_counter>`, but offering better resolution.
-It measures the time between rising edges on a pin, for each pulse it outputs the frequency in pulses/min.
+for the :doc:`pulse counter integration </components/sensor/pulse_counter>`.
+It measures the time between rising edges on a pin, for each pulse it outputs the frequency in ``pulses/min``, as opposed to the pulse counter sensor which outputs the number of pulses in a time interval.
```
clydebarrow:
```suggestion
It measures the _width_ of pulses on a pin, and converts this to an equivalent frequency expressed in pulses per minute, as opposed to the pulse counter sensor which counts the number of pulses in a time interval.
```

clydebarrow (changes requested):

components/sensor/pulse_meter.rst
```diff
@@ -6,8 +6,16 @@ Pulse Meter Sensor
...
+
+This leads to a higher resolution, especially for low pulse rates, as the pulse counter sensor is limited by the number of pulses within a time interval. The difference between ``1 pulse/min`` and ``2 pulses/min`` is much higher than ``2000 pulses/min`` and ``2001 pulses/min``. The pulse_meter is only limited by its accuracy to measure the time between pulses, which leads to an inverse relationship between pulse rate and accuracy. The lower the pulse rate, the higher the accuracy. Lower accuracy with very high pulse rates is less likely to be a problem, as the ESP32 can measure time intervals with a resolution of ``1us``.
```
clydebarrow: You've used back-ticks a lot to emphasise units - back-ticks are meant for code. I'm not sure it's even necessary to emphasise real-world units, but consider other options like ordinary quotes, italic or bold instead.

Just because ESP code can measure with a resolution of 1us does not mean it has an accuracy of 1us. I'd be inclined to remove that sentence.

dirrgang: Mmh, the default was in backticks, I guess I misinterpreted it as a way to also mark units and such.

As to the resolution, the API talks about a microsecond resolution. https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/esp_timer.html

components/sensor/pulse_meter.rst
dirrgang: But it doesn't measure the width of the pulses. It measures the time between two rising edges, at least according to the documentation that was in place before. Like so:

images

So *not* the width of the pulses.

components/sensor/pulse_meter.rst
dirrgang: "rising edges only are detected", I'd rephrase that as "only the rising edges are detected, and the pulse width is calculated as the time between successive edges.

components/sensor/pulse_meter.rst
dirrgang: It depends on the meter, the two modern ones I've seen in Germany, which admittedly were 3 phase meters, had 10000/kWh. 

Browsing through image results of smart meters has given results such as 400 and 600 pulses also so it's fairly arbitrary. I went with what I had because I wanted to avoid an accidental miscalculation or the like.

components/sensor/pulse_meter.rst
dirrgang: I think the note was removed by accident when replacing the method. I'll re-add it.

components/sensor/pulse_meter.rst
clydebarrow: It depends on the mode. In PULSE mode it does measure the width of the positive pulse. In EDGE mode it measures the period of the pulse train. So you could say "measures the width or period of pulses on a pin". My suggestion later on the mode description clarifies this.

components/sensor/pulse_meter.rst
clydebarrow: I don't see the difference but whatever. BTW "one of EDGE and PULSE" should be "one of EDGE or PULSE".

components/sensor/pulse_meter.rst
clydebarrow: Resolution does not guarantee accuracy. A clock might display time to within 1 second but be wrong by minutes. In the case of this component the timing of edges is measured within an interrupt by reading the microsecond timer. The delay between an edge happening and the interrupt being serviced is variable, and could easily run into several microseconds.

Using a hardware compare/capture module might guarantee microsecond accuracy, but that's not how this component is implemented.

I would just avoid making statements that are opinions and/or not verifiable. Without testing I really don't know how well this meter would do at very high pulse rates - whereas at rates typical for your application I agree it should do very well.

IMHO it's probably sufficient to simply say "This leads to a higher resolution, especially for low pulse rates, as the pulse counter sensor is limited by the number of pulses within a time interval." and leave it at that.

components/sensor/pulse_meter.rst
```diff
@@ -17,66 +23,65 @@ It measures the time between rising edges on a pin, for each pulse it outputs th
...
 
 - **internal_filter_mode** (*Optional*, string): Determines how the internal filter is applied.
-  One of ``EDGE`` and ``PULSE``. Defaults to ``EDGE``. In ``EDGE`` mode subsequent rising edges are compared and if they fall into an interval lesser than the internal filter value, the last one is discarded. In ``PULSE`` mode the rising edge is discarded if any further interrupts are detected before the internal_filter time has passed. In other words, a high pulse must be at least internal_filter long to be counted. This is useful if you are detecting long pulses that may bounces before and/or after the main pulse.  
-
-- **timeout** (*Optional*, :ref:`config-time`): If we don't see a pulse for this length of time, we assume 0 pulses/sec. Defaults to ``5 min``.
+  One of ``EDGE`` or ``PULSE``. Defaults to ``EDGE``.
+  - In ``EDGE`` mode, only rising edges are detected and pulse width is calculated as the time between successive rising edges.  
+  - In ``PULSE`` mode, both rising and falling edges are detected and pulse width is calculated as the time between a rising and a falling edge.
```
TrentHouliston: This is not how it works, the time measured by the pulse meter is always the time between successive rising edges.
The difference between the two filter modes is how it calculates what counts as a "rising edge"

In `EDGE` mode, the next rising edge must be at least `internal_filter` after the previous one. So a rising edge that happens earlier than this is just ignored. It works well when you have a clean signal and is good for higher rates as you don't need to process both edges (and the actual filter is simpler).

The `PULSE` mode requires that after an edge, the value must stay in that state for at least the `internal_filter` time before that edge counts. So if you have lots of noisy bounces it won't count them as rising or falling edges until after it's stayed in that state for long enough. So if you had say a 100ms filter time in pulse mode, your signal would need to be high for at least 100ms without changing, then low for at least 100ms without changing and then on the next rising edge, that pulse width from rising edge to rising edge would be counted.
It's good if you have lots of noisy bounces in your data as it'll ignore them better than the edge mode.

TrentHouliston (commented):
The description of the pulse filter mode isn't correct.

As for pulse rates, in my usage, they are much much higher than even 10000/kWh.
For me 1 pulse represents `5.74745653e-7kWh`
I use it on a shelly 1pm which has a hlw8012 chip in it but the hlw8012 integration in esphome is terrible because it uses pulse_counter in the backend.
So I either have the choice of getting an update once every 60 seconds, or lowering that and getting random 0s and weird values as the pulse meter triggers with no pulses counted.

components/sensor/pulse_meter.rst
TrentHouliston: It 100% does not measure the width of the pulses, it always measures rising edge to rising edge

components/sensor/pulse_meter.rst
dirrgang: My mistake, I guess I misread the code. Thanks for the clarification.

nagyrobi (approved):

PR #3187, opened 2023-09-15
Increase body width to something more reasonable.

clydebarrow:
## Description:

The width of the body section is currently fixed to 800px. This is unreasonably low when used on a desktop monitor.

This PR increases it. It should not affect mobile devices, AFAIK.

## Checklist:


nagyrobi:
Not really nice, as images are defined as width percentages, they become huge now.

clydebarrow:
> Not really nice, as images are defined as width percentages, they become huge now.

I added a `max-width` for images generally (had already done that for the front page.)

nagyrobi (dismissed):

nagyrobi:
@jesserockz my approval was just a test.

jesserockz:
The homepage/tables looks very weird and way to spaced out on just a regular laptop monitor now.

image

clydebarrow:
> The homepage/tables looks very weird and way to spaced out on just a regular laptop monitor now.
> 

It's in the eye of the beholder I guess. That looks just right to me, whereas the standard version looks silly - I paid a lot of money for that screen space, I want to use it!

Screenshot 2023-11-24 at 9 10 18 am

I don't really care about this PR since I have my own customised version of the docs deployed in Google's Sydney data center, so not only does it look the way I want but it's much faster than the Netlify version, for me at least.

So feel free to close this if you don't see an improvement.

nagyrobi:
The index page is indeed silly...

Well I guess some dynamic table/column generation would be useful here. Set the number of table columns based on the screen width.

Or we could simply drop the table completely, just use rectangular fixed width boxes coming one after another horizontally, which could automatically wrap to the width of the page.

PR #3249, opened 2023-10-09
Clarifications to pulse_width documentation

### LINKED ISSUE #4638: The pulse_width sensor is not properly documented
### The problem

I'd like to measure how long is my pump enabled. So I setup pulse width sensor to measure it.

My expectation is that I'll get time of running of the pump, but I got some weird values of unknown time unit.

So is it measuring time in seconds? If true, then it doesn't work properly. After more than minute of run I got value of 0.013.

2023.6.3

pip

not related

ESP8266

### Board

ESP-01

pulse_width

### Example YAML snippet

```yaml
- platform: pulse_width
    pin:
      number: GPIO1
      inverted: True
    name: Doba běhu čerpadla
    id: cerpadlo_pulse_width
    on_value:
      then:
        - if:
            condition:
              lambda: 'return x > 20.0;'
            then:
              - globals.set:
                  id: global_pulsewidth_watertotal
                  value: !lambda 'return id(global_pulsewidth_watertotal) + (x * .4);'
```

```txt
nope
```

### Additional information

_No response_

RoboMagus:
## Description:

Some additional clarification to the usecases for the pulse_width sensor.

**Related issue (if applicable):** fixes esphome/issues#4638

## Checklist:

    or

### DIFF
--- a/components/sensor/pulse_width.rst
+++ b/components/sensor/pulse_width.rst
@@ -7,7 +7,12 @@ Pulse Width Sensor
 
 The ``pulse_width`` sensor allows you to measure how long a given digital signal
 is HIGH. For example this can be used to measure PWM signals to transmit some
-value over a simple protocol.
+value over a simple protocol. The unit of measurement for this sensor is seconds.
+
+.. note::
+
+    This component is intended for measurements in the microsecond to seconds range! 
+    The largest period this component can measure is just over 70 minutes.
 
 .. code-block:: yaml
 


components/sensor/pulse_width.rst
```diff
@@ -7,7 +7,12 @@ Pulse Width Sensor
 
 The ``pulse_width`` sensor allows you to measure how long a given digital signal
```
konikvranik, line 8: Could you add mode detailed information here, please? My usecase is exactly what is described here, but it doesn't work.

So, I want to measure water consumption by multiplying of the pump runtime by the flow. Pump runtime is the length of the signal HIGH. But this reported very unrelated data. Mostly not reacting on the pulse, otherwise reporting super high value.

konikvranik (changes requested):

konikvranik (changes requested):
Could you add mode detailed information here, please? My usecase is exactly what is described here, but it doesn't work.

So, I want to measure water consumption by multiplying of the pump runtime by the flow. Pump runtime is the length of the signal HIGH. But this reported very unrelated data. Mostly not reacting on the pulse, otherwise reporting super high value.

components/sensor/pulse_width.rst
RoboMagus: Please keep your specific use-case to the issue that you've opened for it.
This is a PR to add generic clarification and missing information to the docs, not to discuss why it does not work for your usecase.

nagyrobi (approved):

nagyrobi:
@konikvranik please head on to our Discord general support channel and ask about your specific use case there.

konikvranik:
Hi @nagyrobi, I already solved my problem different way, but I'm affairs that the documentation is incomplete here. That's why this ticket.

PR #3304, opened 2023-10-29
feat: add documentation for AS5600

ammmze:
The AS5600 is a magnetic rotary position encoder with a 12-bit resolution, allowing for 4096 positions/steps. There's been an open feature request for this for a couple years and a custom component implementation using a 3rd party library. This component uses the esphome native i2c implementation, not the 3rd party library.

## Description:
Related issue (if applicable): resolves esphome/feature-requests#1459

Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable): /esphome/esphome#5174

## Checklist:
[x] I am merging into `next` because this is new documentation that has a matching pull-request in [esphome](https://github.com/esphome/esphome) as linked above.

[x] Link added in `/index.rst` when creating new documents for new components or cookbook.


components/sensor/as5600.rst
```diff
@@ -0,0 +1,307 @@
...
+  Used in combination with the **dir_pin**.
+
+  - ``clockwise`` (default)
+  - ``counterclockwise``
+- **start_position** (*Optional*, int): The raw position that should be considered the start (i.e. ``0``).
+  See :ref:`Position / Range <as5600_position_range>`.
+- **end_position** (*Optional*, int): The raw position that should be considered the end (e.g. 180deg)
+  of the allowable rotation range. Mutually exclusive with **range**. See :ref:`Position / Range <as5600_position_range>`.
+- **range** (*Optional*, int): The allowable rotation range from the **start_position**. Mutually
+  exclusive with **range**. See :ref:`Position / Range <as5600_position_range>`.
```
jesserockz:
```suggestion
  exclusive with **end_position**. See :ref:`Position / Range <as5600_position_range>`.
```

components/sensor/as5600.rst
```diff
@@ -0,0 +1,307 @@
...
+            name: Magnitude
+        status:
+            name: Status
+
+.. _as5600-sensor-config:
+
+Configuration variables:
+************************
+
+- **name** (**Required**, string): The name for this sensor.
```
jesserockz: Please don't duplicate the config lines that exist in the base `sensor` config
```suggestion
```

components/sensor/as5600.rst
```diff
@@ -0,0 +1,307 @@
...
+- **out_of_range_mode** (*Optional*, string): How to treat out of range values. Only applicable if configured
+  for a range less than 360 degrees. Defaults to ``min_max``. See :ref:`Out of Range Mode <as5600-out-of-range-mode>`.
+- **as5600_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the
+   :ref:`AS5600 Hub <as5600-component>` you want to use this sensor.
+- **update_interval** (*Optional*, :ref:`config-time`): The interval
+   to check the sensor. Defaults to ``60s``.
+- **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
```
jesserockz: Same as above

components/sensor/as5600.rst
```diff
@@ -0,0 +1,307 @@
...
+- **raw_position** (*Optional*): Reports the raw (un-scaled, un-filtered) position of the magnet.
+
+  - **name** (**Required**, string): The name for the raw position sensor.
+  - **id** (*Optional*, :ref:`config-id`): Set the ID of this sensor for use in lambdas.
+  - All other options from :ref:`Sensor <config-sensor>`.
```
jesserockz: Same as above and please apply to the below.
```suggestion
- **raw_position** (*Optional*): Reports the raw (un-scaled, un-filtered) position of the magnet.
  All options from :ref:`Sensor <config-sensor>`.
```

components/sensor/as5600.rst
```diff
@@ -0,0 +1,307 @@
...
+          - lambda: 'return x * as5600::RAW_TO_DEGREES * id(my_as5600).get_range_scale();'
+      - platform: as5600
+        update_interval: 1s
+        name: Percent
+        unit_of_measurement: '%'
+        accuracy_decimals: 2
+        icon: mdi:rotate-right
+        filters:
+          - delta: 1
+          - lambda: 'return (x / as5600::POSITION_COUNT) * 100;'
```
jesserockz: Instead of suggesting more sensors which will increase the communications to the chip, it would be better to define a regular position sensor, then use `copy` sensors with the filters to adjust. This is more efficient as it updates instantly when the source sensor is updated.

jesserockz (changes requested):

components/sensor/as5600.rst
ammmze: Awesome! Didn't know that was a thing. You are referring to [this](https://esphome.io/components/copy.html#copy-sensor), correct?

jesserockz: Correct

components/sensor/as5600.rst
ammmze: Fixed! Thank you for spotting that typo!

components/sensor/as5600.rst
ammmze: So while I think the copy sensor is valuable to know. I'm glad to know it exists now, but I think what I am going to do here is actually just split the 2 sensors to 2 separate configuration samples. The intent is really just to show how to convert the position value to an angle or percentage. I would imagine in many cases people would be using one or the other. And if I add extra fluff that is really only needed because I tried to combine 2 different examples would ultimately be bad because I assume a good majority of the time people that may not have a deeper understanding end up copying the code blocks into their config and hope it works. Then I imagine people would just remove the sensor they don't want, and then they have extra sensors that they don't technically need.

jesserockz (approved):

PR #3311, opened 2023-10-30
GCJA5/PMSx003- Add link to esphome-compatible hardware

gcormier:
To follow up on the previous work I did on implementing Panasonic GCJA5 support, I'm adding a link to compatible hardware for those who are looking to use the sensor.

I've also added a link for the PMSx003 since the hardware can support that as well.


nagyrobi:
Instead of having these buried in the text, I'd suggest to move all of them at the bottom in the See Also section. Easier to find.

gcormier:
> Instead of having these buried in the text, I'd suggest to move all of them at the bottom in the See Also section. Easier to find.

I moved the hardware - do you want Datasheet in See Also as well?

nagyrobi:
> > Instead of having these buried in the text, I'd suggest to move all of them at the bottom in the See Also section. Easier to find.
> 
> I moved the hardware - do you want Datasheet in See Also as well?

I meant the datasheets, not the tindie products - as this site is no place for advertisments.

gcormier:
Sorry this fell between the cracks.

1. I've moved datasheets to the bottom. 
1. I've changed the links to GitHub FOSS hardware.

PR #3330, opened 2023-11-05
Update lcd_gpio.svg

RoganDawes:
## Description:
Correct labeling of D0-D3 as 4-bit, when the correct pins to use for 4-bit data are D4-D7, as per https://github.com/duinoWitchery/hd44780/blob/master/examples/ioClass/hd44780_pinIO/HelloWorld/HelloWorld.ino#L24-L27

## Checklist:

    or


nagyrobi:
I think we should remove the `D0 / ` ... `D7 / ` names from the display pin labels, as it can lead to confusion, as some ESP modules also name these them with `D `. `Data X (...` should be enough.

Moreover, I'd also remove D from the example too and change it to something like:
```
  data_pins:
    - GPIO32
    - GPIO33
    - GPIO5
    - GPIO17
```

RoganDawes:
I think from the context it is clear that the D0-D7 labels are applicable to the display, not to the microcontroller, since there are a bunch of other labels that the ESP32 is very unlikely to have (contrast, etc).

I can change the pins in the example to match what I have on my ESP32 if it helps. I always like to be sure that what is in the example actually works in the real world within the limitations of the microcontroller (input only GPIO, no pullup/down resistors, etc), although it does then ideally suggest that the microcontroller that was used for the example also be mentioned.

nagyrobi:
> I can change the pins in the example to match what I have on my ESP32 if it helps.

Yes. My GPIO config example above is working with a WT32-ETH01 board...

nagyrobi (approved):

PR #3337, opened 2023-11-06
Tidying up LD2420 doc

kbx81:
## Description:

Fixing a few details I missed during review.

## Checklist:

    or


nagyrobi (dismissed):

nagyrobi:
_Operating Mode_ is a bit confusing. How do you set it?

kbx81:
> _Operating Mode_ is a bit confusing. How do you set it?

That's the `select` component which selects the operating mode of the sensor -- it's detailed in part of the document. However, that's not something this PR is addressing; it's just making the example consistent with our other docs and adding in a few links to component details.

nagyrobi:
I guess what's missing from making it clear is the "All other options from _select_" link, which appears for all the other components, except this one.

kbx81:
Oh my gosh -- I'm sorry. I misunderstood your first comment...and totally overlooked that one. Good catch! I've pushed an update...thank you! 😄

components/sensor/ld2420.rst
```diff
@@ -189,6 +180,7 @@ Configuration variables:
 ************************
 
 - **operating_mode** (*Required*): Provisions the operating mode select component.
+- All other options from :ref:`Select <config-select>`.
```
jesserockz: These are part of the keys like `operating_mode`, not extra config items alongside it
```suggestion
  All other options from :ref:`Select <config-select>`.
```

jesserockz (changes requested):

components/sensor/ld2420.rst
jesserockz: The same applies to all the rest too, each sub-number etc should have this text in each one.

kbx81: Wow, yeah, thank you...wasn't thinking about it in the right way. All fixed now, though. 😄

jesserockz (approved):

PR #3353, opened 2023-11-10
Quick LD2420 Edit 

descipher:
## Description:

One YAML rename was missed.

## Checklist:

It's not in current yet but will be soon.


descipher:
Arrgg needed to target next

PR #3378, opened 2023-11-15
Update Lambda Magic Cookbook with text-to-float conversion example

nagyrobi:
## Description:

## Checklist:

    or


jesserockz:
I don't understand the upside to this suggestion?

If you want a box to enter a number into just set `mode: box` on a `number` component

nagyrobi:
Yes, the straightforward way would be to use a number component, however, when I need to set very strange values like 0.00034356 for a PID controller parameter testing, which needs to be copy/pasted from elsewhere, I don't always know in advance the precision of the value - which is a requirement of the number.

And HA ui tends to display rounds/digits differently

jesserockz (approved):

PR #3413, opened 2023-11-25
Add HE60R Garage Door Opener

clydebarrow:
## Description:

Add HE60R Garage Door Opener

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#5834

## Checklist:


components/cover/he60r.rst
```diff
@@ -0,0 +1,85 @@
...
+
+    uart:
+      - id: door_uart
+        baud_rate: 1200
+        parity: even
+        stop_bits: 1
+        tx_pin:
+          number: 23
+          inverted: true
+        rx_pin: 22
```
jesserockz: Components should not have external config examples. That is for the user to read about and implement. Otherwise they potentially do not understand why the example they literally copied doesnt work.

For uart etc we usually have a line above the example with a reference to uart docs saying it needs to be set up.

components/cover/he60r.rst
```diff
@@ -0,0 +1,85 @@
...
+          mode:
+            input: true
+            pullup: true
+        name: "Garage Door Button"
+        filters:
+          - delayed_on: 50ms
+          - delayed_off: 50ms
+        on_press:
+          then:
+            - cover.toggle: garage_door
```
jesserockz: Same with this. 

Full examples are allowed at the bottom after all of the main documentation, but should really not include pin numbers or use `REPLACEME` as a placeholder which ESPHome has a specific validation check for.

components/cover/he60r.rst
```diff
@@ -0,0 +1,85 @@
...
+
+
+.. figure:: images/he60r-sch.jpg
+    :target: /_images/he60r-sch.jpg
+
+
+Configuration variables:
+------------------------
+
+- **name** (**Required**, string): The name of the cover.
```
jesserockz: This is part of `All other options from Cover` so should not be included here. (We still need to update existing documentation on this stuff).
```suggestion
```

jesserockz (changes requested):

jesserockz (approved):

PR #3456, opened 2023-12-14
UART changing at runtime

edwardtfn:
## Description:

Updates UART documentation based on the changes implemented by esphome/esphome#5909 with the following description:

This adds the method `load_settings()` to the UART component on ESP32 (`arduino` and `esp-idf`) allowing the interface to reload with the new settings used by methods `set_baud_rate`, `set_parity`, `set_data_bits`, etc. without the assignment of a new software UART.

This can be useful for connecting to devices which supports multiple baud rates or for devices which changes it's serial communication at runtime, but also allows some logic for a user's configurable baud rate to be set at `on_boot` after the UART component is loaded.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#5909

## Checklist:

    or

### DIFF
--- a/components/uart.rst
+++ b/components/uart.rst
@@ -180,6 +180,84 @@ In that case, either make sure that the debugger outputs less data per log line 
 ``after.bytes`` option to a lower value) or increase the logger buffer size using the logger
 ``tx_buffer_size`` option.
 
+.. _uart-runtime_change:
+
+Changing at runtime
+-------------------
+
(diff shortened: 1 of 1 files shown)


nagyrobi:
Thank you!
Maybe just a hint on how to convert in a lambda the strings (coming from a select in your example) to `size_t` `uint8_t` `UARTParityOptions` required by the other settings as this is not very common for non C++ experts...

nagyrobi (approved):

PR #3460, opened 2023-12-15
Update pid.rst - Added information about the derivate_average_samples

erapade:
## Description:
I realized that derivate_average_samples only reflects the internal PID component and not the derivate value the PID sends out. This was confusing until I understood how it works so I updated the documentation

## Checklist:

    or


erapade:
This was not correct

PR #3467, opened 2023-12-18
Revert "UART changing at runtime"

nagyrobi:
Reverts esphome/esphome-docs#3456


nagyrobi:
Reverting because it went in the wrong branch (next instead of beta).

PR #3510, opened 2024-01-02
Initial LVGL doc for ESPHome

nagyrobi:
## Description:

github://clydebarrow/esphome@lvgl

## Checklist:

    or


nielsnl68:
a small suggestion, would it be better to move all the widgets to separate files?

components/binary_sensor/lvgl.rst
```diff
@@ -0,0 +1,43 @@
...
+- **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
+- **name** (**Required**, string): The name of the binary sensor.
+- **widget** (**Required**): The ID of a button widget configured in LVGL.
+- All other options from :ref:`Binary Sensor <config-binary_sensor>`.
+
+
+Example:
+
+.. code-block:: yaml
+
```
nielsnl68, line 31: is it an idea to add the **lvgl** widget's that can be used with the documented sensor here as well?

nagyrobi:
> a small suggestion, would it be better to move all the widgets to separate files?

Harder to maintain as separate files. A widget's options list is not that long.

components/binary_sensor/lvgl.rst
nagyrobi: As a link should be enough.

nielsnl68: Yes that could do as well indeed.

nielsnl68:
> Harder to maintain as separate files. A widget's options list is not that long.

Maybe it is harder to maintain but for the users it is easier to read imho. And for them we make the documentation is it not?

nagyrobi:
In my experience it's harder to oversee the doc if small parts are all in separate files. The majority of the configuration options are common for most of the widgets, only specific ones need to be treated separately. One can easily jump up and down the doc page with Ctrl+F on the page to see for example which widgets support something specific.

It's enough to split to separate files the sideloaded ESPHome components from this platform, with many crosslinks.

The site-level search is a pita, so at leat having it on a well structured page with a good TOC on the left makes it easy to comprehend.

But anyways, we'll see how big it will be in the end. It's far from ready.

components/binary_sensor/lvgl.rst
nagyrobi: https://github.com/esphome/esphome-docs/pull/3510/commits/543d8f4ee5529d392bd79ba439f150c5deb664c0

nielsnl68:
you're doing well. the documentation is really starting to make sense. 🤗🙇‍♂️

components/images/lvgl_symbols.png
nielsnl68, line 1: - Did you created this image yourself @nagyrobi ? When so would it be an good idea to sort them alphabetical instead of the hex code?

- is this only the available symbols or is this just a selection of the possible symbols?

nagyrobi: Yes, I created it wilt LVGL itself.  This is the list of all the available symbol glyphs in the font that's built-in the LVGL library. Names of these symbols are removed for now so the codepoints only remain for now.

Support for external fonts and symbols is added, with is by far more rich.

nagyrobi:
Closing this, use https://github.com/esphome/esphome-docs/pull/3678 instead

PR #3516, opened 2024-01-03
Add questionmark to default glyphs.

RubenNL:
## Description:

Added the questionmark to the default glyphs, as mentioned in my PR. Also added the `/` to the docs, which was added in https://github.com/esphome/esphome/pull/1691, but wasn't documented.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6053

## Checklist:

    or


RubenNL:
Do not merge, seems like it doesn't like the space in the beginning of the string. Working on it.

jesserockz (approved):

PR #3566, opened 2024-01-23
Nextion - Remove soft reset at start

edwardtfn:
## Description:

This introduces a user configurable attribute to Nextion display which allows skipping the soft reset on displays during the component startup.
I don't know really why this reset was introduced from the beginning, so I've left as a configurable attribute to avoid a breaking change, but on my tests it always work without this command and it makes the startup process a bit faster (my main goal).

This is not a breaking change, as the previous behavior is the default.

**Related issue (if applicable):** solves esphome/feature-requests#2074

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6134

## Checklist:

    or


kbx81 (approved):

edwardtfn:
This is not stale.

PR #3583, opened 2024-01-30
Set SN74HC165's clock_inhibit_pin as optional

joaopms:
## Description:

The use of only 3 pins is mentioned, but the configuration values don't reflect that. Datasheet specifies the pin can be directly connected to ground. The code already expects the pin to not be used: https://github.com/esphome/esphome/blob/ec3162282cd5a2c3455dd7a8669b4f762c481a20/esphome/components/sn74hc165/__init__.py#L35

## Checklist:

    or


joaopms:
_bump_

nagyrobi (dismissed):

nagyrobi (approved):

PR #3598, opened 2024-02-09
Update information about using https for TFT upload

edwardtfn:
## Description:

Update information about using https for TFT upload based on changes implemented by esphome/esphome#6192 but also based on how it was before (yes, some of the info was requiring update even before this PR).
I've also removed instructions about NGINX and that don't belongs to this place, imho.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6192

## Checklist:

    or


edwardtfn:
I believe all the proposed changes are still relevant, so this shouldn’t be stale.

PR #3611, opened 2024-02-18
Update ssd1306.rst

MauroMP1:
Invertir los pines de i2C, scl es D1 y sda es D2

## Description:

## Checklist:

    or

### DIFF
--- a/components/display/ssd1306.rst
+++ b/components/display/ssd1306.rst
@@ -34,8 +34,8 @@ improve reliability.
 
     # Example configuration entry
     i2c:
-      sda: D1
-      scl: D2
+      scl: D1
+      sda: D2
 
     display:
       - platform: ssd1306_i2c


guillempages:
It would be nice if you would write your description in English, since this is the project's language.

Nevertheless, I happen to understand Spanish as well, and, at least according to the documentation above the part you propose to change, the description explicitly mentions "Connect D2 to the SCL and D1 to the SDA", so the documentation is already good as it is now.

M4ur0MP:
Hello I will remember this for the next time.
   (It would be nice if you would write your description in English, since this is the project's language.)


Nevertheless, I happen to understand Spanish as well, and, at least according to the documentation above the part you propose to change, the description explicitly mentions "Connect D2 to the SCL and D1 to the SDA", so the documentation is already good as it is now.

In the moment when i suggest this change, the documentation was D1 to SCL and D2 to SDA, then 4 or 5 days it change, i assumed that this suggestion worked, but i'nt understand well how work git in this cases.
Sorry for the complications and thanks.

M4ur0MP:
Sorry, but I saw the site again and it is inverted again, in the previous answer I based myself on your answer, but what you suggested is the opposite, in my case I tested the LCD with an esp8266 nodemcu which has the i2c, SCL=>D1 , SDA=>D2, in the example documentation this is inverted, which caused complications for me until I realized that they were inverted, even in the esp32 they are not even close, I suppose the example is with the esp8266.
Sorry for the complications and thank you
Greetings

guillempages:
It's inverted w.r.t. the nodeMCU default pinout, as you mention. This has not changed.

What I was referring to is the text:
"Connect D2 to the SCL pin you chose for the [I²C Bus](https://esphome.io/components/i2c#i2c), and connect D1 to the SDA pin." which is in the same page; just above the example configuration entry.

If you follow those instructions, regardless of the "default" I2C pins on the nodeMCU, the example configuration is right. So your change would make this inconsistent.
AFAIK, the esp8266 does not have hardware specific I2C pins, so it actually does not make any difference if you use D! and D2 as the pinout says, as the ESPHome documentation says or you pick any other two pins.

Nevertheless, I agree that if there is a "preferred" standard (i.e. the nodeMCU pinout documents), it would be better to use that. Your change is not enough for that though; please also update the "Connect ..." text above so that it is then kept consistent with the example.

nagyrobi:
I don't think there's such a "preferred" standard. Too many boards with too many proprietary "standards". It's no standard, just more and more confusion.

The preferred way is to use normal GPIOxx naming, as it's board-independent. Then everyone can look up its own GPIOs depending on the board he/she has.

M4ur0MP:
guillempages it's correct how you says, my change was incomplete, sorry, i will try pay more atenttion
I think, seeing the boards most used, make the change how says "nagyrobi", and more because in the code in the most cases i think is used nomenclature GPIO and not number of pin
i2c:
  sda: GPIO5
  scl: GPIO4
Thanks
Greetings

PR #3631, opened 2024-02-23
Redo the homeassistant service call API setup instructions

javawizard:
Add notes to the individual action sections with a link to the setup instructions - that way folks (like my past self) don't miss them when they link directly to an action and then spend an inordinate amount of time trying to figure out why their service call isn't working.

I'm not 100% sure if user-defined services also require this, or if `homeassistant.tag_scanned` requires it, but this is at least an improvement on the status quo.

## Checklist:

    or


nagyrobi:
This has to be cofigured for _any_ kind of service call to work.

javawizard:
> This has to be cofigured for any kind of service call to work.

Perfect, this should be good to go as is then. Thank you!

nagyrobi (commented):
I think it's a bit overwhelming to repeat the same note 3 times within that small amount of text. It's just a single page.

IMHO the explanation under the Actions title should be enough.

clydebarrow (dismissed):
LGTM

javawizard:
> I think it's a bit overwhelming to repeat the same note 3 times within that small amount of text. It's just a single page.

> IMHO the explanation under the Actions title should be enough.

The main thing I'm trying to address is when someone comes in via say a Google search for "esphome call home assistant service" (for which this page is the first result), sees the `homeassistant.service` action in the table of contents, clicks it, and finds everything they need in that section - except for the fact that they need to do that extra step of enabling service calls for that node in HA.

I spent the better part of an hour trying to figure out what I was doing wrong and totally didn't think to look further up the page, and it appears [I'm not the only one to have done that](https://community.home-assistant.io/t/home-assistant-rejecting-a-esphome-homeassistant-service/604085).

So I'll respectfully disagree: I think that as long as there's a ToC link directly to the action, folks will click it and miss the setup note earlier in the page, so I think there's value in having the warning above each individual action.

components/api.rst
```diff
@@ -75,20 +75,28 @@ Configuration variables:
...
-    Before a newly added ESPHome device can interact with the Home Assistant API it needs to be allowed to communicate
-    with it. This setting can be found in the ESPHome integration (NOT in the Add-On) by clicking "CONFIGURE" for
-    that device and enabling the "Allow device to make service calls" option.
-
 Actions
 -------
 
+Before using any of the actions below, you'll need to tell Home Assistant to allow your device to
+make service calls:
+
```
jesserockz, line 83: We can also add a My button to link directly to the ESPHome integration list.

```suggestion

.. raw:: html

    <a href="https://my.home-assistant.io/redirect/integration/?domain=esphome" target="_blank" rel="noreferrer noopener"><img src="https://my.home-assistant.io/badges/integration.svg" alt="Open your Home Assistant instance and show an integration." /></a>


```

jesserockz (dismissed):

javawizard:
Oh hey. Keep this one open please 🤖🤖🤖

components/api.rst
javawizard: Done.

components/api.rst
```diff
@@ -75,20 +75,35 @@ Configuration variables:
...
 .. _api-homeassistant_event_action:
 
 ``homeassistant.event`` Action
 ******************************
 
+.. note::
+
+    Be sure to :ref:`follow the instructions above <api-actions>` to tell Home Assistant to allow
+    your device to make service calls.
+
```
clydebarrow, line 106: I would think this instance of the note could be removed, since the paragraph it's referring to is immediately above, and fairly obvious now that the HA button is there.

javawizard: The problem is when you arrive at that section by clicking on it in the sidebar. You get taken to something like this: https://deploy-preview-3631--esphome.netlify.app/components/api.html?highlight=native+api#homeassistant-event-action

The section with the setup instructions is completely out of your scroll view at that point and you may not think to scroll up and see what you're missing. That's exactly what led to my making this change in the first place: I clicked on a sidebar link, saw the instructions I thought I needed to follow, then spent the next hour trying to figure out why it wasn't working before going back and reading the entire page that I had previously thought was mostly about other native API stuff that I didn't need to know.

So I'll respectfully argue for keeping that one in, even if it is otherwise a bit superfluous.

clydebarrow: Fair enough - not an ideal solution, but I can't think of a better one right now.

clydebarrow (approved):

PR #3633, opened 2024-02-23
output is optional

ssieb:
## Description:
Make the output optional and also fix the name which is optional.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6274

## Checklist:

    or

### DIFF
--- a/components/fan/speed.rst
+++ b/components/fan/speed.rst
@@ -23,9 +23,9 @@ supports speed settings.
 Configuration variables:
 ------------------------
 
-- **output** (**Required**, :ref:`config-id`): The id of the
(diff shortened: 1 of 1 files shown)


nagyrobi (changes requested):
I would also add this information the introductory sentence to make aware people of this beneficial change. For example something like:


```
The ``speed`` fan platform lets you represent any float :ref:`output` as a fan that
supports speed settings, or even sent up a *template* fan using just triggers.
```

jesserockz (approved):
I'm gonna merge this to get it out in a release. If we want to add more, we can make another PR to `current` after the release.

PR #3652, opened 2024-02-29
Update index.rst to provide missing information on the on_value behavior

Zynth-dev:
The on_value behavior was unclear about what happens when the selected-value is set to the same value as it already was.

## Description:

## Checklist:

    or


components/select/index.rst
```diff
@@ -70,7 +70,7 @@ For more information on using lambdas with select, see :ref:`select-lambda_calls
 ``on_value``
 ************
 
-This automation will be triggered when a new value is published. In :ref:`Lambdas <config-lambda>`
+This automation will be triggered when a new value is set/published. Even if the value is the same as before. In :ref:`Lambdas <config-lambda>`
```
guillempages:
```suggestion
This automation will be triggered whenever a value is set/published, even if the value is the same as before. In :ref:`Lambdas <config-lambda>`
```

guillempages (approved):

PR #3675, opened 2024-03-11
Add note about fake RP2040 unsupported

nagyrobi:
## Description:

Stores are overwhelmed by a new `RP2040 Pico W-2023` board, sold with the same name _Raspberry Pi Pico W_ which is unfortunate because of a fundamental difference: the radio module is not CYW43455, but an ESP8285(!). Naturally, ESPHome firmware uploaded to such a board will cause failure.

rp2040

## Checklist:

    or

### DIFF
--- a/components/rp2040.rst
+++ b/components/rp2040.rst
@@ -11,7 +11,7 @@ This component contains platform-specific options for the RP2040 platform.
 
     Support for all aspects of ESPHome on the RP2040 is still in development.
     
-    Only the original model of Raspberry Pi Pico W board is supported, which has the Cypress **CYW43455** chip providing wireless connectivity. It can be identified by a metalic shield encapsulating the radio circuitry. Pico W boards having as radio modules chips like ESP8285 or similar (labelled as ``RP2040 Pico W-2023`` etc.), are not supported.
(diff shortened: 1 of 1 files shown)


components/rp2040.rst
```diff
@@ -10,6 +10,9 @@ This component contains platform-specific options for the RP2040 platform.
 .. note::
 
     Support for all aspects of ESPHome on the RP2040 is still in development.
+    
+    Only the original model of Raspberry Pi Pico W board is supported, which has the Cypress **CYW43455** chip providing wireless connectivity. It can be identified by a metalic shield encapsulating the radio circuitry. Pico W boards having as radio modules chips like ESP8285 or similar (labelled as ``RP2040 Pico W-2023`` etc.), are not supported.
```
clydebarrow:
```suggestion
    Only the original model of Raspberry Pi Pico W board is supported, which has the Cypress **CYW43455** chip providing wireless connectivity. It can be identified by a metallic shield encapsulating the radio circuitry. Pico W boards with radio module chips like ESP8285 or similar (labelled as ``RP2040 Pico W-2023`` etc.), are not supported.
```

jesserockz (approved):

PR #3706, opened 2024-03-22
Add note about customizations in on_boot

TikiBill:
After informative discussions with @kbx81 and @nagyrobi, I learned that the best way to have a custom power-on state is to edit the `on_boot` automation script. This pull request adds a note to that effect to (hopefully) lead other folks down the preferred path.

Thanks!


nagyrobi:
This note is outdated by https://github.com/esphome/esphome/pull/7577

PR #3712, opened 2024-03-26
Update documentation for on_update

excalbian:
## Description:

**NOT READY TO MERGE** - There is still conversation about this change in the related PR and the documentation might need to be updated, based on the outcome over there.

**Related issue (if applicable):** fixes <[link to issue](https://github.com/esphome/issues/issues/5629)>

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6420

## Checklist:

  **NOT DONE** when I targeted these changes to `next` it seems the last few changes in `current` aren't yet in `next` so my PR was bringing in other changes. This is a change for a PR that's being discussed so I think `next` is the correct target for the documentation, if the changes can be resolved.


excalbian:
Note I moved the section on `on_update` to match the other automations. I'm not sure why it was put in the docs the way it was, but after reading it a few times, this seemed like the correct approach. Please feel free to comment if that's not correct.

PR #3757, opened 2024-04-14
Add note about HA allowing service calls

jesserockz:
SSIA

Note: PR created on GH mobile app does not use template 🤦‍♂️


jesserockz:
Just noticed we have a note higher up

PR #3799, opened 2024-04-25
[bedjet] Update docs for esphome/esphome#6633

javawizard:
## Description:

Update docs for https://github.com/esphome/esphome/pull/6633

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6633

## Checklist:

    or

### DIFF
--- a/components/climate/bedjet.rst
+++ b/components/climate/bedjet.rst
@@ -119,6 +119,13 @@ Configuration variables:
 
     Whichever is not selected will be made available as a custom preset.
 
+- **temperature_source** (*Optional*, string): The temperature that should be used as the
+  climate entity's current temperature:
+
+    - ``ambient`` (Default) - The temperature of the room the BedJet is in will be
+      reported as the climate entity's current temperature.
+    - ``outlet`` - The temperature of the air being discharged by the BedJet will be
+      reported as the climate entity's current temperature.
 - All other options from :ref:`Climate <config-climate>`.
 
 ``bedjet`` Fan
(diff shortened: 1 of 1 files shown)


components/climate/bedjet.rst
```diff
@@ -119,6 +119,13 @@ Configuration variables:
 
     Whichever is not selected will be made available as a custom preset.
 
+- **temperature_source** (*Optional*, string): The temperature that should be used as the
+  climate entity's current temperature:
+
+    - ``ambient`` (Default) - The temperature of the room the BedJet is in will be
+      reported as the climate entity's current temperature.
+    - ``outlet`` - The temperature of the air being discharged by the BedJet will be
+      reported as the climate entity's current temperature.
```
jhansche, line 128: I would just point out the note that I added in the [code PR](https://github.com/esphome/esphome/pull/6633#discussion_r1580021718) - as a warning to the user if nothing else.

Copying here for visibility:

> One thing to keep in mind is that this outlet temp will now be included in the average temp of the room where the ESPHome climate entity is - so if you have the BedJet in a bedroom Room, and an ambient air temp sensor in the same Room, a Room card will display the average of those temps, which will be incorrect when the outlet temp source is chosen. It's a pretty minor detail overall, but might be something we want to call out in the documentation.

jhansche (approved):
Looks good!

components/climate/bedjet.rst
javawizard: Responded in https://github.com/esphome/esphome/pull/6633#discussion_r1581637229 - it looks like climate entities' current temperatures are _not_ incorporated into area averages; only temperature sensors are - so I'm inclined to think the documentation is fine as is.

(I could add a warning on the temperature sensor section about that, but then there are plenty of other temperature sensor components that don't carry any such similar warning so I'm inclined to think it's probably not worth it - let me know if you disagree. At any rate, the climate entity won't contribute so it doesn't need to carry such a warning.)

jhansche: Sounds good to me

components/climate/bedjet.rst
```diff
@@ -146,6 +153,38 @@ Configuration variables:
...
+  - **name** (**Required**, string): The name for the outlet temperature sensor.
+  - **id** (*Optional*, string): Set the ID of this sensor for use in lambdas.
+  - All other options from :ref:`Sensor <config-sensor>`.
+
+- **ambient_temperature** (*Optional*): If specified, the temperature of the room the
+  BedJet is in will be reported as a sensor.
+
+  - **name** (**Required**, string): The name for the ambient temperature sensor.
+  - **id** (*Optional*, string): Set the ID of this sensor for use in lambdas.
+  - All other options from :ref:`Sensor <config-sensor>`.
```
jesserockz:
```suggestion
- **outlet_temperature** (*Optional*): If specified, the temperature of the air being
  discharged from the BedJet will be reported as a sensor.
  All options from :ref:`Sensor <config-sensor>`.

- **ambient_temperature** (*Optional*): If specified, the temperature of the room the
  BedJet is in will be reported as a sensor.
  All options from :ref:`Sensor <config-sensor>`.
```

jesserockz (approved):

PR #3800, opened 2024-04-26
Documentation for new OpenTherm component

olegtarasov:
## Description:

This PR adds documentation for the new OpenTherm component in ESPHome.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#6645

## Checklist:


components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
+OpenTherm
+=========
+
+.. seo::
+    :description: Instructions for setting up OpenTherm bridge in ESPHome.
+    :image: ../components/images/opentherm-shield.png
+    :keywords: OpenTherm
+
+OpenTherm (OT) is a standard communications protocol used in central heating systems for the communication between a central heating appliances and a thermostatic controller.
```
kbx81:
```suggestion
OpenTherm (OT) is a standard communications protocol used in central heating systems for the communication between central heating appliances and a thermostatic controller.
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+
+.. figure:: images/opentherm-shield.png
+    :align: center
+    :width: 50.0%
+
+    DIYLESS Master OpenTherm Shield.
+
+.. note::
+
+    This component acts only as an OpenTherm master (i.e. a thermostat or controller) and not as a slave or gateway. You can no longer use your existing thermostat if you control your boiler through ESPHome with this component.
```
kbx81:
```suggestion
    This component acts only as an OpenTherm master (for example, a thermostat or controller) and not as a slave or gateway. Your existing thermostat is not usable while you use ESPHome with this component to control your boiler.
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+Quick glossary
+--------------
+
+- CH: Central Heating
+- DHW: Domestic Hot Water
+
+Hub
+---
+
+First of all you need to define the OpenTherm hub in your configuration. Note that most OpenTherm adapters label ``in`` and ``out`` pins relative to themselves, and this component labels its ``in`` and ``out`` pins relative to ESP board. So usually your bridge's ``in`` pin becomes hub's ``out`` pin and vice versa.
```
kbx81:
```suggestion
First, you need to define the OpenTherm hub in your configuration. Note that most OpenTherm adapters label ``in`` and ``out`` pins relative to themselves; this component labels its ``in`` and ``out`` pins relative to the microcontroller ESPHome runs on. As such, your bridge's ``in`` pin becomes the hub's ``out`` pin and vice-versa.
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+- **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.  Required if you have multiple busses.
+
+Note abut sync mode
+*******************
+
+By default this component adheres to ESPHome recommendations and doesn't block the loop while communicating with the boiler. Unfortunately, some other components (like Dallas temperature sensors) don't play well with this behaviour, and may cause a lot of lost frames and protocol warnings from OpenTherm. Since OpenTherm is resilient by design and transmits its messages in a constant loop, these dropped frames don't really matter. But if you want to decrease the number of protocol warnings in your logs, you can enable ``sync_mode``, which blocks ESPHome loop until a single conversation with the boiler is complete. This can greatly reduce the number of dropped frames, but usually doesn't eliminate them entirely. You also need to add a logging directive to your config, so that ESPHome doesn't complain about a single component hogging the loop for too long:
```
kbx81:
```suggestion
The use of some components (like Dallas temperature sensors) may result in lost frames and protocol warnings from OpenTherm. Since OpenTherm is resilient by design and transmits its messages in a constant loop, these dropped frames don't usually cause any problems. Still, if you want to decrease the number of protocol warnings in your logs, you can enable ``sync_mode`` which will block ESPHome's main application loop until a single conversation with the boiler is complete. This can greatly reduce the number of dropped frames, but usually won't eliminate them entirely. With ``sync_mode`` enabled, in some cases, ESPHome's main application loop may be blocked for longer than is recommended, resulting in warnings in the logs. If this bothers you, you can adjust ESPHome's log level by adding the following to your configuration:
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+        id: setpoint
+
+This is especially useful in combination with the PID Climate component:
+
+.. code-block:: yaml
+
+    climate:
+      - platform: pid
+      heat_output: setpoint
+      # ...
```
kbx81:
```suggestion
        heat_output: setpoint
        # ...
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+
+The last point ensures that central heating is not enabled if no heating is requested as indicated by a setpoint of 0. If you use a number as the setpoint input and use a minimum value higher than 0, you NEED to use the ch_enable switch to turn off your central heating. In that case the flag will be set to true in the hub configuration, and setpoint is always larger than 0, so including a switch is the only way you can turn off central heating. (This also holds for cooling and CH2.)
+
+The following switches are available:
+
+- ``ch_enable``: Central Heating enabled
+- ``dhw_enable``: Domestic Hot Water enabled
+- ``cooling_enable``: Cooling enabled
+- ``otc_active``: Outside temperature compensation active
+- ``ch2_active``: Central Heating 2 active
```
kbx81: The section heading is `switches`, but then the first thing talked about is how to not use the switches. 😅

I think I would reorganize this section like so:

```suggestion

Switches are available to allow manual toggling of any of the following five status codes:

- ``ch_enable``: Central Heating enabled
- ``dhw_enable``: Domestic Hot Water enabled
- ``cooling_enable``: Cooling enabled
- ``otc_active``: Outside temperature compensation active
- ``ch2_active``: Central Heating 2 active

If you do not wish to have switches, the same values can be permanently set in the hub configuration, like so:

.. code-block:: yaml

    opentherm:
      ch_enable: true
      dhw_enable: true

This is useful when you'd never want to toggle it after the initial configuration.

The default values for these configuration variables are listed below.

To enable central heating and cooling, the flag is only sent to the boiler if the following conditions are met:

- the flag is set to true in the hub configuration,
- the switch is on (if configured),
- the setpoint or cooling control value is not 0 (if configured)

For domestic hot water and outside temperature compensation, only the first two conditions are necessary.

The last point ensures that central heating is not enabled if no heating is requested as indicated by a setpoint of 0. If you use a number as the setpoint input and use a minimum value higher than 0, you **must** use the ``ch_enable`` switch to turn off your central heating. In such a case, the flag will be set to true in the hub configuration and the setpoint is always larger than 0, so including a switch is the only way you can turn off central heating. (This also holds for cooling and CH2.)
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+
+    logger:
+
+    api:
+    ota:
+    wifi:
+      ap:
+        ssid: "Thermostat"
+        password: "MySecretThemostat"
+
```
kbx81: These can be omitted, they're not really relevant to the example.

```suggestion
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
```
kbx81: As above
```suggestion
```

kbx81 (changes requested):
Mostly clean-up & copy fixes.

Also, please wrap all lines at no more than 120 characters.

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+Hub
+---
+
+First of all you need to define the OpenTherm hub in your configuration. Note that most OpenTherm adapters label ``in`` and ``out`` pins relative to themselves, and this component labels its ``in`` and ``out`` pins relative to ESP board. So usually your bridge's ``in`` pin becomes hub's ``out`` pin and vice versa.
+
+.. code-block:: yaml
+
+    opentherm:
+      in_pin: 4
+      out_pin: 5
```
kbx81:
```suggestion
      in_pin: GPIOXX
      out_pin: GPIOXX
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
+    api:
+    ota:
+    wifi:
+      ap:
+        ssid: "Thermostat"
+        password: "MySecretThemostat"
+
+    opentherm:
+      in_pin: 21
+      out_pin: 26
```
kbx81:
```suggestion
      in_pin: GPIOXX
      out_pin: GPIOXX
```

components/opentherm.rst
```diff
@@ -0,0 +1,363 @@
...
```
kbx81:
```suggestion
      in_pin: GPIOXX
      out_pin: GPIOXX
```

kbx81 (changes requested):
Couple other things I missed earlier -- for pins in documentation, we use `GPIOXX` to prompt the user to specify their own pins.

olegtarasov:
@kbx81 Thank you for the review! I've applied your suggestions and wrapped most lines at 120 characters.

kbx81 (dismissed):

kbx81 (dismissed):

kbx81 (dismissed):

kbx81 (dismissed):

kbx81 (dismissed):

FreeBear-nc:
You describe values such as t_set as inputs - Surely they are outputs.

```diff --git a/components/opentherm.rst b/components/opentherm.rst
index 819120c9..91945ae3 100644
--- a/components/opentherm.rst
+++ b/components/opentherm.rst
@@ -147,7 +147,7 @@ output and number components by default:
 - ``auto_min_value`` (boolean): Automatically configure the minimum value to a value reported by the boiler. Not
   available for all inputs.
 
-The following inputs are available:
+The following outputs are available:
 
 - ``t_set``: Control setpoint: temperature setpoint for the boiler's supply water (°C)
 
```

olegtarasov:
@FreeBear-nc You are correct, this terminology looks confusing. It's part of the original documentation from Arthur Rump that I didn't change. But I wouldn't call them inputs either because then we can declare an input as an output, and it's confusing :)  I would rewrite this whole section as following:

```
Numerical values
*****************

There are three ways to set a numerical value:

- As an input sensor, defined in the hub configuration:

[skipped] 

  This can be useful if you have an external thermostat-like device that provides the setpoint as a sensor.

- As a number:

[skipped] 

  This is useful if you want full control over your boiler and want to manually set all values.

- As an output:

[ skipped ]

The following numerical values are available:```

components/opentherm.rst
jesserockz, line 1: Please add the standard `See Also` block at the bottom of the page with at least a link to the header files and the ghedit link.

jesserockz (changes requested):

components/opentherm.rst
kbx81: Once this is done, this has my 👍🏻

olegtarasov: Done!

jesserockz (approved):

PR #3818, opened 2024-05-06
Update bme680_bsec licence url

ireun:
## Description:

Previous link leads to 404, this one is correct, taken from https://www.bosch-sensortec.com/software-tools/software/bme680-software-bsec/

## Checklist:


kbx81 (approved):
Thanks!

PR #3893, opened 2024-05-31
Document host platform

### LINKED ISSUE #2388: Desktop/native platform support
I would like to be able to compile a ESPHome configuration to a binary that can be run on a desktop. Obviously, such a configuration could not e.g. refer to any pins, nor any actual hardware. 

This could be a valuable way to:
 - run tests (both manual and CI),
 - more easily move pieces of logic between an ESP device and a more standard computer in a HA-less environment (esp. with #2081 implemented).

Obvious alternative for the latter is actually running HA or writing the on-desktop piece of logic in Python, so that it uses the native API client library. I don't see any obvious alternatives for CI-style testing.

**Additional context**

Note that platformio has a ["native" platform](https://docs.platformio.org/en/latest/platforms/native.html) that could be useful here.

clydebarrow:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2388

## Checklist:

    or


nagyrobi (approved):

index.rst
```diff
@@ -131,6 +131,7 @@ Supported Microcontrollers
     RP2040, components/rp2040, rp2040.svg
     BK72xx, components/libretiny, bk72xx.svg
     RTL87xx, components/libretiny, rtl87xx.svg
+    host, components/host, host.svg
```
jesserockz, line 134:
```suggestion
    host, components/host, host.svg, dark-invert
```

PR #3895, opened 2024-06-01
[remote_receiver] buffer_size note

gabest11:
## Description:

Added a note about ESP8266 buffer_size allocation quirks.

It would be better to do a `new uint32_t[s.buffer_size/4]` [at this place](https://github.com/esphome/esphome/blob/2beb1f033611c0984c3c4d11734e967cc8a1416f/esphome/components/remote_receiver/remote_receiver_esp8266.cpp#L47C27-L47C41) and allocate as much memory as the documentation originally says, but I'm note sure if that would break anything.

## Checklist:

    or


gabest11:
Ahh, I think it included my other changes for cc1101. I forgot that it was not a branch. Please ignore. Or please add the remote_receiver notes.

PR #3924, opened 2024-06-07
Extend the UDP message sender Cookbook with 8266 example

nagyrobi:
## Description:

Extended the configuration example of "Send UDP commands" cookbook entry form the "Lambda Magic" chapter with working yaml for ESP8266. This is important because many think that the WiFi implementation of ESP8266 lacks UDP, but WiFiUdp library comes to help with a simple inclusion (which btw is used by two other components).

## Checklist:

    or


jesserockz (commented):
Can this entire section of the cookbook be removed due to the udp component?

nagyrobi:
> Can this entire section of the cookbook be removed due to the udp component?

This has no direct relation to the UDP component because it targets sending UDP packets to any third party devices, with any arbitrary data inside. The UDP component is for communication between ESPHome nodes, while these examples show how to send UDP commands to anything else.

I use these with great success in controlling a BrightSign player.

PR #3926, opened 2024-06-10
New branding for ESPHome

Madelena:
## Description:
* New Logo
* New favicons
* Changed typography of the website to match the new branding.

## Checklist:

    or


nagyrobi:
New font introduces too big line spacing. Looks a bit fell apart. Would worth reducing line spacing to make it look more consolidated.

nagyrobi:
Thanks! A bit better now. Could we try maybe a `line-height: 1.5;` too to see how it would look like with that?

jesserockz (approved):

jesserockz:
> Thanks! A bit better now. Could we try maybe a `line-height: 1.5;` too to see how it would look like with that?

I just had a quick play with the line-height, and yes I think it should be less too.. 
I'm not a designer though, will let @Madelena decide on that.

jesserockz:
I will merge for today though and we can adjust the line-height separately

PR #3933, opened 2024-06-10
Fix usage of response in http_request.rst samples

glmnet:
## Description:

Simple fixes of http request samples as I'm testing now, the response object is in shared pointer so it needs the arrow operator

`` error: 'class std::shared_ptr<esphome::http_request::HttpContainer>' has no member named 'status_code'``


jesserockz:
Woops, did not see this. Sorry @glmnet. Thanks for trying to update it.

It got merged in #4049

PR #4034, opened 2024-07-10
Improves Mac install instructions

heythisisnate:
## Description:
Better guidance on how to install ESPHome on a Mac

**Related issue (if applicable):** NA

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** NA

## Checklist:

    or

### DIFF
--- a/guides/installing_esphome.rst
+++ b/guides/installing_esphome.rst
@@ -49,14 +49,26 @@ with the following:
 Mac
 ---
 
-There are no tested installation instructions for Mac. ESPHome does support
-Mac & will run with no problem.
+The simplest way to install ESPHome on a Mac is with `Homebrew<https://brew.sh/>`:
 
-Contributions are welcome!
+.. code-block:: console
 
-The process will likely be similar to Windows. You can install Python from the
-official site, and then install ESPHome with ``pip3 install esphome``.  You can
-then test that things are properly installed with the following:
+    $ brew install esphome
+
+Or, if you have a working Python installation you can install it directly without Homebrew.
+There are several ways to install Python on a Mac, including with Homebrew, Apple's X-Code command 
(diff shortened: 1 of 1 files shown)


guides/installing_esphome.rst
```diff
@@ -49,14 +49,26 @@ with the following:
...
 ---
 
-There are no tested installation instructions for Mac. ESPHome does support
-Mac & will run with no problem.
+The simplest way to install ESPHome on a Mac is with `Homebrew <https://brew.sh/>`_:
 
-Contributions are welcome!
+.. code-block:: console
+
+    $ brew install esphome
```
jesserockz, line 56: Brew (and other system package installers) are "unsupported" as they are not maintained by core ESPHome maintainers and this is why we do not mention them in the documentation.

jesserockz (changes requested):

guides/installing_esphome.rst
heythisisnate: hey @jesserockz, I get it that you don't want to take on support. But Homebrew in this case is just a wrapper around the PyPi packages (see the [forumula source](https://github.com/Homebrew/homebrew-core/blob/master/Formula/e/esphome.rb)) which just makes it more convenient for Mac users to install. So you could make the argument that since the _actual code_ installed by homebrew is the exact same PyPi package as you would get from pip, then you are actually maintaining it!

FWIW I've been using the homebrew package for ESPHome for a long time and never had any issue.

But I'll follow your lead and if you don't want Homebrew mentioned at all, what's the next best alternative? Mac has a suitable version of Python that's shipped with Xcode command line tools which an end-user can install with `xcode-select --install`. Once that's installed, the Mac user should have access to pip3 normally. Ok to mention that?

PR #4061, opened 2024-07-17
Added "and/or brightness (e.g. monochromatic)" to the Random Effect description

GizmoGoody:
The Random Effect can also affect the brightness instead of color (e.g. monochromatic light components). I'm not sure if the brightness is also affected for RGB light components, so I put it as an "and/or" instead of just an "or." Of course, adjust based on how it actually works with RGB lights.

## Description:

## Checklist:

    or

### DIFF
--- a/components/light/index.rst
+++ b/components/light/index.rst
@@ -514,7 +514,7 @@ Configuration variables:
 Random Effect
 *************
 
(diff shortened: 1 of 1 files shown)


GizmoGoody (commented):
@frenck I don't know why this simple edit was allowed to be closed out by the bots, but it's still valid.

nagyrobi (approved):

PR #4065, opened 2024-07-18
Explicitly specify as lambda

### LINKED ISSUE #5394: data_template for homeassistant.event action must be explicitly defined as a lambda
### The problem

When using `data_template` and accompanying `variables` tag in the `homeassistant.event` action, the lambda code for the variables must be explicitly declared as a lambda (using `!lambda`). The provided example in the [api documentation](https://esphome.io/components/api#homeassistant-event-action) (which is for `homeassistant.service`) does not do this. 

In the yaml snippet, `brightness_inp` is defined in the explicit way, whereas `range_inp` is defined as shown in the documentation. In the additional information, `brightness` is set to the input value integer, whereas `range` is the lambda code as defined in the ESP config.

I'm not sure if this is a bug or an error in the documentation, however.

2023.12.6

pip

2024.1.2

ESP32

### Board

az-delivery-devkit-v4

api

### Example YAML snippet

```yaml
- homeassistant.event:
          event: esphome.led_panel_range_set
          data_template:
            brightness: '{{ brightness_inp }}'
            range: '{{ range_inp }}'
          variables:
            brightness_inp: !lambda 'return brightness_val;'
            range_inp: |-
              return id(led_strip_rgb_range).state;
```

_No response_

### Additional information

The event data is returned in home assistant as:
```
event_type: esphome.led_panel_range_set
data:
  device_id: xxx
  brightness: 100
  range: return id(led_strip_rgb_range).state;
origin: LOCAL
time_fired: "xxx"
context:
  id: xxx
  parent_id: null
  user_id: null
```

mjsir911:
Fixes esphome/issues#5394

## Description:

## Checklist:

    or

### DIFF
--- a/components/api.rst
+++ b/components/api.rst
@@ -157,8 +157,7 @@ calls straight from ESPHome :ref:`Automations <automation>`.
           data_template:
             message: The humidity is {{ my_variable }}%.
           variables:
-            my_variable: |-
-              return id(my_sensor).state;
+            my_variable: !lambda 'return id(my_sensor).state;'
 
 Configuration variables:
 ````````````````````````


jesserockz (commented):
According to the schema in the codebase, this should actually be valid. But something got broken along the way. I think I will fix the code instead of changing the docs.

PR #4075, opened 2024-07-22
Correct ENS160 default address

jadegraaf:
Default address was changed in [#6369](https://github.com/esphome/esphome/blob/f1aa254e4810999d7bb70619ba9c09bb9be5cf98/esphome/components/ens160_i2c/sensor.py#L16) and documentation was not updated.

## Checklist:

    or


jesserockz (approved):
Looks like I missed seeing that the default was being changed (which should not have happened)

PR #4080, opened 2024-07-25
RMT documentation clarification and refactoring

nagyrobi:
## Description:

- `repeat` options applicable to all the actions is templatable, `wait_time` is in µs (not the usual ms!) when lambda is used
- declutter the TOC by removing the undistinguishable actions (similar to Ethernet, ESP32 Camera, IR Remote Climate docs)
- change `switch` in examples to `button` as it's more appropriate for this kind of usage (at the time this doc was originally created `button` didn't exist yet)
- move the **Setting up** guides to the guides section of the site
- added the missing "- All other options from Remote Transmitter Actions" to the ones not having it
- remove the IRRemoteESP8266 seealso link as it's not currently supported in any way
- comment to `buffer_size: 2kb` in the example that this should only be used on ESP8266 (ESP32 defaults to 10kb, this would be counterproductive).
- remote_receiver.rst is affected in correction of the hyperlinks pointing to items in remote_transmitter.rst
- fixed See Also links (also in the RF Bridge page)
- Remote Receiver link appeared twice in the index (was left mistakenly in the misc section)

## Checklist:

    or


components/remote_transmitter.rst
```diff
@@ -48,6 +48,189 @@ Configuration variables:
 -  **id** (*Optional*, :ref:`config-id`): Manually specify
    the ID used for code generation. Use this if you have multiple remote transmitters.
 
+.. _remote-setting-up-infrared:
```
jesserockz: I feel like these blocks should be on a page in the `guides` folder and just linked from here.

jesserockz (changes requested):
Generally looks good. Just one comment about the guide blocks

components/remote_transmitter.rst
nagyrobi: Done.

jesserockz (approved):

PR #4081, opened 2024-07-25
Changed Default address to 0x76

3ative:
Changed Default address to 0x76. Plus added note to change address by pulling SDO Pin High.

## Description:

## Checklist:

    or

### DIFF
--- a/components/sensor/bmp280.rst
+++ b/components/sensor/bmp280.rst
@@ -54,7 +54,7 @@ Configuration variables:
   - All other options from :ref:`Sensor <config-sensor>`.
 
 - **address** (*Optional*, int): Manually specify the I²C address of
-  the sensor. Defaults to ``0x77``. Another address can be ``0x76``.
+  the sensor. Defaults to ``0x76``. Another address can be ``0x77`` when SDO pin is pulled High.
 - **iir_filter** (*Optional*): Set up an Infinite Impulse Response filter to increase accuracy. One of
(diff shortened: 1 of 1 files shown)


nagyrobi:
Afaik the default address is still 0x77: 
https://github.com/esphome/esphome/blob/39c0019534f5e6aa67d2e71f69a04d8c3ff28343/esphome/components/bmp280/sensor.py#L77

3ative:
> Afaik the default address is still 0x77: https://github.com/esphome/esphome/blob/39c0019534f5e6aa67d2e71f69a04d8c3ff28343/esphome/components/bmp280/sensor.py#L77

That may well be the case - Fact is every BMP/E board (without mod) I have is 'found' as 0x76

3ative (commented):
"(corresponding to SDO pin pulled low)" is not necessary as it's pulled low internally anyway.

jesserockz:
Thanks for the pull request but documentation states the defaults that are present in the ESPHome codebase at the time the code was merged in. We are not going to change the default as this would be a breaking change for an unknown amount of gain.

Jesse

3ative:
> Thanks for the pull request but documentation states the defaults that are present in the ESPHome codebase at the time the code was merged in. We are not going to change the default as this would be a breaking change for an unknown amount of gain.
> 
> Jesse

I appreciate the response dude. Others have also mentioned the same. Like I also mentioned: Every board I've let the code scan for defaults to 0x76, hence the reason for the PR...
- Maybe I have a batch of factory mod-ed boards then. LOL

PR #4095, opened 2024-07-30
Break out packages + substitutions docs

kbx81:
## Description:

Breaks out package & substations docs into their own pages.

## Checklist:

    or


components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+        refresh: 1d # optional
+
+      # A single file can be expressed using `file` or `files` as a string
+      remote_package_two:
+        url: https://github.com/esphome/non-existant-repo
+        file: file1.yml # cannot be combined with `files`
+        # files: file1.yml
+
+      # shorthand form github://username/repository/[folder/]file-path.yml[@branch-or-tag]
+      remote_package_three: github://esphome/non-existant-repo/file1.yml@main
```
jesserockz: Lets flip this around so simpler comes first

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+      remote_package:
+        url: https://github.com/esphome/non-existant-repo
+        ref: main # optional
+        files: [file1.yml, file2.yml]
+        refresh: 1d # optional
+
+      # A single file can be expressed using `file` or `files` as a string
+      remote_package_two:
+        url: https://github.com/esphome/non-existant-repo
+        file: file1.yml # cannot be combined with `files`
```
jesserockz: I regret adding `file` **and** `files` haha
I think we should undocument `file` and just use `files` everywhere

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+.. seo::
+    :description: How to use packages in ESPHome
+    :image: settings.svg
+
+When you have many ESPHome devices (or are producing and distributing them at scale), a common need tends to surface:
+configuration modularization. You'll likely want to break your configuration into common (groups of) elements, building
+it into reusable pieces which can subsequently be used by many/all devices. Only unique pieces of your configuration
+remain in any given device's YAML configuration file.
+
+This can be accomplished with ESPHome's "packages" feature.
```
jesserockz:
```suggestion
This can be accomplished with ESPHome's ``packages`` feature.
```

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+          - switch.turn_on: open_${door_location}_door_switch
+        open_duration: ${open_duration}
+
+        close_action:
+          - switch.turn_on: close_${door_location}_door_switch
+        close_duration: ${close_duration}
+
+        stop_action:
+          - switch.turn_off: open_${door_location}_door_switch
+          - switch.turn_off: close_${door_location}_door_switch
```
jesserockz: Can we simplify these examples down to a bare minimum, just maybe a single entity with a single substitution and single default

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+Local Packages
+--------------
+
+Consider the following example where the author put common pieces of configuration (like Wi-Fi and I²C) into base files
+and then extends it with some device-specific configuration in the main configuration.
+
+Note how the piece of configuration describing ``api`` component in ``device_base.yaml`` gets merged with the services
+definitions from main configuration file.
+
+.. code-block:: yaml
```
jesserockz, line 34: Lets try to simplify this down. I don't have any specific suggestions though.

components/substitutions/index.rst
```diff
@@ -0,0 +1,181 @@
...
+      name: $devicename
+      # ...
+
+    sensor:
+    - platform: dht
+      # ...
+      temperature:
+        name: ${upper_devicename} Temperature
+      humidity:
+        name: ${upper_devicename} Humidity
```
jesserockz: With the `friendly_name` being added, this is not a good example to have here.

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+      - <<: !include common/sensor/wifi_signal.config.yaml
+    binary_sensor:
+      - <<: !include common/binary_sensor/connection_status.config.yaml
+
+    switch:
+      - <<: !include common/switch/restart_switch.config.yaml
+
+.. _config-git_packages:
+
+Remote/Git Packages
```
jesserockz, line 73: Should have a `Configuration variables` section for this one as it has a specific schema.

components/substitutions/index.rst
```diff
@@ -0,0 +1,181 @@
...
+      foo: yellow
+      bar_yellow_value: !secret yellow_secret
+      bar_green_value: !secret green_secret
+
+    something:
+      test: ${bar_${foo}_value}
+
+.. _YAML-insertion-operator:
+
+YAML insertion operator
```
jesserockz: Not sure this fits in the `substitutions` docs.

components/substitutions/index.rst
```diff
@@ -0,0 +1,181 @@
...
+.. code-block:: yaml
+
+    substitutions:
+      name: default
+      platform: ESP8266
+
+    esphome:
+      name: $name
+      platform: $platform
+      board: $board
```
jesserockz: We should not use the old, deprecated syntax here for an example.

jesserockz (changes requested):

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+        file: file1.yml # cannot be combined with `files`
+        # files: file1.yml
+
+      # shorthand form github://username/repository/[folder/]file-path.yml[@branch-or-tag]
+      remote_package_three: github://esphome/non-existant-repo/file1.yml@main
+
+Packages as Templates
+---------------------
+
+Since packages are incorporated using the ``!include`` system, variables can be provided to them. This means that
```
nielsnl68, line 113: When i checked this is the first time ``!include`` is mentioned in this document, would it be a good idea to explains what the differences are (when there are off-coerce).

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+          open_switch_gpio: 15
+          close_switch_gpio: 18
+          open_duration: "1min"
+          close_duration: "50s"
+
+
+.. code-block:: yaml
+
+    # In garage-door.yaml
+    defaults:
```
nielsnl68: I use "substitutions" here what is the difference?

components/packages/index.rst
```diff
@@ -0,0 +1,259 @@
...
+For example, to set a specific update interval on a common uptime sensor that is shared between configurations:
+
+.. code-block:: yaml
+
+    packages:
+      common: !include common.yaml
+
+    sensor:
+      - id: !extend uptime_sensor
+        update_interval: 10s
```
nielsnl68, line 170: This example does not mirror the documentation above.

components/substitutions/index.rst
nielsnl68: Are we still need to add this anyway? It is already prefix it.

nielsnl68 (commented):
just some points i was wondering about.

components/packages/index.rst
jesserockz: the difference is that `defaults` are scoped to the package only. `substitutions` "leak" out to the root yaml merge. I think

components/substitutions/index.rst
jesserockz: Yeah that was my point why this is a bad example. This example should do something else that is more useful to users.

components/substitutions/index.rst
kbx81: Agreed -- any thoughts on somewhere it _could_ go? I'm drawing a blank. But it is indeed useful...

kbx81:
@jesserockz @nielsnl68 I think I have addressed the feedback from your reviews in [`fcd27b9` (#4095)](https://github.com/esphome/esphome-docs/pull/4095/commits/fcd27b9991dcd7200bba019395e9bd5548d2a5c3). Please have a look and let me know -- thanks! 🍻

components/packages/index.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Configuration variables:
+------------------------
+
+For each package:
+
+- **url** (**Required**, string): The URL for the repository.
+- **username** (*Optional*, string): Username to be used for authentication, if required.
+- **password** (*Optional*, string): Password to be used for authentication, if required.
+- **files** (*Optional*, list of strings): List of files to include. If omitted, all files in the repo are included.
```
jesserockz: Files is required

kbx81: `has_at_least_one_key`...right... 🤦🏻‍♂️

components/packages/index.rst
```diff
@@ -0,0 +1,212 @@
...
+    # any additional configuration...
+
+.. code-block:: yaml
+
+    # In wifi.yaml
+    wifi:
+      ssid: !secret wifi_ssid
+      password: !secret wifi_password
+      domain: .yourdomain.lan
+      fast_connect: true
```
jesserockz:
```suggestion
```

components/packages/index.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. code-block:: yaml
+
+    # In device_base.yaml
+    esphome:
+      name: ${node_name}
+      platform: ESP32
+      board: wemos_d1_mini32
+      build_path: ./build/${node_name}
```
jesserockz:
```suggestion

    esp32:
      board: wemos_d1_mini32
```

components/packages/index.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    # I²C Bus
+    i2c:
+      sda: GPIOXX
+      scl: GPIOXX
+
+    logger:
+      level: ${log_level}
```
jesserockz: Not being supplied anywhere
```suggestion
```

components/packages/index.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    api:
+      encryption:
+        key: !secret api_encryption_key
+      reboot_timeout: 1h
```
jesserockz:
```suggestion
```

jesserockz (changes requested):
Simplify even more =)
We only want people to understand that things can go in different files. We dont need to add so much detail, especially things that most people dont even add/change

jesserockz (approved):

PR #4162, opened 2024-08-16
Removed mode: mono

sanchosk:
## Description:
Seems like there is error in docs - mono seems not to be valid option in mode:, only left/right/stereo are allowed.

## Checklist:

    or


nielsnl68:
IMHO. It is better to implement this instead of removing it. :D

PR #4224, opened 2024-09-04
Add SPI bus to spi_led_strip example

DrBlokmeister:
## Description:
spi_led_strip example was missing an spi bus entry. This PR adds an spi config entry to the example in the documentation.

**Related issue (if applicable):** fixes none

## Checklist:

    or

### DIFF
--- a/components/light/spi_led_strip.rst
+++ b/components/light/spi_led_strip.rst
@@ -18,6 +18,10 @@ LEDs, or any others with a similar interface - SPI, 8 bits per colour and BGR or
 .. code-block:: yaml
 
     # Example configuration entry
+    spi:
+      mosi_pin: GPIO06
+      clk_pin: GPIO07
+
     light:
       - platform: spi_led_strip
         num_leds: 30


DrBlokmeister:
This is still a simple and valid update of the documentation.

ssieb:
Most examples don't include extra bits like this.  There should be a link to the spi docs at the top where it says SPI is required like other components do.

DrBlokmeister:
That makes sense. However, when I look at the MH-Z19 sensor documentation, I see a link to the UART component and a uart_id config option. There it is much clearer that this is needed. In this case, I was quite lost when first setting this up, even though I'm quite experienced with ESPHome.

So I'd like to propose to add a link to the SPI component and to add spi_id (or the relevant key) to the documentation. Does that make sense?

ssieb:
That's how most components do it and that would be ok.  Link to SPI in the text at the top and document the `spi_id:` parameter.  You can copy the text from another SPI device.

DrBlokmeister:
Do you know if spi_id is the valid key?

ssieb:
Yes, that parameter is available for every SPI device.  See https://github.com/esphome/esphome-docs/blob/current/components/sensor/max31865.rst?plain=1#L55 for an example of how to do it.

frenck (approved):
Makes sense, thanks, @DrBlokmeister 👍 

../Frenck

PR #4273, opened 2024-09-23
Remote transmitter trigger docs

swoboda1337:
## Description:

Need a way to seamlessly turn the radio transmitter on / off for certain radios. Add two triggers to accomplish this.

Why this is needed:
- The transceiver could be locked to tx mode permanently but its generally not a good idea. Even in OOK there will be some leakage. If ASK or FSK are used its a really bad idea as it will be transmitting real power all the time.
- Many transceivers support both rx and tx. Without a way to switch modes the transceiver would have to be locked into rx or tx all the time. With this change the transceiver can switch modes automatically.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7483

## Checklist:

    or


swoboda1337:
Screwed up my branches will open a new one

PR #4275, opened 2024-09-23
Add remote transmitter triggers to support auto tx on/off 

swoboda1337:
## Description:

Need a way to seamlessly turn radio hardware transmitters on / off or switch rx / tx modes. Add two triggers to accomplish this.

Why this is needed:
- The transceiver could be locked to tx mode permanently but its generally not a good idea. Even in OOK there will be some leakage. If ASK or FSK are used its a really bad idea as it will be transmitting real power all the time.
- Many transceivers support both rx and tx (but not simultaneously). Without a way to switch modes the transceiver would have to be locked into rx or tx all the time. With this change the transceiver can be in rx mode and automatically switch to tx for a short period of time via these triggers.

These triggers will make supporting various radios / configurations much cleaner. Any state change can be done in one spot now instead of everywhere around every transmit automation. In addition to being cleaner these radio state changes can happen right before / after the actual transmit.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7483

## Checklist:

    or

### DIFF
--- a/components/remote_transmitter.rst
+++ b/components/remote_transmitter.rst
@@ -48,6 +48,26 @@ Configuration variables:
 -  **id** (*Optional*, :ref:`config-id`): Manually specify
    the ID used for code generation. Use this if you have multiple remote transmitters.
 
+Automations:
+------------
+
+- **on_transmit** (*Optional*, :ref:`Automation <automation>`): An automation to perform before
+  data is sent. Useful if the radio / IR hardware needs to change state or power on.
+- **on_complete** (*Optional*, :ref:`Automation <automation>`): An automation to perform after
+  data has been sent. Useful if the radio / IR hardware needs to change state or power off.
+
+.. code-block:: yaml
+
+    # Example automation
+    remote_transmitter:
+      ...
+      on_transmit:
+        then:
+          - lambda: 'id(radio_id)->start_tx();'
+      on_complete:
+        then:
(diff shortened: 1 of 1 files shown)


components/remote_transmitter.rst
```diff
@@ -48,6 +48,26 @@ Configuration variables:
...
+    # Example automation
+    remote_transmitter:
+      ...
+      on_transmit:
+        then:
+          - lambda: 'id(radio_id)->start_tx();'
+      on_complete:
+        then:
+          - lambda: 'id(radio_id)->stop_tx();'
+
```
jesserockz: We should not have an example of something that doesnt yet exist

jesserockz (changes requested):

components/remote_transmitter.rst
swoboda1337: I was hoping that [PR ](https://github.com/esphome/esphome/pull/7483) would get merged, however I can rework the docs.

swoboda1337: Yeah good point but not sure what else to show as an example. Could just show a print I guess?

swoboda1337: Ok I will show a switch or gpio. Updating now

jesserockz (approved):

PR #4300, opened 2024-10-02
Update lvgl.rst - Prevent burn-in of LCD recipe fix

yyolk:
## Description:

This PR removes an extraneous `lvgl.pause:` step under the `turn_off_action` for the "[Prevent burn-in of LCD recipe]" which would result in:

1. lvgl is resumed
2. the screen is redrawn
3. lvgl is paused again (without `show_snow: true`) which is unnecessary

Now the logic is fixed to where it is not paused again after redrawing.

1. lvgl is resumed
2. the screen is redrawn

I also took out the `delay: 1s` under both actions because they weren't explained as being necessary and elsewhere [where `lvgl.widget.redraw` is used], there is no subsequent delay added. Testing it myself, I didn't see a reason to keep the delay either.

[where `lvgl.widget.redraw` is used]: https://esphome.io/cookbook/lvgl.html#turn-off-screen-when-idle
[Prevent burn-in of LCD recipe]: https://esphome.io/cookbook/lvgl.html#prevent-burn-in-of-lcd

## Checklist:

    or

### DIFF
--- a/cookbook/lvgl.rst
+++ b/cookbook/lvgl.rst
@@ -2204,7 +2204,6 @@ In the example below, pixel training is done four times for a half an hour every
               then:
                 - lvgl.resume:
                 - lvgl.widget.redraw:
-                - delay: 1s
           - lvgl.pause:
               show_snow: true
         turn_off_action:
@@ -2214,8 +2213,6 @@ In the example below, pixel training is done four times for a half an hour every
               then:
                 - lvgl.resume:
                 - lvgl.widget.redraw:
-                - delay: 1s
-                - lvgl.pause:
 
     touchscreen:
       - platform: ...


nagyrobi:
The need for a pause there is that, the antiburn is likely started when lvgl is already paused. Now if you resume it manually, it won't pause itself back again unless a physical touch happens, after which the idle counter resets...

yyolk:
> The need for a pause there is that, the antiburn is likely started when lvgl is already paused. Now if you resume it manually, it won't pause itself back again unless a physical touch happens, after which the idle counter resets...

I see the condition for detecting if it's paused underneath the action. 
I don't understand the logic you're eluding to. The action is encountered when it's idle, as I understand a means to turn off the anti-burn cycle. I've had to remove that extraneous pause so the screen doesn't stay frozen after the routine of antiburn cycles. I followed the recipe directly.

What idle counter are you referring to?

nagyrobi:
Lvgl's own idle counter...

yyolk:
@nagyrobi 

> Lvgl's own idle counter...

I don't seem to encounter its own idle counter. Perhaps it's because I'm running a widget update every minute?

Could you elaborate on the logic you're suggesting it's there for? I don't seem to grasp it, and I'm seeing it pause subsequent updates from happening, that's why I removed it.

I see the logic only occurring after the screen_burn is turned off, which implies it was previously on - where lvgl.pause: show_snow=true is necessarily on.

frenck (approved):
Thanks, @yyolk 👍 

../Frenck

PR #4309, opened 2024-10-06
Update homeassistant.rst

byjosue:
Added the note that the component will not work unless home assistant actions are activated in Home Assistant and a brief how-to

## Description:

## Checklist:

    or

### DIFF
--- a/components/switch/homeassistant.rst
+++ b/components/switch/homeassistant.rst
@@ -14,6 +14,11 @@ states and allow **control** via your Home Assistant instance using the :doc:`na
     in Home Assistant, this component still requires you to register the node under Home Assistant. See:
     :ref:`connecting-your-device-to-home-assistant`.
 
+.. warning::
+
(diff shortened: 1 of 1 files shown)


components/switch/homeassistant.rst
```diff
@@ -14,6 +14,11 @@ states and allow **control** via your Home Assistant instance using the :doc:`na
     in Home Assistant, this component still requires you to register the node under Home Assistant. See:
     :ref:`connecting-your-device-to-home-assistant`.
 
+.. warning::
+
+    In order for this component to work, you'll need to allow the device to perform Home Assistant actions by
+    going to the ESPHome integration, clicking 'Configure' on the device and enabling it.
```
frenck, line 20: This is not always the case. Home Assistant will raise a repair in case it detect this might be needed. I think we should omit this suggested documentation, as it might encourage enabling it by default, which is not a practice we should encourage.

Nevertheless, thanks for being willing to contribute 👍 

../Frenck

PR #4360, opened 2024-10-21
Add publish Action for Nextion components

pkejval:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7646

## Checklist:

    or


components/binary_sensor/nextion.rst
```diff
@@ -87,6 +87,56 @@ should be prefixed with the page name (page0/page1).
...
+          state: !lambda 'return true;'
+          # These are optional. Defaults to true.
+          publish_state: true
+          send_to_nextion: true
+
+Configuration options:
+
+- **id** (**Required**, :ref:`config-id`): The ID of the Nextion switch.
+- **state** (**Required**, string, :ref:`templatable <config-templatable>`): The boolean state to publish.
+- **publish_state** (**Optional**, bool, :ref:`templatable <config-templatable>`): Publish new state to HASS. Default is true.
```
edwardtfn: I wouldn't use "HASS". I think "Publish new state to Home Assistant." makes it more clear.

components/binary_sensor/nextion.rst
edwardtfn, line 1: What if one uses `publish_state: false` and `send_to_nextion: false`? Will this function do anything?
Is there really a need for an optional `send_to_nextion`? Why not make it mandatory (always publish to Nextion) if this component is all about the Nextion?

components/binary_sensor/nextion.rst
pkejval: Agreed. I copied this line from existing documentation - lambda call section.

components/binary_sensor/nextion.rst
pkejval: You are right, that call will be nonsense. But I tried to keep same functionality as lambda calls offers. You can check API: https://esphome.io/api/classesphome_1_1nextion_1_1_nextion_component_base

I can image someone needs to publish value into sensor (that triggers filters and on_value actions) and not yet update Nextion...
Yeah, it sounds overcomplicated but as I said I just tried to keep all options possible in lambdas in YAML actions.

PR #4380, opened 2024-10-23
Update installing_esphome.rst

yashmulgaonkar:
Adding instructions for installing on Mac.

### DIFF
--- a/guides/installing_esphome.rst
+++ b/guides/installing_esphome.rst
@@ -48,20 +48,39 @@ with the following:
 
 Mac
 ---
+Install Homebrew using the followng command. If you have it installed already, skip the next two steps.
 
-There are no tested installation instructions for Mac. ESPHome does support
-Mac & will run with no problem.
+.. code-block:: console
 
-Contributions are welcome!
+    $ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 
-The process will likely be similar to Windows. You can install Python from the
-official site, and then install ESPHome with ``pip3 install esphome``.  You can
-then test that things are properly installed with the following:
+Run these commands in your terminal to add Homebrew to your PATH:
+
+.. code-block:: console
+
+    $ echo >> /Users/ymulgaonkar/.zprofile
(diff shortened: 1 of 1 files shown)


hippo5329:
I installed esphome on a Macbook air m3 2024, followed the procedure for Linux. I think homebrew is not required for esphome, though I used brew to install some other applications.

hippo5329:
Just noticed my python3 was installed by homebrew.

hippo5329:
Tested on a Macbook air m3 2024.

1. Install homebrew. It might be better to use an external link to brew.sh site, instead of describing homebrew installation instructions here.

2. brew install esphome

% esphome version
Version: 2024.10.2

ChadMatsalla:
I do not agree that installing homebrew is necessary or even desired to install esphome on a mac. `pip3 install esphome` is simple, unlikely to fail, does not require installing another package manager, and mirrors the install method in windows and Linux.

If anything, the sections for Mac and Linux can be merged because they are effectively identical.

Why do you think that homebrew is a better way?

hippo5329:
How do you install python3 for Mac? I used homebrew to install python3.

ChadMatsalla:
The instructions that already exist say:
```
The process will likely be similar to Windows. You can install Python from the official site, and then install ESPHome with pip3 install esphome. You can then test that things are properly installed with the following:
```
Why would you choose an installation method that is completely different? Is there some advantage to installing an additional package manager?

hippo5329:
At first, I followed the instructions for Linux. I used homebrew to install python3. Then I used the venv and pip3 to install esphome.

Then I tried using homebrew to install esphome.

Both work.

I think both ways can be described in the guide and let users choose.

The point is we use homebrew to install python3 either directly or by dependency. If you haven't installed python3, "brew install esphome" will  install python3 automatically.

hippo5329:
At first, I followed the instructions for Linux. I used homebrew to install python3. Then I used the venv and pip3 to install esphome.

Then I tried using homebrew to install esphome.

Both work.

I think both ways can be described in the guide and let users choose.

The point is we use homebrew to install python3 either directly or by dependency. If you haven't installed python3, "brew install esphome" will  install python3 automatically.

MacOS was based on BSD. Its Unix core is more similar to Linux than Windows. Not quite makes sense to follow Windows way.

PR #4399, opened 2024-10-29
Update send_wait_time text to reflect modbus timing changes

exciton:
Documentation update to reflect change in how send_wait_time is handled in modbus.

Previously send_wait_time had to be set to the longest possible message on the bus, which is difficult to calculate, and error prone. It would clobber incoming messages by sending outgoing commands in the middle of receipt.

New functionality changes this to wait for the response to be complete, and then sending the next command.

Now send_wait_time only needs to be large enough to allow time for devices to send their first byte of response.

## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7674

## Checklist:

    or

### DIFF
--- a/components/modbus.rst
+++ b/components/modbus.rst
@@ -55,10 +55,9 @@ Configuration variables:
   This is useful for RS485 transceivers that do not have automatic flow control switching,
   like the common MAX485.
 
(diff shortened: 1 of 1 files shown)


exciton:
@jesserockz - this relates to PR #7674 which is already merged

nagyrobi (approved):

PR #4422, opened 2024-11-10
Update user manual links for HLK-LD2410 variants

amitkeret:
## Description:
There are three variants of this module, and it would be useful to have direct links to their manuals.
I was searching for this when I wanted to confirm that the C variant uses similar UART configurations as the others. Which it does.

## Checklist:

    or

### DIFF
--- a/components/sensor/ld2410.rst
+++ b/components/sensor/ld2410.rst
@@ -9,8 +9,12 @@ Component/Hub
 -------------
 .. _ld2410-component:
 
-The ``ld2410`` sensor platform allows you to use HI-LINK LD2410 motion and presence sensor
-(`datasheet and user manual <https://drive.google.com/drive/folders/1p4dhbEJA3YubyIjIIC7wwVsSo8x29Fq-?spm=a2g0o.detail.1000023.17.93465697yFwVxH>`__) with ESPHome.
+The ``ld2410`` sensor platform allows you to use HI-LINK LD2410 motion and presence sensors with ESPHome.
+There are three variants with similar communication protocols:
+
+- LD2410 (`datasheet and user manual <https://drive.google.com/drive/folders/1lCQv3mfHJ3XKXzweeHPFnJ_8_D_EWEKk>`__)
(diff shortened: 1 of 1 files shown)


components/sensor/ld2410.rst
```diff
@@ -9,8 +9,12 @@ Component/Hub
 -------------
 .. _ld2410-component:
 
-The ``ld2410`` sensor platform allows you to use HI-LINK LD2410 motion and presence sensor
-(`datasheet and user manual <https://drive.google.com/drive/folders/1p4dhbEJA3YubyIjIIC7wwVsSo8x29Fq-?spm=a2g0o.detail.1000023.17.93465697yFwVxH>`__) with ESPHome.
+The ``ld2410`` sensor platform allows you to use HI-LINK LD2410 motion and presence sensors with ESPHome.
+There are three variants with similar communication protocols:
+
+- LD2410 (`datasheet and user manual <https://drive.google.com/drive/folders/1lCQv3mfHJ3XKXzweeHPFnJ_8_D_EWEKk>`__)
```
frenck, line 15: It is odd to me these are Google Drive folders. Besides that, those folders are owned by a different owner compared what was in here before.

Can we make use of something else in this case? Are there other public resources we can link to?

jesserockz: While I agree on your point, the official page from the manufacturer links to the google drive folder :facepalm: 

https://www.hlktech.net/index.php?id=988

frenck: lol OK... not way out for us. 🙈 

WHY!?! 😭

frenck (approved):
Thanks, @amitkeret 👍 

../Frenck

components/sensor/ld2410.rst
amitkeret: Sorry I'm a bit late to the conversation :)
Just to add confirmation - indeed I got these GDrive links from the Hi-Link website. 
This is from the product page:
hlk


> https://www.hlktech.net/index.php?id=988

The product page @jesserockz mentions has a link to the parent folder on GDrive (also a good one to have!); the links I updated are for each specific model's subfolder.

PR #4443, opened 2024-11-15
Add documentation on Shelly Dimmer calibration capability

### LINKED ISSUE #1711: Support for calibrating dimming devices (e.g. Shelly Dimmer 2)
Support for the Shelly Dimmer 2 was recently added in https://github.com/esphome/esphome/pull/2954.

However, there doesn't appear to be any way to calibrate the dimming device.

This [issue](https://github.com/jamesturton/shelly-dimmer-stm32/issues/11) on the STM32 firmware for the Shelly Dimmer 2 (which the ESPHome) integration uses, I believe) talks about how such calibration needs to be done in the parent framework (e.g. Tasmota).

Is there some way of adding the ability to calibrate the dimmer to ESPHome?

This would allow using the Shelly Dimmer 2's properly - and avoids things like flickering lights, or the dimming controls being non-linear etc.

**Additional context**

On the stock Shelly firmware, the calibration dialog looks like this:

image

### LINKED ISSUE #4400: Shelly Dimmer: Calibrating min and max brightness with helper value
### The problem

Trying to calibrate min and max brightness settings in a high quality way.

I thought that maybe I could use a helper in HA and pull that in to ESPHome to dynamically control those parameters and really fine-tune things but it seems that these parameters are not template-able.

I've tried directly pulling an input helper value in using lambda, setting a global variable and using that global variable as the parameter value but both of those strategies fail with either an error regarding that option "not being template-able".

Here are the relevant bits of my ESPHome Template

```
globals:
  - id: g_max_brightness
    type: int
    initial_value: '500'

sensor:
 - platform: homeassistant
    name: "Helper Max Brightness"
    id: helper_max_brightness
    entity_id: input_number.calibrate_bathroomhallwaydimmer_high
    on_value:
      then:
        - globals.set:
            id: g_max_brightness
            value: !lambda 'return int(x);' 

light:
  - platform: shelly_dimmer
    name: "${device_name}"
    id: hallwaydimmer
    icon: mdi:globe-light-outline
    power:
      name: "${device_name} Power"
      id: power_dimmer
    voltage:
      name: "${device_name} Voltage"
    current:
      name: "${device_name} Current"
    min_brightness: 24
    #max_brightness: 500
    max_brightness: g_max_brightness
    firmware:
      version: "51.6"
      update: true
```

The alternative of course is to manually change the max and min brightness values and then run an install, but this would be a much more time consuming process to get right.

2023.3.2

Home Assistant Add-on

2023.4.2

ESP8266

### Board

Shelly Dimmer 2

Shelly_Dimmer

### Example YAML snippet

```yaml
See above
```

_No response_

### Additional information

_No response_

olegtarasov:
## Description:

This PR documents the process of calibrating Shelly Dimmer using its new capability.

- fixes https://github.com/esphome/feature-requests/issues/1711
- fixes https://github.com/esphome/issues/issues/4400

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7773

## Checklist:

    or


olegtarasov:
Wow, this is disappointing. After a year of investing the time to keep up with changes from upstream, the PR is closed by a bot. @jesserockz

PR #4483, opened 2024-11-29
Create guide for using an ESP devboard with USB_UART bridge chip as serial adapter

shvmm:
All functioning ESP devboards which can be flashed over USB are automagically also USB serial adapter boards.
The file `guides/devboard_as_flasher.rst` documents how to use one.

## Description:
devboard-as-flasher-pi

pi serial tty

Sources:
https://pinout.xyz/
PNGs Credits:
https://freesvg.org/esp32-devkitc
https://freesvg.org/by/adlerweb
Idea from this video:
https://www.youtube.com/watch?v=ndEeFQ5mi_g

## Checklist:

    or


guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
+Using an ESP devboard as serial flasher
+=======================================
```
clydebarrow:
```suggestion
Using an ESP devboard as a USB-UART bridge
==========================================
```

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. _devboard-as-flasher:
+
+If you have a functioning ESP devboard onto which you can flash firmware via USB port,
+it can be also used to flash/communicate via serial to other devices as well.
```
clydebarrow:
```suggestion
ESP development boards usually have an onboard USB interface, either built into the chip (e.g. ESP32-S3) or via a USB-UART bridge chip. However some ESP based devices not designed for development work don't bother with this, and only expose the UART0 pins (TX and RX) for flashing purposes.

Normally you would use a dedicated USB-UART interface board for this but what if you don't have one? In this "emergency" situation it is possible to use a development board that does have a USB-UART bridge chip to flash another device. This is achieved by holding the ESP chip in reset so that it doesn't interfere with the bridge chip operation.

It does NOT require any firmware to be flashed onto the development board and will not change anything already flashed onto it - it's purely a way to use the serial interface chip.

Note: most ESP32 S and C series boards do *not* have a separate USB-UART chip - they have it built into the ESP - so are not suitable for this application.
```

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+- ``TX`` of flasher devboard to ``TX`` of the target device
+- ``RX`` of flasher devboard to ``RX`` of the target device
+
+Pulling down ``EN`` by connecting it to ``GND`` on the flasher board prevents
+the ESP chip on flasher module from booting and polluting the serial lines.
+
+.. note::
+
+    - On some boards like the AiThinker ESP32Cam, you additionally need to connect ``IO0`` with ``GND``
+      on the target device to force flash/download mode.
```
clydebarrow:
```suggestion
    - If the board has not previously had ESPHome loaded, it is probable that you will need to hold the ``IO0` pin low (i.e. connected to ground) to force the board into flash mode. This must be done before power is applied.
```

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+Make sure you've read the :doc:`/guides/physical_device_connection` for properly understanding of the functionality of your flasher devboard.
+
+.. figure:: /guides/images/devboard-as-flasher.png
+    :align: center
+    :width: 75.0%
+
+    Connection diagram for an ESP flash target
+
+You need to make the following electrical connections:
+
```
clydebarrow, line 28:
```suggestion
You need to make the following electrical connections:

Note that the 5V connection on either board may be labelled either ``5V`` or ``VIN``. Some boards may not have a 5V connection and will require 3.3V only.

Rather than powering the target board from the flasher board, it is also possible to use a separate power supply, just make sure all the ground pins are connected together.
```

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+    .. figure:: /guides/images/devboard-as-flasher-pi.jpg
+        :align: center
+        :width: 75.0%
+
+        Connection diagram for a RPi serial target. Refer to https://pinout.xyz/
+
+    .. warning::
+
+        Powering a Raspberry Pi directly through the flasher devboard is not advised.
```
clydebarrow: Is this section necessary? It seems too specific and may just cause confusion.

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+- ``RX`` of flasher devboard to ``RX`` of the target device
+
+Pulling down ``EN`` by connecting it to ``GND`` on the flasher board prevents
+the ESP chip on flasher module from booting and polluting the serial lines.
+
+.. note::
+
+    - On some boards like the AiThinker ESP32Cam, you additionally need to connect ``IO0`` with ``GND``
+      on the target device to force flash/download mode.
+    - Do not connect 3V3 to VIN of the target devices with a 3V3 LDO as it may lead to brownouts.
```
clydebarrow, line 48:
```suggestion
    - Do not connect 3V3 to VIN of the target devices with a 3V3 LDO as it may lead to brownouts.
    
Once the connections are made, plug the flasher board into your computer via USB and proceed with flashing the target board via whichever means you intend to use.
```

clydebarrow (changes requested):
I have to say it took me a little time to realise what this guide was trying to show. Once the penny dropped that you are just using the USB-UART bridge chip it made more sense. So I've suggested additional clarification to make that clear.

I'd also suggest that the image showing the wiring connections has a poor choice for the target board - its image shows it *does* have a USB connector and a USB-UART chip, so there would be no need to use this technique to flash that board.

Finally, it would be worth adding a link to this in the `physical_device_connection` page, probably in the Materials table where it talks about a USB to serial adapter, since this what the dev board would be replacing.

guides/devboard_as_flasher.rst
```diff
@@ -0,0 +1,64 @@
...
+    .. warning::
+
+        Powering a Raspberry Pi directly through the flasher devboard is not advised.
+
+    Future Todo
+    -----------
+    
+    Verify functionality for UART debug ports on the following devices:
+        - Raspberry Pi Pico series
+        - Raspberry Pi 5
```
clydebarrow: I would remove this. It's not useful to the reader.

shvmm: Yeah, I'd move this artifact to the main PR description

guides/devboard_as_flasher.rst
shvmm: The plan is to expand this section to include debug instructions for RPico series boards. That's why
It is in a separate collapsible tho

shvmm:
> I'd also suggest that the image showing the wiring connections has a poor choice for the target board - its image shows it _does_ have a USB connector and a USB-UART chip, so there would be no need to use this technique to flash that board.

It is easier for me to just remove the USB connector, chip and accompanying mosfets from the target board.
The harder part would be finding a high quality svg for a target board with proper pinout labels.

shvmm:
> Finally, it would be worth adding a link to this in the `physical_device_connection` page, probably in the Materials table where it talks about a USB to serial adapter, since this what the dev board would be replacing.

Added two links:
One in the section devoted to explaining the functionality of development boards 
And another in materials required section.

clydebarrow (approved):
LGTM

PR #4485, opened 2024-11-30
Added docs for new `to_ntc_*` sensor filter

rhabacker:
Ref https://github.com/esphome/esphome/pull/7898

## Description:
This pr adds documentation for the `to_ntc_resistance` and `to_ntc_temperature` sensor filters.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7898

Depends on  https://github.com/esphome/esphome-docs/pull/4735

## Checklist:

    or


rhabacker:
Anything left to do for this pull request ?

components/sensor/ntcsim.rst
```diff
@@ -0,0 +1,93 @@
...
+Configuration variables:
+------------------------
+
+- **sensor** (**Required**, :ref:`config-id`): The sensor to read the temperature values from
+  to convert to resistance readings.
+- **output** (**Optional**, :ref:`output`): The output to send resistance readings to.
+- **output_resistance** (**Optional**, float): NTC resistance value by which the value sent
+  to the output is divided. Scaling is required as the output component can only receive values
+  between 0 and 1.
+- **calibration** (**Required**, float): The calibration parameters of the sensor - see above
```
DjordjeMandic:
```suggestion
- **calibration** (**Required**, float or list): The calibration parameters of the sensor - see above
```

Even this does not exactly fit. calibration is not single float. 

```suggestion
- **calibration** (**Required**, list): The calibration parameters of the sensor - see above
```

This might be better

DjordjeMandic (changes requested):

DjordjeMandic (dismissed):

clydebarrow:
What was the reason for using an include rather than putting the new filters inline? The issue with that is it results in creation of two new pages as well as the text appearing in the sensor page. It may well be that the sensor folder could do with a reorganisation but that's outside the scope of this PR.

rhabacker:
> What was the reason for using an include rather than putting the new filters inline? 

Including everything in one file makes it difficult to get a sorted view and add new entries in the right place. Keeping parts of the documentation separate and including them in other files is already used in other places. 

> The issue with that is it results in creation of two new pages as well as the text appearing in the sensor page.

Fixed with commit 727c653

> It may well be that the sensor folder could do with a reorganisation but that's outside the scope of this PR.

This is intended with https://github.com/esphome/esphome-docs/pull/4735

clydebarrow:
It's best not to force-push unless needed to recover from a hopeless mess. The commits will be squashed when merging anyway, and force-pushing obscures what changes have been made, which makes reviewing harder.

clydebarrow:
> This is intended with #4735

So defer that reorganisation until then, others may have opinions about it. Just keep it inline for now to keep the PRs separate.

rhabacker:
> > This is intended with #4735
> 
> So defer that reorganisation until then, others may have opinions about it. Just keep it inline for now to keep the PRs separate.

Where should new filter documents be inserted? At the end after `lambda` or after `throttle_average` to at least partially maintain an alphabetical order?

clydebarrow:
Your guess is as good as mine, and probably not so important since it will be revised in due course anyway.

clydebarrow (approved):

PR #4494, opened 2024-12-08
Fix description of `accuracy_decimals`

damz:
## Description:

**Related issue (if applicable):** fixes #2164

The description introduced in #2164 doesn't make sense: it says that the `accuracy_decimals` setting both controls and doesn't control the number of decimals to use when reporting values.

This setting plainly DOES impact the actual value reported to Home Assistant. It also controls the default display precision of the Home Assistant entity, but I do not believe this is a very important information in this context. In the interest of simplicity, I suggest we just remove the second sentence.

## Checklist:


damz:
Looking into this a bit more: it is technically true that `accuracy_decimals` _"does not impact the actual value reported to Home Assistant"_. But the first thing that the Home Assistant integration does is to [round the received value based on `accuracy_decimals`](https://github.com/home-assistant/core/blob/9f0356fcfed89b6ec134235f98b837ae853da1c7/homeassistant/components/esphome/sensor.py#L105). That makes the second part of the sentence (_"it just sets the number of decimals to use when displaying it"_) incorrect.

PR #4499, opened 2024-12-10
[nextion] Adds a command pacer with `command_spacing` attribute

edwardtfn:
## Description:

This introduces a "pacer" for the communication with the Nextion display, with the main objective to reduce the chances of a buffer overflow on Nextion side.
The buffer overflow can be quite critical, as it removes it's queue, while the component on ESPHome keeps waiting for responses for those removed commands, so a way to control the communication pace and give to users a chance to reduce the chances of a buffer overflow by compromising (a bit?) the performance which many times wouldn't be perceived by end users.
As default I'm setting the pace to 0ms, to keep this transparent to existing installation, although I would highly recommend the use of at least 10ms. Anyhow, users can increase this dynamically, if a buffer overflow happens.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7948

## Checklist:

    or


edwardtfn:
This should not be stale

jesserockz (approved):

PR #4528, opened 2024-12-21
Documentation for Duco component

kokx:
## Description:

Documentation for a newly created Duco component.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#7993

## Checklist:

    or


jorritolthuis:
I have just deployed it on my D325, it works really well! :heart:

Couple of questions:
- The comfort temperature setting does not seem to be mentioned in the docs. Any reason for this?
- A suggestion to also state explicitly in the docs that users need to also connect their ESP to one of the ground pins (of course being careful not to short USB and Duco ground). Without ground, UART will probably not work.
- Any clue what the BOX, UCBAT, SWITCH and UC nodes are (I only detect BOX and UCBAT myself)?
- Possibly related: I can see on the Duco display that it can measure indoor and outdoor temperatures. Can those sensors be read over UART as well?

kokx:
I'm happy that it works well for you!

About the comfort temperature setting, that is an omission from my side, thank you for pointing it out. I will document it as well (probably after christmas).

Thank you as well for the note about connecting one of the ground pins of the ESP as well. Will add it too.

About the BOX, UCBAT, SWITCH and UC nodes, they are part of the controls of the ventilation system. The BOX component represents the ventilation box itself, the UCBAT represents a battery-powered remote control, the SWITCH component seems to be a sensor for the current mode and the UC component represents a wired wall-switch.

My system has a BOX, UCBAT, UCCO2 (2x), SWITCH and UC node. But your system may not have as many (most of them are add-ons). Do please let me know if you do have a wired remote control or CO2-sensors (or anything else that doesn't show up in discovery), because then discovery of nodes may not work correctly.

(Do take this with a grain of salt, most information I got from the [Modbus TCP documentation](https://www.duco.eu/Wes/CDN/1/Attachments/information-sheet-Modbus-TCP-(en)_638543249352352865.pdf) that duco published themselves and from [reverse engineering](https://github.com/kokx/duco-analysis) the protocol used by the Connectivity Board.)

The indoor- and outdoor temperatures are something that I would myself like to read as well. Unfortunately the Connectivity Board of Duco itself doesn't read them either, so capturing how that works wasn't possible. It is very likely that it is possible to get this from the box, but at the moment I do not know how.

PR #4541, opened 2024-12-28
Update rf_bridge.rst

zd3sf:
Added references to mightymos firmware

## Description:
1)  Clarified introduction for easier readability.
2) Clarified uart_id configuration variable 
3) Added reference to "Mightymos" firmware alongside Portisch
4) Moved bridge.beep out of portisch section because no Portisch is needed for beeping
5) Add a link for another, more user friendly B1 converter
6) extended the example at the end. 

Closes #4521 

## Checklist:

    or


nagyrobi (commented):
Removed `uart_id` and put `id` twice in the example?

zd3sf:
Hi thanks for the review. 

Yes UART_id is not a valid syntax in esphome, there's a `id` for the UART component and and `id` for the bridge component. UART_id was confusing to some users and I have seen few questions about it in the community forum. we can remove the `id` for UART hub as users can go to the UART component page if they want to learn more about UART.

nagyrobi:
You can't have in the documentation, in the same section two `id` options meaning different things...

zd3sf:
Now worries, UART id configuration variable was removed.

zd3sf:
@nagyrobi Is it good to go now?

zd3sf:
Are we still reviewing this?

zd3sf:
@nagyrobi. Please let me know if you propose any additional changes.

zd3sf:
Still pending review?

components/rf_bridge.rst
```diff
@@ -19,17 +23,21 @@ which is 19200bps.
...
+    Sonoff RF Bridge 433 (version R1 or R2 V1.0)
 
 .. code-block:: yaml
 
     # Example configuration entry
     uart:
+      id: uart_hub
       baud_rate: 19200
+      tx_pin: 1
+      rx_pin: 3
```
jesserockz:
```suggestion
```

components/rf_bridge.rst
```diff
@@ -41,9 +49,7 @@ which is 19200bps.
 
 Configuration variables:
 ------------------------
-
-- **uart_id** (*Optional*, :ref:`config-id`): Manually specify the ID of the UART hub.
```
jesserockz, line 45: This shouldnt be removed

components/rf_bridge.rst
```diff
@@ -41,9 +49,7 @@ which is 19200bps.
...
-- **id** (*Optional*, :ref:`config-id`): Manually specify the ID used for code generation.
+- **id** (*Optional*, :ref:`config-id`): Manually specify the ID of the RF bridge. Used for code generation and event tracking in home assistant.
```
jesserockz: Not sure where this statement came from? The id doesn't leave the device
```suggestion
- **id** (*Optional*, :ref:`config-id`): Manually specify the ID of the RF bridge. Used for code generation. 
```

jesserockz (changes requested):

components/rf_bridge.rst
zd3sf: Previous reviewer wanted it removed, and I agree. It pertains to UART component, it is shown here as an example of essential config to make the bridge work, but its a configuration variable for UART not RFbridges. Let me know what you think.

components/rf_bridge.rst
zd3sf: hmm I'll phrase it in a better way. The ID of the hub gets attached to events in home assistant. So when the user wants to listen to an RF code, they subscribe to an event based on the ID of the hub.

components/rf_bridge.rst
zd3sf: The RF bridge will not run without a UART component. I think it is necessary to keep it in the example config.

jesserockz: While this is true, this is not how we write the basic examples in esphome. 
A full example should be added to https://devices.esphome.io/devices/Sonoff-RF-Bridge

components/rf_bridge.rst
jesserockz: The `uart_id` is a configuration item of the `rf_bridge` component. 

UART components themselves have `id` which matches the value.

It is automatically filled by ESPHome if there is a single UART entry configured which is why it is optional.

components/rf_bridge.rst
jesserockz: As above this id never leaves the device, even for the events in the examples below. 

If you want a specific hub id in the event, you would add one as the user, but that has nothing to do with this id, even if the value is the same

components/rf_bridge.rst
zd3sf: You are correct. I will add it back!

components/rf_bridge.rst
zd3sf: Yep makes sense. I removed it!

components/rf_bridge.rst
zd3sf: Correct again. I removed "Used for code generation."

jesserockz (approved):

PR #4557, opened 2025-01-04
esphome.rst: Show on_boot/on_Shutdown priority as list item

hramrach:
## Description:

Multiple priorities can be used in a list. Show the value as a list item in the code example to make this clear.

## Checklist:

    or


nagyrobi (approved):

nagyrobi:
Thanks

PR #4571, opened 2025-01-09
increase number of cells from 16 to 18

### LINKED ISSUE #3018: Daly BMS higher than 16s cell voltage measure

The usage of the Daly-BMS module is really useful but it is limited to 16 series of cells. There are Daly BMS modules with higher number of series. It would be needed to have more than 16S cells voltage measures at least.
The specification is showing from 1~48 cells in most parameters. Just getting the voltage of each cell it would be a small change and would allow the usage of this module for higher number of cells than 16.  
Nowadays a big number of batteries are using more than 16 series of cells, as the usage of batteries for higher power systems is needed.

Get the voltage of more than 16s cells in Daly-BMS module. as in the official documentation is exposed that all the values from 1~48 cells will be available, the change should not be disruptive.

**Additional context**

There are Daly BMS modules with more than 16S cells. so it should be matching accordingly the maximum number of cells permitted in the uart protocol.

## Labels
- [integration: daly_bms](https://github.com/esphome/feature-requests/labels/integration%3A%20daly_bms)

j-sepul:
## Description:
increase number of cells from 16 to 18

**Related issue (if applicable):** fixes [3018](https://github.com/esphome/feature-requests/issues/3018)

## Checklist:

    or
  - [*] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.

### DIFF
--- a/components/sensor/daly_bms.rst
+++ b/components/sensor/daly_bms.rst
@@ -127,7 +127,7 @@ Configuration variables:
 - **temperature_2** (*Optional*): The second temperature sensor.
   All options from :ref:`Sensor <config-sensor>`.
 
-- **cell_1_voltage** (*Optional*): The voltage of cell number 1. Cell number can be from 1 to 16.
+- **cell_1_voltage** (*Optional*): The voltage of cell number 1. Cell number can be from 1 to 18.
   All options from :ref:`Sensor <config-sensor>`.
 
 Text Sensor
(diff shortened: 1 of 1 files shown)


components/sensor/daly_bms.rst
```diff
@@ -78,6 +78,34 @@ A sensor platform to read BMS data
...
+        cell_14_voltage:
+          name: "Cell 14 Voltage"
+        cell_15_voltage:
+          name: "Cell 15 Voltage"
+        cell_16_voltage:
+          name: "Cell 16 Voltage"
+        cell_17_voltage:
+          name: "Cell 17 Voltage"
+        cell_18_voltage:
+          name: "Cell 18 Voltage"
```
jesserockz: Don't need to bloat it this much
```suggestion
```

jesserockz (approved):

PR #4577, opened 2025-01-11
Update adc.rst attenuation values

basvanderwielen:
## Description:

Attenuation can only be set to the following values: 0db, 2.5db, 6db, 11db according to documentation.

## Checklist:

    or


ximex:
@basvanderwielen sure this is correct? in the source code there is a check that says `11dB` is deprecated and you should use `12dB` instead

jesserockz:
12 is the correct value

PR #4637, opened 2025-02-03
Add available range to measurement interval of SCD30

JMyrng:
added a note about the availiable range found in datasheet

## Description:

After I had entered a measuring time of 1s and received measurements that were far too high, I looked in the data sheet and found a minimum measuring time of 2s. 

## Checklist:

    or


DjordjeMandic:
https://github.com/esphome/esphome/blob/dev/esphome%2Fcomponents%2Fscd30%2Fsensor.py#L75-L79

Codegen allows you to set it to 1. That should be fixed also.

jesserockz (approved):

PR #4643, opened 2025-02-05
Document packet_transport and udp changes

clydebarrow:
## Description:

Note: there are improvements to this PR incorporated into #4645, too messy to backport those here but merging #4645 will bring those in.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#8187
## Checklist:

    or


components/binary_sensor/packet_transport.rst
```diff
@@ -1,21 +1,27 @@
...
+      - platform: packet_transport
         id: switch_status
         provider: light-switch
         remote_id: light_switch
 
+     packet_transport:
+       - platform: udp
+
+     udp:
+
```
jesserockz: The above:
> It requires a ``packet_transport`` component to be configured.

should be a link to the page for the users to follow and configure accordingly.
```suggestion
```

components/packet_transport/index.rst
jesserockz, line 1: While `udp` is currently the only platform available, this page should not really include that in the examples unless at the end with a full example, and should be more like the sensor core pages that explain the common configurations and general idea of this component.

Ideally, there should be a page at `components/packet_transport/udp.rst` as well which is the platform implementation configuration.

components/sensor/packet_transport.rst
```diff
@@ -1,21 +1,27 @@
...
+      - platform: packet_transport
         id: temperature_id
         provider: thermometer
         remote_id: temp_id
 
+     packet_transport:
+       - platform: udp
+
+     udp:
+
```
jesserockz: Same as above
```suggestion
```

jesserockz (changes requested):

components/packet_transport/index.rst
clydebarrow: > there should be a page at `components/packet_transport/udp.rst` as well which is the platform implementation configuration

See #4645

components/sensor/packet_transport.rst
jesserockz:
```suggestion
     packet_transport:
       - platform: ...

```

jesserockz (approved):

PR #4677, opened 2025-02-17
Clarify LVGL widget update properties

bdm310:
## Description:

Most widgets with update actions don't adequately describe what properties can be changed with the action. This leaves the impression that either far less can be changed than actually can or that properties can be changed that aren't actually accessible.

The `animated` property is written as if it is a persistent setting of widgets that have it. In reality, it is a per-update and per-creation single use flag.

Also fixed some minor mismatches with actual implementation.

**Related issue (if applicable):** fixes [<link to issue>](https://github.com/esphome/issues/issues/6759)

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** esphome/esphome#8260
## Checklist:


nagyrobi:
@clydebarrow

clydebarrow (approved):

PR #4693, opened 2025-02-20
More details for ARMv7

ssieb:
## Description:

expands the ARMv7 section to be more clear.

## Checklist:

    or

### DIFF
--- a/changelog/2025.2.0.rst
+++ b/changelog/2025.2.0.rst
@@ -57,7 +57,7 @@ Custom components were deprecated over a year ago and, as promised, support for 
 encourage all developers to migrate their custom components to :doc:`/components/external_components`. Please see
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
-Removal of ``armv7`` Docker (and Home Assistant add-on) Support
+Removal of ``armv7`` Docker and Home Assistant add-on support
 ---------------------------------------------------------------
 
 ESPHome no longer supports ``armv7`` (32-bit ARM) hardware. This is due to both waning support as it relates to tooling and
--- a/changelog/2025.2.0.rst
+++ b/changelog/2025.2.0.rst
@@ -58,7 +58,7 @@ encourage all developers to migrate their custom components to :doc:`/components
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
 Removal of ``armv7`` support
--------------------------------------------------------------
+----------------------------
 
 ESPHome no longer supports the 32-bit  ``armv7`` architecture. This is due to both waning support as it relates to tooling and
 performance reasons. We strongly recommend moving to a more modern architecture, especially if you're using the ESPHome
--- a/changelog/2025.2.0.rst
+++ b/changelog/2025.2.0.rst
@@ -64,11 +64,9 @@ ESPHome no longer supports ``armv7`` (32-bit ARM) hardware. This is due to both 
 performance reasons. We strongly recommend moving to a more modern architecture, especially if you're using the ESPHome
 Device Builder to build/compile firmware for your devices.
 
-- The toolchains required for compiling ESPHome are dropping support for armv7, so if the toolchain is not available,
-ESPHome can't do the compile.
+- The toolchains required for compiling ESPHome are dropping support for armv7, so if the toolchain is not available, ESPHome can't do the compile.
 - Building on armv7 is very slow and is likely to crash due to lack of RAM, which is not a good user experience.
-- ESPHome can be installed anywhere you can run Docker or Python pip, which will provide a better experience with
-much faster compiling.
+- ESPHome can be installed anywhere you can run Docker or Python pip, which will provide a better experience with much faster compiling.
 
 ``esp32_rmt`` Updates
 ---------------------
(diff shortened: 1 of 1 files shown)


changelog/2025.2.0.rst
```diff
@@ -57,12 +57,13 @@ Custom components were deprecated over a year ago and, as promised, support for
 encourage all developers to migrate their custom components to :doc:`/components/external_components`. Please see
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
-Removal of ``armv7`` Docker Support
------------------------------------
+Removal of ``armv7`` Docker (and HA add-on) Support
```
frenck: We shouldn't be abbreviating Home Assistant as `HA`

changelog/2025.2.0.rst
```diff
@@ -57,12 +57,13 @@ Custom components were deprecated over a year ago and, as promised, support for
...
+---------------------------------------------------
 
-ESPHome no longer supports ``armv7`` hardware. This is due to both waning support as it relates to tooling and
+ESPHome no longer supports ``armv7`` (32-bit ARM) hardware. This is due to both waning support as it relates to tooling and
```
frenck: 32-bits ARM is not just `armv7`, so this addition seem incorrect.

changelog/2025.2.0.rst
```diff
@@ -57,12 +57,13 @@ Custom components were deprecated over a year ago and, as promised, support for
...
 performance reasons. We strongly recommend moving to a more modern architecture, especially if you're using the ESPHome
 Device Builder to build/compile firmware for your devices.
+`More detailed explanation. <https://github.com/esphome/issues/issues/6770#issuecomment-2672644628>`__
```
frenck: It shouldn't be needed to link issues. If there is missing context here, we should add it. Driving users into technical issues isn't something we should aim for.

frenck (dismissed):

changelog/2025.2.0.rst
ssieb: Assuming we're not also providing `armhf`, then it's correct.  And it's more clear for users that don't understand what `armv7` is.  The 64-bit platforms actually have `64` in the name.

changelog/2025.2.0.rst
ssieb: Do you want me to copy that section into here?  I was trying to avoid bloating the changelog needlessly.

changelog/2025.2.0.rst
frenck: Sure, but there are multiple 32 bits platforms for 32 bits.... 🤷

changelog/2025.2.0.rst
frenck: That is not what I'm suggesting. I'm suggesting that if the context is incomplete, we need to complete it.

changelog/2025.2.0.rst
ssieb: But are we providing any of them?

frenck: from the add-on perspetive, we could, yes; so we should be clear IMHO.

changelog/2025.2.0.rst
```diff
@@ -57,13 +57,19 @@ Custom components were deprecated over a year ago and, as promised, support for
 encourage all developers to migrate their custom components to :doc:`/components/external_components`. Please see
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
-Removal of ``armv7`` Docker Support
------------------------------------
+Removal of ``armv7`` Docker (and Home Assistant add-on) Support
```
frenck:
```suggestion
Removal of ``armv7`` Docker and Home Assistant add-on support
```

changelog/2025.2.0.rst
kbx81: Maybe this?

```suggestion
ESPHome no longer supports the 32-bit  ``armv7`` architecture. This is due to both waning support as it relates to tooling and
```

changelog/2025.2.0.rst
```diff
@@ -57,13 +57,17 @@ Custom components were deprecated over a year ago and, as promised, support for
 encourage all developers to migrate their custom components to :doc:`/components/external_components`. Please see
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
-Removal of ``armv7`` Docker Support
------------------------------------
+Removal of ``armv7`` Docker and Home Assistant add-on support
+---------------------------------------------------------------
```
kbx81:
```suggestion
-------------------------------------------------------------
```

changelog/2025.2.0.rst
```diff
@@ -57,13 +57,17 @@ Custom components were deprecated over a year ago and, as promised, support for
 encourage all developers to migrate their custom components to :doc:`/components/external_components`. Please see
 :ref:`a_note_about_custom_components` for a detailed explanation and next steps.
 
-Removal of ``armv7`` Docker Support
------------------------------------
+Removal of ``armv7`` Docker and Home Assistant add-on support
```
DjordjeMandic:
```suggestion
Removal of ``armv7`` Docker (incl. Home Assistant add-on support for ``armv7``) 
```
Title looks like HA add-on is completely unsupported.

frenck: Yeah agreed with that comment. The other approach would be just:

```suggestion
Removal of ``armv7`` support
```

Simple, to the point, applies everywhere.

ssieb: If you're going to change the title text, then you need change the underline too.  Otherwise, it can't be trivially committed. :-)

DjordjeMandic (approved):

PR #4783, opened 2025-03-31
Update time and temperature cookbook

yincrash:
## Description:
Use a temperature that comes with the default Home Assistant setup - weather 
Show how to use google fonts so a user doesn't have to setup font files to get something running.

## Checklist:

    or


yincrash:
I'm a new ESPHome user and definitely really love the cookbooks, but I wanted to get something up and running as fast as possible to see what I could do, and so I thought a few changes to this cookbook would make it simpler for a new user who may not have a temperature sensor set up, but could get something showing up immediately.

cookbook/display_time_temp_oled.rst
```diff
@@ -61,7 +61,15 @@ By adding ``internal: true`` to the sensors they won't be published back to Home
 
       - platform: homeassistant
         id: outside_temperature
-        entity_id: sensor.10_00080192969d_temperature
+        entity_id: weather.forecast_home
+        attribute: temperature
```
jesserockz: We actually have a placeholder for this that will generate a configuration error in ESPHome, ensuring the user adjusts to match their set up.
```suggestion
        entity_id: REPLACEME
```

jesserockz (changes requested):

cookbook/display_time_temp_oled.rst
jesserockz: My homeassistant does not have a `weather.forecast_home` entity, it is `weather.<suburb>` for me

jesserockz (approved):

PR #4816, opened 2025-04-12
Update to correct math for watt/kilowatt conversion

jsnfwlr:
1kW = 1000 watts
1 ÷ 1000 = 0.001

See [Home Assistant Glow config](https://github.com/klaasnicolaas/home-assistant-glow/blob/develop/components/pulse_meter.yaml#L79) for reference

## Checklist:

    or


nagyrobi:
That depends on the meter type. There are meters pulsing 10000 times per kWh.

jsnfwlr:
I see. Maybe the change should be to make that more prominent then

PR #4866, opened 2025-05-01
HA style version label

shvmm:
## Description:
HA style version label from home-assistant.io

NONE
NONE

## Checklist:

    or

### DIFF
--- a/_static/index.css
+++ b/_static/index.css
@@ -293,6 +293,12 @@
     .feature-icon {
         color: #00bfff;
     }
+
+    .release-date a {
+        background: #393939;
+        color: #3fc3ff !important;
+    }
+
 }
 
 @media (max-width: 768px) {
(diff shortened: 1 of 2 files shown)


DjordjeMandic:
This does not look centered compared to version on https://www.home-assistant.io/
image

image

shvmm:
> This does not look centered compared to version on https://www.home-assistant.io/ image
> 
> image

Centering is probably the easy part.
The harder part is templating the version number string and Date released so that it can be built automatically at release

DjordjeMandic:
> Centering is probably the easy part. The harder part is templating the version number string and Date released so that it can be built automatically at release

I agree, that would be the best solution. Did you think of any specific way of doing it? Github workflow or maybe python script?

shvmm:
> > Centering is probably the easy part. The harder part is templating the version number string and Date released so that it can be built automatically at release
> 
> I agree, that would be the best solution. Did you think of any specific way of doing it? Github workflow or maybe python script?

I thought of adding needed release date consts just like version and then referencing them all together

PR #4867, opened 2025-05-01
Add rounded backgrounds to feature-icons

shvmm:
## Description:

Add rounded backgrounds to feature icons

Before:

After:

NONE
NONE

## Checklist:

    or


mrdarrengriffin:
@shvmm, is it possible to get the icon circles to be round. The padding is messing with it I think. Adding a fixed width/height or using aspect-ratio should work

abmantis:
Personally, I prefer the current cleaner look without the icon backgrounds. The background makes the icons harder to see due to the similar colors. It also highlights the icons too much (which I don't think they need, since they are not interactive).

Still, if the backgrounds are added then they should be consistent circles.

shvmm:
> Personally, I prefer the current cleaner look without the icon backgrounds. The background makes the icons harder to see due to the similar colors. It also highlights the icons too much (which I don't think they need, since they are not interactive).
> 
> Still, if the backgrounds are added then they should be consistent circles.

You are correct.
Even to me, its look in dark mode isn't even close to how good it looks in light mode.

> @shvmm, is it possible to get the icon circles to be round. The padding is messing with it I think. Adding a fixed width/height or using aspect-ratio should work

Thanks. Will try the aspect-ratio approach

mrdarrengriffin:
The preview looks better, just make the `.feature-icon i` have `display: flex; align-items: center;` to ensure they are aligned correctly

shvmm:
> The preview looks better, just make the `.feature-icon i` have `display: flex; align-items: center;` to ensure they are aligned correctly

Done. Thanks for the tip. I have also tried fixing dark mode colors

Looks mostly good for now.

clydebarrow:
I personally prefer it without the background. But best wait for the new site and see what you think.

PR #4874, opened 2025-05-03
Add YAML doc

### LINKED ISSUE #6978: [Docs] Information about yaml is impossible to find
### The problem

Information about the yaml features and how to create/use them is impossible to find.
[Substitions](https://esphome.io/components/substitutions.html#substitute-include-variables) is under "supported microcontrollers"

Image

But if I click it then a link to the page does not appear.
But substitutions also describes the `!include` operation which is not described anywhere else.

Imho there should be a dedicated `yaml` section which describes these things.

latest

pip

_No response_

ESP8266

### Board

_No response_

_No response_

### YAML Config

```yaml

```

```txt

```

### Additional information

_No response_

clydebarrow:
## Description:

Add a comprehensive guide to ESPHome YAML syntax.

**Related issue (if applicable):** fixes https://github.com/esphome/issues/issues/6978

## Checklist:

    or


guides/yaml.rst
```diff
@@ -0,0 +1,298 @@
...
+
+In this example, both sensors share the metadata from `common_adc`, but the second sensor overrides the `pin` and `name` values.
+
+ESPHome YAML Extensions
+-----------------------
+
+ESPHome adds several powerful features to standard YAML:
+
+Secrets and the secrets.yaml File
+---------------------------------
```
kbx81: I think the headings below here (except for "See Also") should be sub-headings since they're ESPHome-specific extensions.

```suggestion
Secrets and the secrets.yaml File
*********************************
```

guides/yaml.rst
```diff
@@ -0,0 +1,298 @@
...
+
+    # The contents of common/wifi.yaml
+
+    wifi:
+      ssid: "MyWiFi"
+      password: !secret wifi_password
+
+Note that the key in a packages line is just a placeholder - it must be unique within the `packages` mapping, and should ideally be chosen to indicate its purpose, but otherwise can be anything you like that is a valid key.
+
+Variable can be provided for substitutions when including a package file, just as for regular includes.
```
kbx81:
```suggestion
Variables can be provided for substitutions when including a package file, just as for regular includes.
```

guides/yaml.rst
```diff
@@ -0,0 +1,298 @@
...
+
+The `packages:` key can only appear at the root level of the YAML file, i.e. it must start in column 1. Its location within the file is however not important. Files included under a `packages:` key will be inserted at the root level of the file, so the contents of the file should look like fragments of a standard configuration. This is in contrast to regular includes, which are inserted at the same level as the `!include` key.
```
kbx81: I've been avoiding use of "e.g.", "i.e." and the like as it may be confusing for/unfamiliar to non-native English speakers. The HA folks don't use these and I think we should follow that lead.

```suggestion
The `packages:` key can only appear at the root level of the YAML file; in other words, it must start in column 1. Its location within the file is however not important. Files included under a `packages:` key will be inserted at the root level of the file, so the contents of the file should look like fragments of a standard configuration. This is in contrast to regular includes, which are inserted at the same level as the `!include` key.
```

guides/yaml.rst
```diff
@@ -0,0 +1,298 @@
...
+        on_multi_click: !include
+          # multi-line syntax
+          file: on-multi-click.yaml
+          vars:
+            id: 2
+
+Packages
+********
+
+The ``packages`` feature allows you to define reusable and potentially partial configurations that can be included in your main configuration. Including a package file will merge its contents with your main configuration in a non-destructive way.
```
nielsnl68:
```suggestion
The ``packages:`` feature allows you to define reusable and potentially partial configurations that can be included in your main configuration. Including a package file will merge its contents with your main configuration in a non-destructive way.
```

guides/yaml.rst
```diff
@@ -0,0 +1,298 @@
...
+
+    wifi:
+      ssid: "MyWiFi"
+      password: !secret wifi_password
+
+Note that the key in a packages line is just a placeholder - it must be unique within the `packages` mapping, and should ideally be chosen to indicate its purpose, but otherwise can be anything you like that is a valid key.
+
+Variables can be provided for substitutions when including a package file, just as for regular includes.
+
+The `packages:` key can only appear at the root level of the YAML file, in other words it must start in column 1. Its location within the file is however not important. Files included under a `packages:` key will be inserted at the root level of the file, so the contents of the file should look like fragments of a standard configuration. This is in contrast to regular includes, which are inserted at the same level as the `!include` key.
```
nielsnl68: I suggest to be make this paragraph more general because this is also true for the `substitutions` as well.

nielsnl68 (commented):
Just 2 small suggestions.

guides/yaml.rst
clydebarrow: This is explaining the difference between using `!include` in two different contexts. `!include` is not relevant to the `substitutions:` section.

guides/packages.rst
jesserockz, line 1: These files are specifically in the `components` directory because they are top level configuration entries and I don't think they should move.

guides/yaml.rst
```diff
@@ -0,0 +1,364 @@
...
+
+    - {
+        "platform": "gpio",
+        "name": "Temperature 1"
+      }
+    - {
+        "label": {
+          "text": "Temperature 1"
+        }
+      }
```
jesserockz: Either make it fully valid json
```suggestion
.. code-block:: json

    [
      {
        "platform": "gpio",
        "name": "Temperature 1"
      },
      {
        "label": {
          "text": "Temperature 1"
        }
      }
    ]
```

Or leave the block language as yaml...
```suggestion
.. code-block:: yaml

    - {
        "platform": "gpio",
        "name": "Temperature 1"
      }
    - {
        "label": {
          "text": "Temperature 1"
        }
      }
```

guides/yaml.rst
```diff
@@ -0,0 +1,364 @@
...
+    number:
+      - platform: template
+        <<: *AnchorNumber # Include the anchor previously defined
+        id: "SwitchMainDelay"
+        name: "Main Switch Delay"
+
+The hidden key name is not important, and indeed can be just a single dot, but using a more descriptive name is recommended.
+
+Lambdas
+^^^^^^^^^
```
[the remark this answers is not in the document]
jesserockz: RST allows for having equal or more underline character count than the heading itself, but does not allow less than the test.

(Trying to teach coderabbitai here)

jesserockz (changes requested):

guides/yaml.rst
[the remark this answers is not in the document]
clydebarrow: Also note that where `doc` references are used, the leading slash is often required since the source document may not be at the top of the tree.

clydebarrow:
@coderabbitai Your suggestion to remove the leading slash from doc references is wrong - those are needed in this context since the source document is not at the root of the tree, so an absolute path to the referenced document is required.

The suggestion to change the code block format from yaml to json is not appropriate, since although the code block uses json style, it's within a yaml document - yaml is a superset of json, but not all yaml is valid json.

guides/yaml.rst
```diff
@@ -0,0 +1,364 @@
...
+          vars:
+            id: 2
+
+Packages
+^^^^^^^^
+
+The ``packages:`` feature allows you to define reusable and potentially partial configurations that can be included in your main configuration.
+The data is merged with the main configuration, with values in the main configuration taking precedence over values in the package data.
+
+See :doc:`/components/packages` for more details.
```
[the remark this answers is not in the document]
clydebarrow: Your suggestion to remove the leading slash from doc references is wrong - those are needed in this context since the source document is not at the root of the tree, so an absolute path to the referenced document is required.

clydebarrow:
@coderabbitai 
> The link currently points to the component docs instead of the dedicated substitutions guide. Update it to reference the guide path:

There is no substitutions guide, the link is correct.

jesserockz (approved):

PR #4876, opened 2025-05-03
[lvgl] document lvgl.widget.refresh

clydebarrow:
## Description:

- esphome/esphome#8675

## Checklist:

    or


iBobik:
Hi, thank you for this change.

Could you please better clarify how it differs to lvgl.widget.redraw? From the docs I expected it does the same.

clydebarrow:
> From the docs I expected it does the same.

I'm puzzled why you would expect that when the descriptions are quite different:

 for `redraw`:

> This action redraws the entire screen, or optionally only a widget on it.

For `refresh`:

> This action re-evaluates any templated properties set on the specified widget.

iBobik:
Aha, so jist redraw will not reevaluate?

reevaluate will also redraw when needed?

clydebarrow:
Yes. Any change to a widget property will trigger a redraw, as you would expect. The only common use for the explicit redraw action is after LVGL has been paused.

jesserockz (approved):

PR #4927, opened 2025-05-21
Update bl0942.rst

Renegade605:
## Description:
Additional detail on calibrating the voltage and current reading of the BL0942.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):**  N/A

## Checklist:

    or


components/sensor/bl0942.rst
```diff
@@ -60,7 +60,9 @@ Configuration variables:
...
+There are two fundamental calibration parameters which are dependent on the hardware: ``voltage_reference`` and ``current_reference``.  These can be determined by using an accurate voltage and current meter with a simple resistive load. Higher current loads will give more resolution to the measurement than smaller ones. Consider the largest load your device can support for this procedure. Example resistive loads to consider: space heater, toaster, incandescent light bulb(s), etc.
+
+The default value of ``voltage_reference`` is ``15873.35944299`` and for ``current_reference`` it is ``251213.46469622``. Flash the device with these values and then measure the voltage and current of the load while comparing it to the value reported in ESPHome. The formula to determine the correct reference value is ``new_reference = reported_value * default_reference / actual_value``. (If the reported value is low, the reference value should decrease, and vice versa.) This procedure can be repeated several times to hone in on the correct calibration. *See the comment in* ``bl0942.h`` *(../api/bl0942_8h_source) for more details.*
```
DjordjeMandic: Could you link it?

Renegade605: Oh weird. It was a link to https://api-docs.esphome.io/bl0942_8h_source but the bot told me I had to change it to ../api/bl0942_8h_source. Apparently that doesn't work? 

Am I missing some syntax specific to .rst files to make it a link?

Renegade605: I think that's done it?

DjordjeMandic: Does not work still.


Try it like this:
```
:apiref:`bl0942/bl0942.h`
```

Renegade605: That doesn't go directly to the source file though. It goes to "https://api-docs.esphome.io/bl0942_8h.html" when the actual target wanted is "https://api-docs.esphome.io/bl0942_8h_source".

If it's the only option I guess it's better than nothing, but when I was trying to figure all this out it took me quite a while to find there were details in the source code that weren't anywhere else.

DjordjeMandic: Proper doxygen comment will show up on that page.

You need to format / make comment doxygen compatible. I'll look into it to suggest a fix.

Renegade605: Changed to `../../api/bl0942_8h_source` and I suspect that may not work in the preview but it should work in the real thing.

DjordjeMandic: Small suggestion regarding linking in the documentation: hard-linking directly to specific files or paths might not be ideal, especially since the apiref plugin exists for a reason, it helps keep links flexible when things like domains, URLs, or file structures change over time.

It might be better to either link to the relevant header file on GitHub or, even better, update the comment in the header and link to its apiref- doxygen generated documentation page. That way, things stay maintainable and easier to navigate in the long run.

Renegade605: I mean I could link to the GitHub. Or just dump the whole comment explaining it in a collapsible section. Or remove the link entirely and trust this is a good enough explanation of what to do that most people don't need to go learn it for themselves.

I'll leave it up to you. I just wanted to save someone else the hours of work I did figuring it out. Especially since the devices page didn't mention anything about having to calibrate the voltage and current (I updated that already too).

jesserockz: There is a very simple solution to this...

https://github.com/esphome/esphome-docs/pull/4954

jesserockz: Please also put some newlines in your text. We prefer that source code can be readable without having to scroll or have wrapping turned on.

Single newlines will not break up a paragraph when rendered

jesserockz (changes requested):
As per my comment above

components/sensor/bl0942.rst
```diff
@@ -60,7 +60,9 @@ Configuration variables:
...
+There are two fundamental calibration parameters which are dependent on the hardware: ``voltage_reference`` and ``current_reference``.  These can be determined by using an accurate voltage and current meter with a simple resistive load. Higher current loads will give more resolution to the measurement than smaller ones. Consider the largest load your device can support for this procedure. Example resistive loads to consider: space heater, toaster, incandescent light bulb(s), etc.
+
+The default value of ``voltage_reference`` is ``15873.35944299`` and for ``current_reference`` it is ``251213.46469622``. Flash the device with these values and then measure the voltage and current of the load while comparing it to the value reported in ESPHome. The formula to determine the correct reference value is ``new_reference = reported_value * default_reference / actual_value``. (If the reported value is low, the reference value should decrease, and vice versa.) This procedure can be repeated several times to hone in on the correct calibration. *See the comment in* `bl0942.h <../../api/bl0942_8h_source>`__ *for more details.*
```
DjordjeMandic:
```suggestion
The default value of ``voltage_reference`` is ``15873.35944299`` and for ``current_reference`` it is ``251213.46469622``. Flash the device with these values and then measure the voltage and current of the load while comparing it to the value reported in ESPHome. The formula to determine the correct reference value is ``new_reference = reported_value * default_reference / actual_value``. (If the reported value is low, the reference value should decrease, and vice versa.) This procedure can be repeated several times to hone in on the correct calibration. *See the comment in* :apisource:`source code <bl0942/bl0942.h>` *for more details.*
```

This should work now

Renegade605: Only thing is it seems like this should be
```
:apisource:`source code <bl0942/bl0942.h>`
```
right?

components/sensor/bl0942.rst
Renegade605: > Please also put some newlines in your text. We prefer that source code can be readable without having to scroll or have wrapping turned on.

You know, I could have sworn I did. Not sure what happened to them?

components/sensor/bl0942.rst
DjordjeMandic: right, sorry, my bad. had wrong stuff in clipboard :)

DjordjeMandic (approved):
LGTM

jesserockz (approved):

PR #4969, opened 2025-05-31
Add support for custom request headers in online_image component

numo68:
## Description:

The `http_request` component is able to add custom request headers to the request, but the `online_image` is not. As I want to fetch a pre-computed image from a server requiring a bearer API token, I have added this possibility.

- esphome/esphome#8985

## Checklist:

    or

### DIFF
--- a/components/online_image.rst
+++ b/components/online_image.rst
@@ -32,8 +32,6 @@ This component has a dependency to :doc:`/components/http_request`; the configur
 
     online_image:
       - url: "https://example.com/example.png"
-        request_headers:
-          Authorization: !secret bearer_api_token
         format: png
         id: my_online_image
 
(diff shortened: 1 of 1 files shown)


components/online_image.rst
```diff
@@ -32,13 +32,16 @@ This component has a dependency to :doc:`/components/http_request`; the configur
 
     online_image:
       - url: "https://example.com/example.png"
+        request_headers:
+          Authorization: !secret bearer_api_token
```
jesserockz: The main example should be kept to a minimum as people (for some reason) like to copy paste it directly to their YAML, and we would like them to read and understand the configuration options instead.
```suggestion
```

jesserockz (changes requested):

jesserockz (approved):

PR #4987, opened 2025-06-09
Update installing_esphome.rst to mention git for extern. components

shvmm:
## Description:
include a mention of installing git in notes for using external components
Also minor touchup of the overall guide
* Mentioned Python version bumped from 3.9 to 3.10
* ESPHome version mentioned bumped to 2025.5.2

**Related issue (if applicable):** fixes <link to issue> NONE

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** NONE

## Checklist:

    or

### DIFF
--- a/guides/installing_esphome.rst
+++ b/guides/installing_esphome.rst
@@ -42,7 +42,7 @@ with the following:
 
 
 .. note::
-   
+
     You may additionally need to install git for the external components feature.
     Download git from `the official link <https://git-scm.com/downloads>`_
 
--- a/guides/installing_esphome.rst
+++ b/guides/installing_esphome.rst
@@ -69,7 +69,7 @@ Verify the installation:
 .. code-block:: console
 
     $ esphome version
-    Version: 2024.12.0 
+    Version: 2025.5.2
 
 
 
@@ -151,7 +151,7 @@ At this point, you should be able to confirm that ESPHome has been successfully 
 .. code-block:: console
 
     $ esphome version
-    Version: 2022.11.4
+    Version: 2025.5.2
 
 If you get an error like "Command not found", you need to add the binary to
 your ``PATH`` using ``export PATH=$PATH:$HOME/.local/bin``.
--- a/guides/installing_esphome.rst
+++ b/guides/installing_esphome.rst
@@ -29,25 +29,25 @@ Looks good? You can go ahead and install ESPHome:
 
 .. code-block:: console
 
-    > pip3 install wheel
-    > pip3 install esphome
+    pip3 install wheel
+    pip3 install esphome
 
 And you should be good to go! You can test that things are properly installed
 with the following:
 
 .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


guides/installing_esphome.rst
```diff
@@ -44,7 +38,16 @@ with the following:
 .. code-block:: console
 
     > esphome version
-    Version: 2021.12.3
+    Version: 2025.5.2
+
+
+.. note::
+
+    - Don't copy the ``>``. That's used to show that this is a command that goes
```
glmnet: I guess > is ok, but more global "terminology" used in this kind of stuff is the $ sign to represent a user prompt and a # to represent a root prompt. So this might be a good time to change that with a $.

guides/installing_esphome.rst
```diff
@@ -44,7 +38,16 @@ with the following:
...
+      in the console, and to let you see what the expected results are (shown on
+      the next line without a ``>``)    
+    - You may additionally need to install git for the external components feature.
+      Download git from `the official website <https://git-scm.com/downloads/win>`_
```
glmnet: this is too windows, i guess user should figure out how to install git on his platform

glmnet (changes requested):

guides/installing_esphome.rst
shvmm: I figured it out as other platforms usually have git pre-installed.
Plus, this is under windows section of the guide

guides/installing_esphome.rst
shvmm: Thanks
Will update the respective win and Linux sections with their respective symbols. 

Don't have access to a Mac so that has to come from elsewhere

clydebarrow: The prompt symbols shouldn't even be there. Just put the command.

clydebarrow: The prompt symbols shouldn't even be there. Just put the command. It's extremely annoying when copying commands (especially when using an embedded copy button) to then have to remove unnecessary crud.

kbx81: Second what @clydebarrow said. haha

shvmm: I have chosen for a compromise.

Only for version number checking where terminal output is involved, I have used prompt symbol to distinguish the output from the command.

glmnet: That's not consistent 

I guess ultimately best will be the page allowing to copy commands. 
I still like the $ and # 
But if we remove prompt you better add a break in between saying "you should see..."

clydebarrow: Yes, split the command (in a directly copiable form) as a separate block to the expected output.

shvmm: > Yes, split the command (in a directly copiable form) as a separate block to the expected output.

Done. Resolving this

glmnet (approved):

PR #4988, opened 2025-06-10
Clarify MCP23017 GPIO indexing in ESPHome docs

viniciusvillas:
This update fixes the MCP23017 component documentation. Changes were made to the description of the pin mapping, clarifying that the A port is reverse-mapped. Changes were made in the example to reflect the way the component actually works (A0 is actually pin 7, because it's reverse-mapped)

## Description:

## Checklist:

    or

### DIFF
--- a/components/mcp230xx.rst
+++ b/components/mcp230xx.rst
@@ -147,7 +147,9 @@ and can therefore be used with many of ESPHome's components such as the GPIO
 binary sensor or GPIO switch.
 
 GPIO pins in the datasheet are labelled A0 to A7 and B0 to B7, these are mapped
-consecutively in this component to numbers from 0 to 15.
+consecutively in this component to numbers from 0 to 15. There are two parts to
+note here: port A is reverse-mapped (A7=0 to A0=7) and port B is forward-mapped
+(B0=8 to B7=15).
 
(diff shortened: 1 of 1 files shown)


clydebarrow:
Please provide some documentation to support the assertion that port A is reverse mapped.

viniciusvillas:
> Please provide some documentation to support the assertion that port A is reverse mapped.

I couldn’t find authoritative documentation for this behavior. My findings are empirical: in practice the current ESPHome documentation for the integration with the MCP23017 diverges from what the hardware actually does (the datasheet doesn’t cover ESPHome’s integration details, so this has to be inferred from real-world tests).

If there’s a canonical reference I’ve missed, I’d appreciate a pointer. If this change isn’t a fit for main, no problem—but the current documented behavior is incorrect in practice. At minimum, this PR can serve as a breadcrumb for future readers.

clydebarrow:
> I've tracked down this behavior in practice.

Well, could you detail your findings here then. The datasheet and the ESPHome implementation appear on paper to be in agreement, so if the datasheet is wrong you'll need to provide some evidence of that. 

If your assertion is correct then it's certainly important for others to know that, just need something to back it up.

PR #5002, opened 2025-06-17
Add interrupt support to GPIO binary sensors

bdraco:
## Description:

Add interrupt support to GPIO binary sensors

- esphome/esphome#9115

## Checklist:

    or


components/binary_sensor/gpio.rst
```diff
@@ -69,7 +118,13 @@ Debouncing Values
 -----------------
 
 Some binary sensors are a bit unstable and quickly transition between the ON and OFF state while
-they're pressed. To fix this and debounce the signal, use the :ref:`binary sensor filters <binary_sensor-filters>`:
+they're pressed. To fix this and debounce the signal, use the :ref:`binary sensor filters <binary_sensor-filters>`.
+
+.. note::
+
+    When using interrupt mode, proper debouncing becomes even more important as every
```
nagyrobi: Interrupt mode by default could behave like a breaking change exactly because of this. I would suggest extending the note with a hint on debounce value, which would be approximately equivalent with the most usual main loop time (~20ms?). Also, multiple filters could be used to debounce, which one would be the recommended one for this situation (maybe `settle`?).

bdraco: I should probably remove that comment because it's still only being updated at the same rate in the loop. However I'm not sure I'm moving this forward unless I can come up with a safe way to reenable the loop from ISR

components/binary_sensor/gpio.rst
```diff
@@ -69,7 +119,8 @@ Debouncing Values
 -----------------
 
 Some binary sensors are a bit unstable and quickly transition between the ON and OFF state while
-they're pressed. To fix this and debounce the signal, use the :ref:`binary sensor filters <binary_sensor-filters>`:
+they're pressed. To fix this and debounce the signal, use the :ref:`binary sensor filters <binary_sensor-filters>`.
+
```
bdraco:
```suggestion
they're pressed. To fix this and debounce the signal, use the :ref:`binary sensor filters <binary_sensor-filters>`:
```

components/binary_sensor/gpio.rst
bdraco: Updated docs to reflect the current implementation is designed to be backwards compatible and only switches the state in the loop to be both backwards compat and ISR safe

jesserockz (approved):

PR #5041, opened 2025-06-29
[sen5x] Extend with SEN6X support esphome

mikelawrence:
## Description:

Add support SEN6X sensors.

- esphome/esphome#9254

## Checklist:

    or


chilipepperpizza:
It looks like some of the defaults may be different between the SEN5x and SEN6x devices?

In the example SEN66 config it has:
```yaml
    nox:
      name: "NOx"
      algorithm_tuning:
        index_offset: 100
        ...
        gating_max_duration_minutes: 180
```

But under the nox description below it says:
> - index_offset (Optional, integer): NOx index representing typical (average) conditions. Allowed values are in range 1..250. The default value is **1**.
> ...
> - gating_max_duration_minutes (Optional, integer): Maximum duration of gating in minutes (freeze of estimator during high NOx index signal). Zero disables the gating. Allowed values are in range 0..3000. The default value is **720** minutes.

Are the defaults different between these two or is the SEN66 intentionally being set to something other than the default?
Just wondering about the discrepancy of 1 vs 100 for index_offset and 180 vs 720 for gating_max_duration_minutes.

mikelawrence:
> It looks like some of the defaults may be different between the SEN5x and SEN6x devices?
> 
> In the example SEN66 config it has:
> 
> ```yaml
>     nox:
>       name: "NOx"
>       algorithm_tuning:
>         index_offset: 100
>         ...
>         gating_max_duration_minutes: 180
> ```
> 
> But under the nox description below it says:
> 
> > * index_offset (Optional, integer): NOx index representing typical (average) conditions. Allowed values are in range 1..250. The default value is **1**.
> >   ...
> > * gating_max_duration_minutes (Optional, integer): Maximum duration of gating in minutes (freeze of estimator during high NOx index signal). Zero disables the gating. Allowed values are in range 0..3000. The default value is **720** minutes.
> 
> Are the defaults different between these two or is the SEN66 intentionally being set to something other than the default? Just wondering about the discrepancy of 1 vs 100 for index_offset and 180 vs 720 for gating_max_duration_minutes.

I think I just made a mistake. The latest documentation doesn't even have these shown in the example. I believe the SEN5X and SEN6X sensors have the same defaults. VOC defaults are different from the NOx defaults and that wasn't handled correctly in the original sen5x component. There is another PR that addresses the default issue with VOC and NOx.  It was just merged so I'm updating this PR now.

There has been a significant change to the documentation format too. No more reStructuredText (rst) files. Now it is Markdown (md).

mikelawrence:
I have updated the documentation to use the new Markdown structure and changes from PR #5176.

content/components/sensor/sen5x.md
```diff
@@ -186,23 +235,140 @@ to blow out the accumulated dust inside the fan.
...
   every week if the sensor is switched off and on periodically (e.g., once per day).
-- The cleaning procedure can also be started manually with the `start_autoclean_fan` Action
+- The cleaning procedure can also be started manually with the `start_autoclean_fan` Action.
 
-The Sen5x sensor has an automatic fan-cleaning which will accelerate the built-in fan to maximum speed for 10
-seconds in order to blow out the dust accumulated inside the fan.
-The default automatic-cleaning interval is 168 hours (1 week) of uninterrupted use. Switching off the sensor resets
-this time counter.
+SEN6X sensors supports fan cleaning but not the automatic fan cleaning interval. You have to trigger
+`sen5x.start_fan_autoclean` action below occasionally.
```
danielkent-net: Per https://github.com/esphome/esphome/pull/9254#issuecomment-3304145174, I think it would be a good idea to provide an example of how to trigger SEN6x autocleaning periodically in esphome. Suggestion:

```suggestion
`sen5x.start_fan_autoclean` action below occasionally. One way this can be implemented within ESPHome is by adding an `interval` component to perform this action periodically. For example, to clean the fan every 7 days while the device is on, as recommended by the manufacturer, the following configuration can be added:
    
    interval:
        - interval: 7d
          then:
            - sen5x.start_fan_autoclean: sen66_01
```

mikelawrence: This is a good idea! I've added it.

content/components/sensor/sen5x.md
```diff
@@ -1,217 +1,478 @@
...
+          value: !lambda "return id(co2_forced_cal_value).state;"
+          id: sen66_sensor
+sensor:
+  - platform: sen5x
+    type: SEN66
+    id: sen66_sensor
+    co2:
+      name: "CO₂"
+```
+
```
danielkent-net, line 364: I think it would also be a good idea to also provide an example using a [user-defined action](https://esphome.io/components/api/#user-defined-actions). Something like this:

```yaml
api:
  encryption:
    key: !secret encryption_key
  actions:
    - action: sen69c_recalibrate_co2
      variables:
        co2: int
      then:
        - if:
            condition:
              lambda: 'return (co2 >= 400) and (co2 <= 1200);'
            then:
              - sen5x.perform_forced_co2_recalibration:
                  value: !lambda 'return co2;'
                  id: sen69c
              - api.respond:
                  success: true
            else:
              - api.respond:
                  success: false
                  error_message: "CO₂ input calibration must be between 400 and 1200."
```

You can test this added action by going into the developer console and searching for the action by name.

mikelawrence: This is a great idea but I'm away for a few days. I will get to this by the end of the week.

mikelawrence:
Devs have decided to go with another PR

PR #5085, opened 2025-07-10
Add note about TMC stepper drivers

ggodlewski:
## Description:

This remark would save me 2 hours of troubleshooting.

## Checklist:


DjordjeMandic (approved):
Thanks!

PR #5115, opened 2025-07-17
esp32_camera: Remove remaining references to i2c_pins-option from examples

RubenKelevra:
## Description:

Small fix for the examples in ESP32_Camera to all show the new `i2c:`-block + `i2c_id:` option in ESP32_Camera, instead of the now undocumented `i2c_pins:` option.

**Related issue (if applicable):** discussed in [#9601](https://github.com/esphome/esphome/issues/9601)

## Checklist:

    or


components/esp32_camera.rst
```diff
@@ -489,13 +489,15 @@ Configuration examples
...
 
+    i2c:
+      - id: camera_i2c
+        sda: GPIO4
+        scl: GPIO5
     esp32_camera:
       external_clock:
         pin: GPIO15
         frequency: 20MHz
-      i2c_pins:
```
jesserockz, line 496: This one was actually left in on purpose due to the external component
<https://github.com/MichaKersloot/esphome_custom_components/issues/2>

RubenKelevra: Ah okay, then I revert that part. :)

jesserockz: If you want to implement the changes as specified in that issue, I'm sure everyone will be thankful =)

RubenKelevra: See [#9630](https://github.com/esphome/esphome/pull/9630)

I reverted the removal in this case and added the new config option instead of the external component. :)

jesserockz (commented):
Please add the entry to the configuration variables list =)

RubenKelevra:
@jesserockz wrote:
> Please add the entry to the configuration variables list =)

I've added a minimal documentation to the next branch:

#5129 

Maybe @MichaKersloot can add some details on when to use? :)

jesserockz:
I think it's only when the board doesn't have psram

RubenKelevra:
Okay, but why then export this as option? Just detect that there's no psram and use dram in this case... or am I missing something?

MichaKersloot:
No you do not miss something. To me it looks like there is some detection for PSRAM implemented somewhere (there is a log entry at boot for that). But somehow it does not work for all boards, in my case the TTGO T-journal. So an option to override this was needed for me.  

There could be a better fix at a lower level, but that is way beyond my capabilities and this option at least makes the esp32_camera module useable for me now.

I tried to find the original discussion that leaded to this fix, but was unable to find it.

RubenKelevra:
Yeah I did a lot of fixing in the esp32-camera library in the last couple of days, I'll just have a look why it may fail.

I think I spotted the detection somewhere.

Can I mention you for testing with hardware? Because my hardware all got PSRAM.

RubenKelevra:
Let's suspend the PR for this time.

MichaKersloot:
I'll be on holydays for a few weeks, so testing would not be possible for a while. Further the camera is in use, so switching between Home Assistent and current master releases of ESPHome is not that easy for me. That's the reason why my patch only applies to official releases and not to the current master. 

Postponing this fix would mean ESPHome would not be useable for a couple of boards without my unofficial fix for the time being.

RubenKelevra:
Hey @MichaKersloot I'm not postponing a fix, I just think it should be automatic instead of a user decision.

I'll have a look if there's a detection and maybe it's just broken in the esp32-camera lib, when I'm home. I'm on the go right now.

If that's not possible, it's maybe possible to do this in ESPHome itself.

MichaKersloot:
Hmm.. i've read my previous comment and I see it looks a bit grumpy. Sorry! Thank you for your time and effort to fix this. I'll try to make some time next week for this.

RubenKelevra:
No worries!

Enjoy your free time and just let me know if you got *really* the head and mind free to look at this. :)

I'll just mention you on everything I'll find about that topic.

RubenKelevra:
@jesserockz alright, after reading the esp32-camera lib source code on that matter:

If we don't specify anything, the default would be PSRAM and there's no detection on steering this decision, not "auto" mode, no fallback.

So I was misremembering things - sorry.

So my suggestion would be to implement this setting like it's currently proposed, and then add in the future a third value, `auto` - if we implement some kind of automatic detection, which would try PSRAM first and if there's no PSRAM, it would fall back.

RubenKelevra:
I'll resubmit this changes for `next`, as they introduce the option which was just merged.

PR #5157, opened 2025-07-22
Allow to disable the phase balancer for PCA9685

jonau:
## Description:

The PCA9685 chip allows for flexible phase angles. Currently the component uses this ability to set the phase angles equidistantly for all output channels. So 3 output channels would have a phase angle of 0°, 120° and 240°. This however can cause flickering of the LEDs when animating / fading, because when the output changes and the start angle is higher than the end angle of the output the chip will need an additional frame where no signal is outputted (as indicated by the (datasheet)[https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf#page=20]

This pull requests documents a new configuration option `phase_balancer` with the options of `linear`, which keeps the current behavior and is the default. And `none` which sets all phase angles to 0°, so that the problem described above does not happen.

- esphome/esphome#9792

## Checklist:

    or


swoboda1337:
Docs need to be updated from rst to md

jesserockz (changes requested):
As above, please rebase the docs. I tried but it wont let me push to your fork.

Update: I just noticed it is a fork in an org which for some reason GitHub doesn't allow pushes by maintainers.

PR #5169, opened 2025-07-24
[esp32] Document LWIP core locking options for improved socket performance

bdraco:
## Description:

This PR documents the new LWIP TCP/IP core locking options added to the ESP32 component's advanced configuration. These options enable significant socket performance improvements on ESP-IDF by using direct function calls with mutex protection instead of mailbox message passing.

The documentation explains:
- What the options do and their performance impact (20-200% improvement)
- Default values (both enabled by default)
- When users might want to disable them (uncommon, since ESPHome uses an event loop)
- Example configuration showing how to use the options

- esphome/esphome#9857

## Checklist:

    or


bdraco:
next!

jesserockz (approved):

PR #5172, opened 2025-07-24
[waveshare_epaper] Add compatible display type/board manufacturer

soundstorm:
## Description:
As previously trying to get it running with other types (after switching from one which was compatible with 4.20in def), a clarification on what display is used, is helpful and speeds up the tinkering. Seems to be a common board.

## Checklist:

    or


swoboda1337:
Conflicts

soundstorm:
Rebased and re-applied

PR #5234, opened 2025-08-12
add documentation for the advanced configuration option 'enable_lwip_tcp_sack'

RubenKelevra:
## Description:

add documentation for the advanced configuration option 'enable_lwip_tcp_sack'

- esphome/esphome#10211

## Checklist:

    or

### DIFF
--- a/components/esp32.rst
+++ b/components/esp32.rst
@@ -161,6 +161,8 @@ LWIP (Lightweight IP) behavior. Some options improve performance while others sa
   improves socket operation performance by 20-200% but may reduce multi-threaded scalability. Defaults to ``true``.
 - **enable_lwip_check_thread_safety** (*Optional*, boolean): Enable LWIP thread safety checks to detect incorrect usage of
   the TCP/IP stack from multiple threads. This helps catch thread safety issues when core locking is enabled. Defaults to ``true``.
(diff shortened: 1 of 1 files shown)


components/esp32.rst
```diff
@@ -161,6 +161,8 @@ LWIP (Lightweight IP) behavior. Some options improve performance while others sa
   improves socket operation performance by 20-200% but may reduce multi-threaded scalability. Defaults to ``true``.
 - **enable_lwip_check_thread_safety** (*Optional*, boolean): Enable LWIP thread safety checks to detect incorrect usage of
   the TCP/IP stack from multiple threads. This helps catch thread safety issues when core locking is enabled. Defaults to ``true``.
+- **enable_lwip_tcp_sack** (*Optional*, boolean): Enable TCP selective acknowledgments
+  to improve performance on lossy or high latency networks. Defaults to ``true``.
```
kepstin, line 165: The documentation should mention that the option is framework-dependent (only applies when using esp-idf)

RubenKelevra: Ah, yeah, sorry forgot that.

The other lwip options should get the same treatment, though.

kepstin: Actually, never mind. I was missing that context. The header for the entire LWIP options section already says "LWIP Optimization Options (ESP-IDF only):" so there is no need to add notes on the specific options.

PR #5258, opened 2025-08-20
remove throttle from ld2450 example config

fabio-garavini:
## Description:

Fix `ld2450` sensor example config removing `throttle` option as it has been removed in esphome version 2025.8.0

## Checklist:

    or


fabio-garavini:
As `throttle` option has been remove in
- esphome/esphome#10019

PR #5332, opened 2025-09-04
Update the microphone configuration example for the M5Stack Atom Echo

hollie:
## Description:

**Pull request in [esphome](https://github.com/esphome/esphome) to update the documentation with a working YAML configuration for the microphone on the M5Stack Atom Echo as the pin configuration was missing in the current documentation.

I also added a link to a gist with a complete minimal example on how to use the microphone to report sound level using the M5Stack Atom Echo. 

## Checklist:

    or


content/components/microphone/i2s_audio.md
```diff
@@ -75,10 +75,13 @@ microphone:
 microphone:
   - platform: i2s_audio
     adc_type: external
-    i2s_din_pin: GPIOXX
-    pdm: true
+    i2s_din_pin: GPIO23
```
ximex, line 78: revert at least the GPIO pin. see:
> Pin numbers used in examples should be the string GPIOXX -- not a specific pin number.

https://developers.esphome.io/contributing/docs/#documentation-guidelines

cat say something to the rest

ximex (changes requested):

PR #5333, opened 2025-09-04
Add example config for M5Stack Atom Echo sound level meter

hollie:
Adds a simple link to the gist with the example configuration for a M5Stack Atom Echo based sound level meter.


swoboda1337:
This would probably make more sense to put here: https://devices.esphome.io/

PR #5352, opened 2025-09-12
[uart] Multiple ESP32 features and fixes

brambo123:
## Description:

**Related issue (if applicable):** Multiple ESP32 features and fixes

- esphome/esphome#8103
## Checklist:

    or


content/components/uart.md
```diff
@@ -65,10 +65,17 @@ uart:
 - **rx_pin** (*Optional*, [Pin](#config-pin)): The pin to receive data on from the ESP's perspective. Use the full pin
   schema and set `inverted: true` to invert logic levels. Not supported by host platform.
 
+- **flow_control_pin** (*Optional*, [Pin](#config-pin)): ESP32 only. The pin used to for hardware RS485 flow control.
+  When using this setting, half-duplex mode is enabled. Use the full pin schema and set `inverted: true` to invert logic levels.
```
kbx81:
```suggestion
  Use of this setting enables half-duplex mode. Use the full pin schema and set `inverted: true` to invert logic levels.
```

content/components/uart.md
```diff
@@ -65,10 +65,17 @@ uart:
...
+
 - **port** (*Optional*, string): Host platform only. Unix style name of the port to use.
 - **rx_buffer_size** (*Optional*, int): The size of the buffer used for receiving UART messages. Increase if you use an
   integration that needs to read big payloads from UART. Defaults to `256`.
 
+- **rx_full_threshold** (*Optional*, int): ESP32 only. After this many bytes have been received, the data becomes available for processing.
+  Defaults calculate during compilation to be approximately 10ms (about 8 bytes at 9600 baud, 114 bytes at 115200 baud).
+  
+- **rx_timeout** (*Optional*, int): ESP32 only. After this amount time of no data been received (in bytes-time), the data becomes available for processing. Defaults to ``2``.
```
kbx81:
```suggestion
- **rx_timeout** (*Optional*, int): ESP32 only. This value specifies a number of bytes used to determine the duration of the timeout. The duration of the timeout is based on the amount of time it would take to transmit/receive this number of bytes. In other words, for a given value and baud rate, doubling the baud rate will halve the timeout value. After the timeout has elapsed, the data becomes available for processing. Defaults to ``2``.
```

content/components/uart.md
```diff
@@ -65,10 +65,17 @@ uart:
...
 
+- **flow_control_pin** (*Optional*, [Pin](#config-pin)): ESP32 only. The pin used to for hardware RS485 flow control.
+  When using this setting, half-duplex mode is enabled. Use the full pin schema and set `inverted: true` to invert logic levels.
+
 - **port** (*Optional*, string): Host platform only. Unix style name of the port to use.
 - **rx_buffer_size** (*Optional*, int): The size of the buffer used for receiving UART messages. Increase if you use an
   integration that needs to read big payloads from UART. Defaults to `256`.
 
+- **rx_full_threshold** (*Optional*, int): ESP32 only. After this many bytes have been received, the data becomes available for processing.
+  Defaults calculate during compilation to be approximately 10ms (about 8 bytes at 9600 baud, 114 bytes at 115200 baud).
```
kbx81:
```suggestion
- **rx_full_threshold** (*Optional*, int): ESP32 only. After receiving this number of bytes, the data becomes available for processing.
  The default is calculated at compilation time to be approximately ten milliseconds (about 8 bytes at 9600 baud, 114 bytes at 115200 baud).
```

content/components/uart.md
```diff
@@ -88,7 +95,7 @@ be accurate at higher baud rates.
 logger and leave others available. If you have configured the logger to use a different hardware UART, the pins
 used for hardware sharing change accordingly.
 
-The ESP32 has three UARTs. ESP32 lite variant chips (ESP32-C3, ESP32-S2, ESP32-S3, etc) may have fewer UARTs
+The ESP32 has three UARTs. ESP32 lite variant chips (ESP32-C3, ESP32-S2, etc) may have fewer UARTs
```
kbx81:
```suggestion
The ESP32 has three UARTs. ESP32 variants (ESP32-C3, ESP32-S2, etc.) may have fewer UARTs
```

kbx81 (changes requested):
Copy fixes, add some detail/clarification

content/components/uart.md
brambo123: Perhaps this entire line should be changed a bit.
The original ESP32 had three; newer variants sometimes have less, sometimes the same, but sometimes more.
The ESP32-P4 has a whopping six (five UARTs in the HP system and one low-power (LP)
UART).

content/components/uart.md
brambo123: I think it's a bit too much information, and conflicting.
It's not "transmit/receive", only receive.
And "doubling the baud rate will halve the timeout value" could be misread, the timeout value does not change, the time changes.

kbx81: Great, so this is a clear indication that I didn't understand it, either. 😅

> It's not "transmit/receive", only receive.

Easy to fix, although is the transmit speed not the same as the receive speed? I suppose it doesn't technically _have_ to be, but I don't think we support different speeds for Tx and Rx. Regardless, I'm fine with just "receive" here. 🙂

> And "doubling the baud rate will halve the timeout value" could be misread, the timeout value does not change, the time changes.

Does the amount of time not determine the duration of the timeout? Maybe I'm misunderstanding what's going on here.

kbx81: Maybe this?:

```suggestion
- **rx_timeout** (*Optional*, int): ESP32 only. This value specifies a number of bytes used to determine the duration of the timeout. The duration of the timeout is based on the amount of time it would take to receive this number of bytes. In other words, for a given value and baud rate, doubling the baud rate will halve the duration of the timeout. After the timeout has elapsed, the data becomes available for processing. Defaults to ``2``.
```

content/components/uart.md
kbx81: If we want to call out what has what, we should add a table below. But I'd rather just point to the Espressif docs and not have to maintain our own copy of this information.

My aim with this suggestion was to eliminate the words "lite" and "chips". Nothing more. 🙂

brambo123: I changed the line a bit 😉

kbx81: Looks good, thanks! 🍻

kbx81 (approved):
Thanks!

PR #5395, opened 2025-09-24
Fix the table with examples in Wireguard component

Chupaka:
## Description:

[Previously](https://raw.githubusercontent.com/esphome/esphome-docs/4d6c6291c2974a629d44bf220d3786be6f74c63f/content/components/wireguard.md) the table was a bit complex:

image

[Recent conversion to Markdown](https://github.com/esphome/esphome-docs/pull/5242) broke the table: no newlines, even some text was lost:

image

As there's no possibility to use `rowspan` in Markdown, converting the table to HTML manually:

image

## Checklist:

    or

### DIFF
--- a/content/components/wireguard.md
+++ b/content/components/wireguard.md
@@ -156,16 +156,16 @@ Incoming connections are not affected by `netmask`.
 
 Let's explain with some examples:
 
-| address      | netmask                      | allowed ips                                    | working outgoing connections                                                                                                               |
-| ------------ | ---------------------------- | ---------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
-| 172.16.0.100 | *omitted* or 255.255.255.255 | *omitted* or any other value                   | **none**, no routes are created                                                                                                            |
-|              | 255.255.255.0                | *omitted*                                      | only to `172.16.0.0/24`                                                                                                                    |
-|              |                              | - 172.16.0.0/24 - 192.168.0.0/24 - *any other* | and any other network will be outside `172.16.0.0/24`                                                                                      |
-|              |                              | - 192.168.0.0/24                               | **none** because `192.168.0.0/24` is not part of `172.16.0.0/24`                                                                           |
-| 10.44.0.100  | 255.0.0.0                    | *omitted*                                      | to `10.0.0.0/8` network                                                                                                                    |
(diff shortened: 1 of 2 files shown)


Chupaka:
Huh... HTML [is allowed in Hugo](https://github.com/esphome/esphome-docs/blob/13673917d0fe5dd29c2a4e0e3cc17a2cbc15a58d/hugo.yaml#L137) but is forbidden in linter... What's the use case?..

Chupaka:
@clydebarrow could you check this as the author of the ["breaking change"](https://github.com/esphome/esphome-docs/pull/5242)?

clydebarrow:
No, we don't want embedded HTML. Since rowspan is not available in markdown, just add blank cells to simulate spanned rows. It's not ideal, but it gets the job done.

Which is what happened in the conversion, just looks like some text got lost which should be restored. Combine the two cells for "omitted" and "176.16..."

Chupaka:
> looks like some text got lost which should be restored

Not only that. What to do with yaml lists ("allowed ips" column)? We need
```
- 172.16.0.0/24
- 192.168.0.0/24
```
Not just `- 172.16.0.0/24 - 192.168.0.0/24`

I believe, `- 172.16.0.0/24 <br/>- 192.168.0.0/24` is forbidden too...

clydebarrow:
You can insert line breaks inside table cells with `<br>`

Chupaka:
I updated the PR, but neither `<br/>` nor `<br>` tags are allowed...

Also, please pick one - probably blank cells are less informative:

image
VS
image

clydebarrow:
> I updated the PR, but neither `<br/>` nor `<br>` tags are allowed...

I added a break shortcode. There is also the possibility to add a table render hook that could implement row and column spans, but that's a bit more work.

Re the two alternatives, probably repeating the same text is better than a blank cell.

Chupaka:
> probably repeating the same text is better than a blank cell.

I like this better too. So, looks like we're done, thanks for your assistance!

clydebarrow:
Not a big deal in this case, but force-pushing to a PR after review has started is generally not a Good Idea - it means the reviewer has to start over again since the history is lost. All commits get squashed when merged so there is no problem with having lots of small commits.

clydebarrow (approved):

PR #5520, opened 2025-10-24
Clarrify use of iteration for the repeat action

Aperion:
## Description:
More clear language for accessing the `iteration` parameter within a lambda for the repeat action. Also update example to demonstrate how to use iteration.

## Checklist:

    or


Aperion:
I don't think I created this patch commit correctly

PR #5527, opened 2025-10-26
[automations] Add anchor and fix regex for anchors

clydebarrow:
## Description:

**Related issue (if applicable):** fixes https://discord.com/channels/429907082951524364/930955987584684102/1431418211714207904

And is a simpler alternative to #5490 

## Checklist:

    or

<details>

1. Comment on this pull request with the following command, replacing `COMPONENT_NAME` with your component name in **UPPER_CASE** format with **underscores** (e.g., `BME280`, `SHT3X`, `DALLAS_TEMP`):

   ```
   @esphomebot generate image COMPONENT_NAME
   ```

3. Extract the SVG file and place it in the `images/` folder of this repository.

4. Use the image in your component's index table entry in `/components/index.rst`.

```
@esphomebot generate image DHT22
```

</details>

### DIFF
--- a/script/md_anchors.py
+++ b/script/md_anchors.py
@@ -32,7 +32,7 @@ def process_markdown_file(file_path):
             continue
 
         # Shortcodes
-        shortcode_match = re.search(r'{{<\s*anchor\s*"([^"]+)"\s*>}}', line)
+        shortcode_match = re.search(r'\{\{<\s*anchor\s+"([^\s>]+)"\s*>}}', line)
         if shortcode_match:
             anchor_id = shortcode_match.group(1)
             # Look ahead for heading
--- a/content/automations/_index.md
+++ b/content/automations/_index.md
@@ -7,6 +7,8 @@ params:
     image: auto-fix.svg
 ---
 
+{{< anchor "automation" >}}
+
 Automations are a very powerful aspect of ESPHome; they allow you to easily perform actions given some condition(s).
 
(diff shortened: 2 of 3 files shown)


script/md_anchors.py
```diff
@@ -32,7 +32,7 @@ def process_markdown_file(file_path):
             continue
 
         # Shortcodes
-        shortcode_match = re.search(r'{{<\s*anchor\s*"([^"]+)"\s*>}}', line)
+        shortcode_match = re.search(r'\{\{<\s*anchor\s+"([^\s>]+)"\s*>}}', line)
```
swoboda1337, line 35: If the opening braced brackets are escaped shouldn't the closing ones be too? It probably doesn't change anything but would be consistent.

clydebarrow: That would imply that `}` is special, which it's not, on its own. The opening brace is interpreted as part of a count or range like `{1-5}` if not escaped in a Python RE.

Escaping characters that don't need it can render the regex invalid or cause unexpected behaviour. In this case it would *probably* be harmless with the Python RE engine, but would be invalid when used with some other engines.

Indeed in "standard" RE format, `{` isn't special and must be escaped to form part of a counting expression. Which is my excuse for getting this wrong in the first place (though I suspect that was written by an AI - I don't remember now.)

swoboda1337 (approved):

PR #5719, opened 2025-12-04
Document `this` built-in config var

jpeletier:
## Description:

Document new built-in substitution var `this` that contains paths to the current YAML file and folder.

- esphome/esphome#12273

## Checklist:


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `!include` scoping example is not a valid ESPHome config

PR #5817, opened 2025-12-23
[mqtt.cover] Adds `mqtt_use_json_payload` option to `cover`.

heythisisnate:
## Description:

- esphome/esphome#12639

## Checklist:

    or

### DIFF
--- a/content/components/cover/_index.md
+++ b/content/components/cover/_index.md
@@ -58,9 +58,9 @@ Advanced options:
 
 MQTT options:
 
-- **mqtt_state_format** (*Optional*, enum): Either `values` or `json`. When set to `json`, state changes will
-  be only published to the `state_topic` as a JSON object. When `values`, individual values are published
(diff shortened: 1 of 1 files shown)


bdraco:
Hey @heythisisnate, a few things on this:

1. The option name doesn't match the code — the docs say `mqtt_use_json_payload` but the code PR uses `mqtt_json_state_payload`. This needs to match or users will get validation errors.

2. Worth mentioning that `position_state_topic` and `tilt_state_topic` can't be used when this is enabled (the code rejects it). Otherwise users will hit a confusing error without knowing why.

3. It'd be nice to show what the JSON payload actually looks like, something like:
   ```json
   {"state": "open", "position": 100, "tilt": 50}
   ```
   Especially since the stated use case is metered brokers where folks might be doing their own integration rather than relying on HA discovery.

bdraco (approved):
Thanks @heythisisnate

PR #5860, opened 2026-01-05
[faq] Improve "Which ESP should I use" recommendations

bdraco:
## Description:

Updates the "Which ESP should I use for my project?" FAQ section to better reflect the current state of ESPHome platform support in 2026. This aligns the documentation with an upcoming blog post.

Key changes:

**ESP8266 guidance updated:**
- Changed "Not Recommended" to "Not Recommended for New Projects"
- Added clear messaging that existing ESP8266 devices work well and will continue to be supported for years to come
- Noted that recent memory optimizations have significantly improved ESP8266 stability
- Retained concrete reasons why ESP32-C3 is preferred for new projects (5x RAM, more flash, more GPIO)

**ESP32-C6 added to Recommended:**
- Moved from "less mature" to recommended section
- Notes Thread (via OpenThread) and Zigbee radio support

**Improved chip descriptions:**
- ESP32 classic: Added note about built-in Ethernet MAC for lowest latency wired connections
- ESP32-S3: Added note about best raw compute power, better overall performance and lower latency
- ESP32-C3: Replaced vague "too powerful" language with concrete technical details (single-core RISC-V, lower cost/power consumption)

**Added LibreTiny section:**
- New section for BK72xx, RTL87xx, LN882x chips
- Clarifies these are for flashing existing devices, not purchasing new chips
- Notes support has matured but is not as mature as ESP8266/ESP32

**Cleaned up Additional Considerations:**
- Removed redundant content now covered in ESP8266 section
- Simplified "Other Microcontrollers" section
- Removed RP2350 (not currently supported)

## Checklist:

    or

### DIFF
--- a/content/guides/faq.md
+++ b/content/guides/faq.md
@@ -9,82 +9,73 @@ params:
 
 ## Which ESP should I use for my project?
 
-We're asked this *all the time.* As with all things engineering, "it depends". Based on the current state of hardware
-support within ESPHome, here's what we suggest:
+We're asked this *all the time.* As with all things engineering, "it depends". ESP32 is the leading platform where
+most new development happens. Based on the current state of hardware support within ESPHome, here's what we suggest:
 
 ### Recommended
 
-- **ESP32**
+- **ESP32 classic**
 
-  - Best supported/most mature
+  - Best supported/most mature.
   - Includes a great set of built-in hardware peripherals, so it's very capable and very flexible.
+  - Best choice for wired Ethernet connections with lowest latency thanks to its built-in Ethernet MAC,
+    which connects directly to a PHY chip. Other variants require SPI-based Ethernet controllers, adding latency.
 
 - **ESP32-S3**
 
   - An update to the original ESP32 with a slightly modified set of hardware peripherals.
-  - Has a built-in USB peripheral/interface (as opposed to relying on an external USB-to-serial chip)
+  - Has a built-in USB peripheral/interface (as opposed to relying on an external USB-to-serial chip).
   - Has instruction set extensions which make it a better fit for applications which require some form of machine
     learning ({{< docref "/components/micro_wake_word" >}}, for example).
+  - Best choice if you need raw compute power, with better overall performance and lower latency.
 
 - **ESP32-C3**
 
   - Generally intended ([per Espressif](https://www.espressif.com/en/news/ESP32_C3)) to replace the well-known ESP8266.
-  - Use if:
+  - A good choice for simpler projects: single-core RISC-V, fewer GPIO pins, but lower cost and power consumption
+    than the dual-core ESP32(-Sx) variants.
 
-    - You're worried that the ESP32(-Sx) is "too powerful".
-    - You need a lower-power device than the ESP32(-Sx) family offers.
+- **ESP32-C6**
 
-### Not Recommended
+  - Similar to the ESP32-C3 with additional connectivity options including Thread (via OpenThread) and a Zigbee radio.
+  - A good choice for new projects.
+
+### Not Recommended for New Projects
 
 - **ESP8266**
 
-  - It's over ten years old and is *quite lacking* in terms of built-in hardware peripherals.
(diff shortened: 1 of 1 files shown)


content/guides/faq.md
```diff
@@ -9,82 +9,73 @@ params:
...
 
-We're asked this *all the time.* As with all things engineering, "it depends". Based on the current state of hardware
-support within ESPHome, here's what we suggest:
+We're asked this *all the time.* As with all things engineering, "it depends". ESP32 is the leading platform where
+most new development happens. Based on the current state of hardware support within ESPHome, here's what we suggest:
 
 ### Recommended
 
-- **ESP32**
+- **ESP32 classic**
```
kbx81: Not sure how I feel about introducing "classic" to the nomenclature. I completely understand the intent (and even like the clarification it adds), but it's not a term I've seen used much elsewhere. There's already some confusion from board makers selling things like "ESP32S" (which is somewhat ambiguous) and other custom labeling which makes the water muddy. I might suggest this:

```suggestion
- **ESP32 (original, non-variant)**
```

content/guides/faq.md
```diff
@@ -9,82 +9,73 @@ params:
...
 
 - **ESP32-S3**
 
   - An update to the original ESP32 with a slightly modified set of hardware peripherals.
-  - Has a built-in USB peripheral/interface (as opposed to relying on an external USB-to-serial chip)
+  - Has a built-in USB peripheral/interface (as opposed to relying on an external USB-to-serial chip).
   - Has instruction set extensions which make it a better fit for applications which require some form of machine
-    learning ({{< docref "/components/micro_wake_word" >}}, for example).
+    learning ([Micro Wake Word](/components/micro_wake_word), for example).
+  - Best choice if you need raw compute power, with better overall performance and lower latency.
```
kbx81, line 30: Do we have a reference for this? As I understand it, it's the same cores as the original ESP32, just with a modified set of peripherals and the instruction set extensions (mentioned above).

content/guides/faq.md
```diff
@@ -9,82 +9,73 @@ params:
...
-  - Use if:
+  - A good choice for simpler projects: single-core RISC-V, fewer GPIO pins, but lower cost and power consumption
+    than the dual-core ESP32(-Sx) variants.
 
-    - You're worried that the ESP32(-Sx) is "too powerful".
-    - You need a lower-power device than the ESP32(-Sx) family offers.
+- **ESP32-C6**
 
-### Not Recommended
+  - Similar to the ESP32-C3 with additional connectivity options including Thread (via OpenThread) and a Zigbee radio.
```
kbx81: To avoid confusion, before we supported OpenThread, we'd omitted it (and Zigbee) because it wasn't supported by ESPHome.

Zigbee currently isn't supported; we can add it later if it gets added. 🙂
```suggestion
  - Similar to the ESP32-C3 with additional connectivity options including Thread (via OpenThread).
```

kbx81 (commented):
A few thoughts 🙂

content/guides/faq.md
kbx81: Hmm, I see it's LX6 vs. LX7, not sure how much raw improvement that alone might add. haha

bdraco: Right LX6 vs LX7 and more modern arch. OTA performance is a good measure as well with the S3 doing an OTA almost 70% faster than classic in my testing.

https://done.land/components/microcontroller/families/esp/esp32/developmentboards/esp32-s3/#math-coprocessor

Espressif publishes CoreMark scores in the datasheets:

Classic ESP32 (dual core at 240 MHz)
CoreMark 1079.96, CoreMark per MHz 4.50. 
[Espressif Documentation](https://documentation.espressif.com/esp32_datasheet_en.html)

ESP32 S3 (dual core at 240 MHz)
CoreMark 1329.92, CoreMark per MHz 5.54. 
[Espressif Documentation](https://documentation.espressif.com/esp32-s3_datasheet_en.html)

kbx81: Nice! Thanks 🍻

kbx81 (approved):
Thanks again! 👍🏻

PR #5936, opened 2026-01-17
docs(api): Enhance API documentation with lambda examples

Easen:
## Description:

Added examples for lambda equivalents of API connection checks.

## Checklist:

### DIFF
--- a/src/content/docs/components/api.mdx
+++ b/src/content/docs/components/api.mdx
@@ -436,7 +436,7 @@ on_boot:
       event: esphome.device_booted
 ```
 
-The lambda equivalent for this is `id(api_id).is_connected(true)`, e.g.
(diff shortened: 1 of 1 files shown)


src/content/docs/components/api.mdx
```diff
@@ -410,7 +410,18 @@ on_...:
...
+
+```yaml
+api:
+  id: api_id # Required to register "api_id"
+
+script:
+ - id: ...
+    then:
+      - lambda: |-
+          if (id(api_id).is_connected()) ...
```
jesserockz, line 423: Can do this without the id
```suggestion
          if (api::global_api_server->is_connected()) ...
```

Easen: If I’m really honest, the current documentation really confused me as I didn’t know where “api_id” came from. I thought it was a global variable, like this. 

Is this preferred way of interacting with the api server/component in a lambda?

PR #6006, opened 2026-01-29
[esp32] Document Arduino selective compilation optimization 

bdraco:
## Description:

Documents the Arduino selective compilation feature for ESP32 Arduino builds.

**Breaking Change:** More Arduino libraries are now disabled by default. Previously, many Arduino libraries (WiFi, Network, BLE, Zigbee, Matter, etc.) were compiled even though ESPHome uses ESP-IDF APIs directly and never called them. In most Arduino configs, none of these libraries were actually used, yet they bloated the binary by 50% or more and consumed significant RAM.

If you were already adding libraries via `libraries` config or calling `cg.add_library()`, no action is needed. If you were using Arduino library APIs directly in lambdas without adding them to the `libraries` config, you will need to explicitly add them.

- esphome/esphome#13623

## Checklist:

    or


ximex (commented):
I'm not a native english speaker but i think the comma after the `e.g.` is not necessary

content/components/esp32.md
```diff
@@ -298,6 +298,30 @@ esp32:
...
+    - Preferences  # If using Arduino Preferences API in lambda
+
+esp32:
+  board: esp32dev
+  framework:
+    type: arduino
+```
+
+> [!NOTE]
+> If you were already adding libraries via `libraries` config or calling `cg.add_library()`, no action is needed. If you were previously using Arduino library APIs directly in lambdas (e.g., `Preferences`, `Wire`, `SPI`) without adding them to the `libraries` config, you will need to explicitly add them.
```
bdraco:
```suggestion
> If you were already adding libraries via `libraries` config or calling `cg.add_library()`, no action is needed. If you were previously using Arduino library APIs directly in lambdas (e.g. `Preferences`, `Wire`, `SPI`) without adding them to the `libraries` config, you will need to explicitly add them.
```

swoboda1337 (approved):

PR #6011, opened 2026-01-30
[modbus_controller] Update EPEVER example to use ESPTime API

bdraco:
## Description:

Update the EPEVER modbus_controller lambda example to use ESPHome's `ESPTime` API instead of libc's `localtime()`.

This is a best practice update related to esphome/esphome#13635. While the old `::localtime()` code will continue to work (ESPHome overrides `localtime()` on embedded platforms), using `ESPTime::from_epoch_local()` is the recommended approach as it's more explicit and doesn't rely on the override.

### Changes:
- Replace `::localtime()` with `ESPTime::from_epoch_local(::time(nullptr))`
- Replace `tm_*` struct fields with ESPTime fields (`time.second`, `time.minute`, etc.)
- Replace `year != 70` check with `time.is_valid()` for clearer intent
- Add note that a time component must be configured
- Fix grammar: "localtime of MCU" → "local time of the MCU"

The behavior is preserved: battery settings still run even if time is not yet synced.

- esphome/esphome#13635

## Checklist:

    or


nagyrobi:
Tested on another device with a similar connection, works fine.

bdraco:
Thanks

jesserockz (dismissed):

bdraco:
thanks

PR #6162, opened 2026-02-26
esp32_hosted: enable changing the speed of the communication link between master and slave

deirdreobyrne:
## Description

Some devices are unable to maintain a communication link between master and slave running at the default 40MHz. Hence there is a need to be able to reduce the speed of that link.

- esphome/esphome#14319

## Checklist

  or


swoboda1337 (changes requested):
Docs have changes to other files. Looks like you created this on current but need to merge into next. I would fix it but can not since you created this pr directly from current instead of creating a branch first.

PR #6216, opened 2026-03-05
Documentation for the template/climate component

rsre:
## Description

Documentation for the `template` climate platform (`/components/climate/template`).

Settable properties (mode, target temperature, target humidity, fan/swing mode, preset) are plain
internal state. A command is forwarded out either through a per-field `set_<field>_action`, which
receives just that field's value, or through the base `on_control` trigger, which sees the whole
`ClimateCall`. `optimistic` decides whether the entity applies the value immediately or waits for
the device to report back.

`current_temperature`/`current_humidity` come from optional `sensor`/`humidity_sensor` references
and are pushed, not polled. `action` has no settable counterpart and is reported through the
`climate.template.publish` action, documented here for the first time, along with the warning that
action logs for a value outside the configured supported list.

The `supports_*` options are derived from the rest of the configuration wherever something implies
them: `sensor`/`humidity_sensor` for the current-value traits, and the matching set action for
two-point target temperature and target humidity. `supported_modes` is required.

Also covers `restore_mode` (`RESTORE`/`NO_RESTORE`) and `initial_state:` for controlling and
seeding boot-time state.

- esphome/esphome#14455

## Checklist

<details>

1. Comment on this pull request with the following command, replacing `component_name` with your component name in **lower_case** format with **underscores** (e.g., `bme280`, `sht3x`, `dallas_temp`):

   ```
   @esphomebot generate image component_name
   ```

3. Extract the SVG file and place it in the `/public/images/` folder of this repository.

4. Use the image in your component's index table entry in `/src/content/docs/components/index.mdx`.

```
@esphomebot generate image dht22
```

**Note:** All images used in ImgTable components must be placed in `/public/images/` as the component resolves them to absolute paths.

</details>

### DIFF
--- a/src/content/docs/components/climate/template.mdx
+++ b/src/content/docs/components/climate/template.mdx
@@ -1,0 +1,208 @@
+---
+description: "Instructions for setting up template climate controllers in ESPHome."
+title: "Template Climate"
+---
+
+import APIRef from '@components/APIRef.astro';
+
+The `template` climate platform allows you to create a climate entity backed by lambda expressions
+and automation actions. It is useful for wrapping custom UART/I²C climate controllers, integrating
+hardware that lacks a dedicated ESPHome component, or prototyping climate integrations.
+
+Once defined, it will automatically appear in Home Assistant as a climate entity and can be
+controlled through the frontend.
+
+```yaml
+# Example configuration entry
+climate:
+  - platform: template
+    name: "Living Room Climate"
+    current_temperature: !lambda "return id(my_temperature_sensor).state;"
+    supported_modes:
+      - "OFF"
+      - HEAT
+      - COOL
+      - FAN_ONLY
+    supported_fan_modes:
+      - AUTO
+      - LOW
+      - HIGH
+    supported_swing_modes:
+      - "OFF"
+      - VERTICAL
+    supported_presets:
+      - NONE
+      - ECO
+      - AWAY
+    set_mode_action:
+      - logger.log:
+          format: "Mode: %d"
+          args: ["(int) x"]
+    set_target_temperature_action:
+      - logger.log:
+          format: "Target temperature: %.1f"
+          args: ["x"]
+    set_fan_mode_action:
+      - logger.log:
+          format: "Fan mode: %d"
+          args: ["(int) x"]
+    set_swing_mode_action:
+      - logger.log:
+          format: "Swing mode: %d"
+          args: ["(int) x"]
+    set_preset_action:
+      - logger.log:
+          format: "Preset: %d"
+          args: ["(int) x"]
+    optimistic: true
+    visual:
+      temperature_step: 0.5
+```
+
+## Configuration variables
+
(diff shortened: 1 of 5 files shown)


rsre:
@esphomebot generate image climate_template

src/content/docs/components/climate/template.mdx
```diff
@@ -0,0 +1,385 @@
...
+title: "Template Climate"
+---
+
+import APIRef from '@components/APIRef.astro';
+
+The `template` climate platform lets you create a fully-featured climate entity out of plain
+ESPHome primitives. It is useful for wrapping custom UART/I²C climate controllers, integrating
+hardware that lacks a dedicated ESPHome component, or prototyping climate integrations.
+
+Settable properties (`mode`, `target_temperature`, `target_temperature_low` /
```
j9brown: I suggest rewriting this whole section in the active voice from the perspective of the developer.

Tell the developer how to configure the template entity in YAML, what each configuration attribute does.

Then tell the developer what to do at runtime to handle control triggers and publish state updates (or use optimistic mode).

There's no need to say that these properties aren't lambdas since they're runtime state and aren't even definable in YAML as part of the configuration.

Same down below for current_temperature and current_humidity, that's all just state and it works the same way so it can be described together with mode and the others.  The only thing that's special is that the developer can configure temperature/humidity sensors whose values are published automatically.

src/content/docs/components/climate/template.mdx
```diff
@@ -0,0 +1,385 @@
...
+    on_press:
+      - climate.template.publish:
+          id: two_point_heatpump
+          mode: HEAT_COOL
+          target_temperature_low: 18.0
+          target_temperature_high: 24.0
+          target_humidity: 50.0
+```
+
+## Custom fan modes and presets example
```
j9brown: I feel ike there are too many examples.  The two-point thermostat and presets could be shown as part of one of the other examples.  They're just a few extra traits and everything else is the same after all.

src/content/docs/components/climate/template.mdx
rsre: Rewrote the intro in active voice from the developer's perspective: first how to declare which traits the entity supports in YAML (pointing to Configuration variables below), then how to handle an incoming command (on_control / optimistic mode), then how to report state back (climate.template.publish). Dropped the "not lambdas" framing entirely, and folded current_temperature/current_humidity in with the rest of the runtime state, the only thing called out as special is the optional sensor/humidity_sensor auto-publish wiring.

src/content/docs/components/climate/template.mdx
rsre: Folded the two-point temperature and custom fan modes/presets examples into the external device one. It's now a single comprehensive example demonstrating every optional trait at once, with a note to drop whichever ones a given device doesn't need. Down to two examples total (simple + external device).

ssieb (approved):

PR #6230, opened 2026-03-07
[usb_uart] Add CH9344 and CH348 to supported devices

p1ngb4ck:
Docs for esphome/esphome#<ch934x code PR number>.

- Mention ESP32-P4 as a supported platform in the component description
- Add CH9344 (4-port) and CH348 (8-port) to the supported devices table with VID/PID
- Add `ch9344` and `ch348` to the `type` option list
- Add device types section note explaining port limits on non-P4 variants vs full 8-port support on ESP32-P4

Why two config types (ch9344 / ch348) for the CH934X family instead of one

The CH934X family could have collapsed into a single ch934x type, but we deliberately kept two. The two sub-families are distinguished by USB PID (0xE018 = CH9344, 4 ports; 0x55D9 = CH348, 8 ports) and differ in port count. Keeping them as separate types lets max_channels be validated exactly at config time (4 vs 8), so a config with too many channels is caught during validation rather than only at boot.

A single combined type would have had to advertise the family maximum (8) as its channel ceiling, which would let an 8-channel config validate against a 4-port CH9344 and only fail (or silently load just 4 channels) at runtime. Two types avoid that ambiguity while still keeping the config small.

We did drop the per-chip L/Q variants (CH9344L/Q, CH348L/Q): those share their family's PID and are indistinguishable over USB descriptors, so they never warranted their own type. The driver already resolves the exact sub-variant at boot (PID for the port count, a register bit for L vs Q), so nothing is lost by not exposing them in the config.

- esphome/esphome#14586

## Checklist

  or

<details>

1. Comment on this pull request with the following command, replacing `component_name` with your component name in **lower_case** format with **underscores** (e.g., `bme280`, `sht3x`, `dallas_temp`):

   ```
   @esphomebot generate image component_name
   ```

3. Extract the SVG file and place it in the `/public/images/` folder of this repository.

4. Use the image in your component's index table entry in `/src/content/docs/components/index.mdx`.

```
@esphomebot generate image dht22
```

**Note:** All images used in ImgTable components must be placed in `/public/images/` as the component resolves them to absolute paths.

</details>


p1ngb4ck:
not stale

PR #6232, opened 2026-03-07
Clarify on_click behavior with multiple entries

nmattia:
I originally assumed that only the first matching entry would trigger.

## Checklist


nmattia:
Screenshot 2026-03-07 at 15 31 45

jesserockz (approved):

PR #6250, opened 2026-03-10
Schema extract

glmnet:
## Description

Schema doc generator with mdast intermediate.

The biggest change is that it not longer scans line by line the docs, but it first exports an mdast (mark down ast) from the mdx files and then walks the ast to fill in the docs, this much better handles the mixed syntax found in the mdx file and avoids many regex patterns.
After trying to get it working from the md directly, and figured out creating the mdast, and created it, I asked Claude to convert the schema doc to consume the mdast instead of the md and it decided to write the program in plain javascript. I think it is ok since much code here is already javascript and it runs really fast.

It needs two scripts to run:
```
    "export-ast": "astro build --ast",
    "schema-doc": "node script/schema-doc/index.js"
```

these are not merged with `&&` because once the export is executed, if you need to re-run the schema-doc program, you don't need to update the ast (unless you modify the mdx) and that is a bit costly operation because it will build all the site anyway. I could not find an elegant way to prevent astro not building everything.

If this is merged then updating the CI action in esphome/esphome-schema will be needed and docs will be better than now (right now for the description of components it is picking the `import ` rules of the mdx files, which is really broken)

Closes https://github.com/esphome/backlog/issues/96


swoboda1337:
What is the improvement vs the current one:
```
{'action_docs': 476,
 'condition_docs': 48,
 'core_docs': 246,
 'core_platform_docs': 35,
 'enum_docs': 323,
 'missing_anchors': [],
 'platform_docs': 529,
 'props': 5491}
 ```

**Comment by glmnet** [comment]
> What is the improvement vs the current one:
> 
> ```
> {'action_docs': 476,
>  'condition_docs': 48,
>  'core_docs': 246,
>  'core_platform_docs': 35,
>  'enum_docs': 323,
>  'missing_anchors': [],
>  'platform_docs': 529,
>  'props': 5491}
> ```

Here is the output of the new one:
```
Building anchor map...
Processing 671 files...
Saving schemas...

── Stats ──
  core component docs:    247
  core platform docs:     33
  platform component docs:476
  config props:           3907
  enum values:            145
  action docs:            471
  condition docs:         50
```
  but most noticeably is what it outputs: 

Current one
  
image
new one
image

So the current one is picking the `import `... code from the mdx as documentation text in many places, trying to fix that by ignoring code was not very straightforward for me as the way mdx mixes javascript markdown and html, so this is why I fell into this rabbit hole, may be just ignoring the import lines at the beginning would fix it but is already looking stitched everywhere
Actually I'm coming up with that approach in PR #6251

swoboda1337:
Config props dropped a lot. The old one was already missing a few 1000.

PR #6395, opened 2026-04-06
[ags10] Change tvoc from Optional to Required

swoboda1337:
## Description

Change `tvoc` from Optional to Required in the AGS10 docs to match the code change in the matching esphome PR.

The `to_code` function accesses `config[CONF_TVOC]` unconditionally, so omitting it caused a `KeyError`. An AGS10 sensor without TVOC has no purpose, so it's now Required.

- esphome/esphome#15474

## Checklist

  or


bdraco (approved):

swoboda1337:
Thanks

PR #6483, opened 2026-04-19
[faq] Document cpu_frequency as brownout mitigation

### LINKED ISSUE #15849: ESPHome 2026.4.0 is Breaking for Modded Consumer Devices & Power Constrained Projects
### The problem

For any project that does not have control of the upstream power delivery hardware, the blind switch from 160MHz to 240MHz can cause breaking and potentially bricking instability. The cause is that upstream hardware, such as voltage regulators, may have been built around the lower power draw of 160 MHz and will experience brownouts at 240MHz.

Example cases where this is an issue:
- An ESP32 consumer device that has been modded to run ESPHome 
  - [emporia-vue-local is experiencing issues requiring reflash from serial](https://github.com/emporia-vue-local/esphome/discussions/409)
 - An ESP32 dev board that is wired into an existing device
   - [neato-brainslug was already having power issues at 160MHz](https://github.com/Philip2809/neato-brainslug/discussions/111)

Given that Home Assistant now just shows ESPHome updates and firmware updates as issues to be repaired with a single click, there's a high likelihood that users will inadvertently put their devices into an unstable state via an OTA update that may require direct hardware access to recover from. This is not great if that hardware is not easily accessible (say, inside a electrical panel)

2026.4.0

Home Assistant Add-on

ESP32

esp32

### YAML Config

```yaml

```

If you're lucky, this happens:
```txt
[W][safe_mode:058]: OTA rollback detected! Rolled back from partition 'app1'
[W][safe_mode:059]: The device reset before the boot was marked successful
```

If you're unlucky, random brownouts.

### Additional information

Proposed solution:
Patch OTA code to keep the existing CPU frequency on any OTA update unless the YAML specifically contains the cpu_frequency flag or all external components set a new max_cpu_frequency flag to certify that their implementation is compatible.

swoboda1337:
## Description

- N/A (docs-only)

Adds lowering \`cpu_frequency\` as a mitigation to the "Brownout detector was triggered" FAQ entry.

With 2026.4.0 the default CPU frequency on ESP32 was raised to 240 MHz. Users whose upstream power hardware (regulators, long/thin wiring, modded consumer devices wired into existing circuits) was sized around the previous 160 MHz default can now see brownouts after an OTA update, sometimes requiring serial recovery. Documenting the \`cpu_frequency\` knob here gives affected users a YAML-only fix that sits alongside the existing WiFi output_power, PSU, and capacitor suggestions.

## Checklist

  or

-fixes https://github.com/esphome/esphome/issues/15849

### DIFF
--- a/src/content/docs/guides/faq.mdx
+++ b/src/content/docs/guides/faq.mdx
@@ -459,11 +459,13 @@ wifi:
 
 2. **Lower the CPU frequency** -- on the ESP32, current draw scales roughly with clock speed. If you cannot
    change the upstream power hardware (e.g. a modded consumer device or a board wired into an existing circuit),
-   dropping the CPU frequency can keep transient current within what the regulator can supply:
+   dropping the CPU frequency can keep transient current within what the regulator can supply. The allowed
+   values are variant-specific; see [`cpu_frequency`](/components/esp32/#configuration-variables) on the ESP32
+   platform page for the values supported by your chip:
 
 ```yaml
 esp32:
-  cpu_frequency: 160MHz  # or 80MHz on variants that support it
+  cpu_frequency: 160MHz
 ```
 
 3. **Use a quality power supply** -- a dedicated 5V/2A supply with a short, thick USB cable.
--- a/src/content/docs/guides/faq.mdx
+++ b/src/content/docs/guides/faq.mdx
@@ -450,7 +450,8 @@ current can spike to 300-500mA. The most common causes:
 
 ### Quick fixes
 
-1. **Reduce WiFi transmit power** -- this is the most effective single change:
+1. **Reduce WiFi transmit power** -- this is the most effective single change. See
+   [`output_power`](/components/wifi/#configuration-variables) on the WiFi component page for the full range:
 
 ```yaml
 wifi:
--- a/src/content/docs/guides/faq.mdx
+++ b/src/content/docs/guides/faq.mdx
@@ -457,9 +457,18 @@ wifi:
   output_power: 13dB  # default is 20dB
 ```
 
-2. **Use a quality power supply** -- a dedicated 5V/2A supply with a short, thick USB cable.
+2. **Lower the CPU frequency** -- on the ESP32, current draw scales roughly with clock speed. If you cannot
+   change the upstream power hardware (e.g. a modded consumer device or a board wired into an existing circuit),
+   dropping the CPU frequency can keep transient current within what the regulator can supply:
 
-3. **Add a capacitor** -- a 100-1000µF electrolytic capacitor across the 3.3V and GND pins can absorb brief current spikes.
+```yaml
+esp32:
+  cpu_frequency: 160MHz  # or 80MHz on variants that support it
+```
+
+3. **Use a quality power supply** -- a dedicated 5V/2A supply with a short, thick USB cable.
+
+4. **Add a capacitor** -- a 100-1000µF electrolytic capacitor across the 3.3V and GND pins can absorb brief current spikes.
 
 ## Component states not restored after reboot
 


flaviut:
This should just be rolled back, not a FAQ entry added.

If it ain't broke, don't fix it. In this case, the default already was working just fine for millions of devices in the field; what problem does changing it solve?

swoboda1337:
> If it ain't broke, don't fix it. In this case, the default already was working just fine for millions of devices in the field; what problem does changing it solve?

Most people don't know the cpu_frequency option exists and end up running at 160 only to have performance issues. 

Arduino used to default to max, we returning to that old default. Most variants are already at the max so why not move the last few to the max?

Plus wifi output power is the big issue, cpu frequency contributes very little. At 160 the cpu draws ~30–35 mA and 240 the cpu draws ~45–55 mA . Wifi @ 20dBm draws 250mA to 350mA which dwarfs the 15-20 mA from the CPU frequency 
increase. 

Brownouts are not common and this doesn't effect many people. I have never seen a device that browns out personally. This change benefits the vast majority of users.

Jason2866:
My 2 cents. In commercial devices there are sometimes espressif OEM esp32. Variants exists which only supports 160 Mhz. Crashes are **not** power issues in this case.

swoboda1337:
> My 2 cents. In commercial devices there are sometimes espressif OEM esp32. Variants exists which only supports 160 Mhz. Crashes are **not** power issues in this case.

That seems rare though and Arduino already defaults to 240.

Jason2866:
Arduino is not designed to be used to build firmware for OEM MCUs.

swoboda1337:
> Arduino is not designed to be used to build firmware for OEM MCUs.

True. Well we have an option to lower it.

Jason2866:
The docu should mention that in rare cases 240 Mhz may not be supported. Not more not less. There are MCUs in the field which are running esphome and do not support 240 Mhz

flaviut (approved):
We have confirmation from at least one person that [the Emporia Vue 3 is fine without changing the frequency](https://github.com/emporia-vue-local/esphome/discussions/409#discussioncomment-16624194), so it seems like this was a red herring all along. With that in mind, and [the extensive research and justification in the original 240MHz PR,](https://github.com/esphome/esphome/pull/15143) this LGTM.

Jason2866:
Speaking with experience. Xiaomi (Lamps) has OEM espressif MCUs. Some batches are 160Mhz only.

swoboda1337:
> The docu should mention that in rare cases 240 Mhz may not be supported. Not more not less. There are MCUs in the field which are running esphome and do not support 240 Mhz

Yeah I can add that to this pr. 

There are also single core versions that don't work by default.

Omnipius:
If the new default is truly telling the chip to set itself to it's maximum, rather than specifically 240 MHz, 160 MHz devices shouldn't be an issue. If that's not the case, then that is bug that needs to be addressed. 

Speaking of which, @swoboda1337: @bdraco slammed the door and locked #15849 as a request for rollback when the request was to add guardrails for a graceful transition (namely, not altering the CPU frequency over OTA if not specifically requested)

bdraco (approved):

bdraco:
Appreciate the feedback, but what was proposed in esphome/esphome#15849 isn't actually possible. An OTA has no way to know what frequency the device was running at before; the firmware only reflects what the YAML tells it. Config is the source of truth, not the running device, so there is nothing to preserve across an upgrade.

Defaulting to full speed is the right call; underutilizing every ESP32 by default to cover a rare edge case is the wrong tradeoff, which is exactly why it was called out as a breaking change in the 2026.4.0 changelog. For the small number of boards that genuinely can't handle 240MHz, setting `cpu_frequency` in YAML is the fix, and that's what this FAQ entry documents. Thanks for putting it together.

Omnipius:
Understood and, for the record, I agree that the default should be whatever the chip is capable of. The heart of the issue is that projects have no way of stopping an incompatible update. Perhaps I should submit a feature request that would allow external components to declare a max ESPHome version they are compatible with?

bdraco:
> Perhaps I should submit a feature request that would allow external components to declare a max ESPHome version they are compatible with?

That sounds like a good path to explore. We already have a min version

PR #6487, opened 2026-04-19
[esp32] Note that some OEM modules cap at 160MHz

swoboda1337:
## Description

**Related issue (if applicable):** Follow-up to feedback on #6483

- N/A (docs-only)

Adds a short note under the \`cpu_frequency\` docs on the ESP32 platform page covering OEM/single-core variants found in some commercial devices (e.g. certain Xiaomi lamps) that identify as ESP32 but are rated only up to 160 MHz. With 2026.4.0 raising the default CPU frequency to the chip maximum, those modules may hang or reset on first boot, and \`cpu_frequency: 160MHz\` is the fix.

## Checklist

  or


bdraco (approved):

swoboda1337:
Thanks

PR #6562, opened 2026-05-02
[cover][time_based] support to pause open/close operation

yvesf:
## Description

Adds an example how to configure a cover with a two-button input control, to stop if the operation is already in progress.

This allows to pause/halt the cover manually before arriving in a final position.


yvesf:
bump, i think this is still valid

yvesf:
bump, i think this is still valid

PR #6564, opened 2026-05-02
[ota] Document --ota-platform CLI flag for web_server OTA

bdraco:
## Description

Documents the new `--ota-platform {esphome,web_server}` CLI flag added to `esphome upload` and `esphome run` in the matching code PR.

Changes:

- **`components/ota/web_server.mdx`** — new "Using the Command Line" section that explains:
  - Auto-selection (web_server picked when it's the only configured platform).
  - The override flag (when both `esphome` and `web_server` platforms are configured).
  - That HTTP Basic auth credentials come from `web_server.auth`, not from `ota: - platform: esphome`.
  - **Why native OTA is the default**: it uses challenge-response auth with hashed nonces, so the password never crosses the wire, while the `web_server` path relies on HTTP Basic auth (cleartext on the LAN). (Compression is *not* claimed as a general benefit; gzip OTA is ESP8266-only and the other backends report `supports_compression() == false`.)
  - A WARNING that Basic auth credentials cross the LAN in cleartext, with a recommendation to prefer `platform: esphome` when there is a choice.

- **`components/ota/esphome.mdx`** — the existing "Removing a Password" recovery instructions used to say "execute the OTA update directly via the ESP web server" (manual web UI click-through). Updated to point at the new CLI flag, which is the much easier path.

- **`components/ota/index.mdx`** — the platform list briefly mentions the CLI selection so readers know the workflow exists.

- esphome/esphome#16207

## Checklist

  or


swoboda1337 (changes requested):
Should it this new option to https://esphome.io/guides/cli/?

swoboda1337 (dismissed):
nit: not sure requests.auth.HTTPBasicAuth is useful in use docs?

bdraco:
It's not. I need to do more cleanup. Busy travel day and broken airplane WiFi has me much behind

bdraco:
fully reviewed now. its good to go

swoboda1337 (approved):

PR #6578, opened 2026-05-06
[binary_sensor] Add debounce filter callout to on_multi_click section

bdraco:
## Description

**Related issue (if applicable):** refs https://github.com/esphome/esphome/issues/15323

- N/A (docs-only)

## Checklist

  or

## Background

#15323 is a long-running thread of users reporting that `on_multi_click` stopped firing reliably after upgrading to 2026.3.x. Investigation traced this to esphome/esphome#14383, which makes the main loop wake from inside the GPIO ISR via FreeRTOS task notifications. Loop response time on a GPIO edge dropped from up to ~16ms (the old `select()` timeout) to microseconds.

Before #14383, mechanical switch bounce was implicitly swallowed by the slow loop poll cadence: most bounce edges happened during the `select()` sleep and got collapsed into a single state change by the time `loop()` ran. After #14383, every bounce edge is published as its own state change. Inside `MultiClickTrigger::on_state_`, an extra edge arriving inside the trigger window after `at_index_` reaches `timing_count_` hits the `!is_valid_` branch and calls `schedule_cooldown_()`, which cancels the pending `MULTICLICK_TRIGGER_ID` timer. The double click never fires.

The matcher itself is working as documented; it never promised debouncing, and the `gpio` binary_sensor docs already recommend a filter for unstable inputs. Configs that follow that guidance are unaffected. The issue is that the `on_multi_click` page does not point users at that recommendation, so when the implicit debounce went away the failure mode looks like a multi-click bug for users who skipped the filter.

## Changes

1. Add an `> [!IMPORTANT]` callout near the top of the `on_multi_click` section (`src/content/docs/components/binary_sensor/index.mdx`) with:

   - A short explanation of why debouncing matters specifically for multi-click (the bounce-cancels-trigger mechanism).
   - A complete copy-paste YAML example using `delayed_on_off: 20ms`.
   - A link back to the existing debouncing section on the `gpio` binary_sensor page.

   `20ms` is a conservative default that comfortably covers most tactile switches without affecting any of the multi-click timing windows (which are typically `100ms`+).

2. Add entries to **2026.3.0** changelog (`src/content/docs/changelog/2026.3.0.mdx`):

   - Upgrade Checklist line for users who use timing-based binary_sensor triggers on a physical button without a filter.
   - "Action required" Breaking Changes entry under the existing Breaking Changes section, framed so users who already follow the documented gpio debouncing guidance see it as unaffecting them.

   #14383 shipped in 2026.3.0, so the change lives there. The original issue confirms broken in 2026.3.x.


swoboda1337 (approved):

bdraco:
Thanks

PR #6631, opened 2026-05-15
[store_yaml] Document new firmware YAML recovery component

bdraco:
## Description

Adds documentation for the new `store_yaml` component, which compresses the
device's YAML configuration with zstd and embeds it in flash so it can be
recovered over the native API. Also points the existing recovery FAQ entry at
the new opt-in workflow.

- esphome/esphome#16445

## Checklist

  or


bdraco:
@esphomebot generate image store_yaml

PR #6635, opened 2026-05-16
LEDC: Remove old ESP32 GPIO list of available pins

RubenKelevra:
## Description

The docs still listed an original ESP32 era list of GPIO-Pins: 'GPIO0-GPIO33'

Instead reference just that the pin must be an output pin, as there's no more detailed documentation of the pins available in the docs right now.
## Checklist


guillempages (approved):

RubenKelevra:
@guillempages thanks for the review!

Can you please have a look at the other 3 PRs I made for the page?

PR #6686, opened 2026-05-25
[docs] Lower RF receiver tolerance from 50% to 25% in setup guide

swoboda1337:
## Description

The RF receiver setup guide recommends \`tolerance: 50%\`. The default in \`remote_receiver\` is \`25%\`; doubling it widens each timing slot's window enough that adjacent slots within the same protocol overlap. Users following the guide get worse decoding than the default — wrong protocol matches and garbage data — and there's no signal in the existing prose that this is a risky knob.

Set the example to \`25%\` (the safe default) and add a short note explaining when to raise it and why values near \`50%\` cause overlap.

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):** None — docs-only.

## Checklist

  or

### DIFF
--- a/src/content/docs/guides/setting_up_rmt_devices.mdx
+++ b/src/content/docs/guides/setting_up_rmt_devices.mdx
@@ -97,11 +97,16 @@ remote_receiver:
   pin: GPIOXX
   dump: all
   # Settings to optimize recognition of RF devices
-  tolerance: 50%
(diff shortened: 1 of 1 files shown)


clydebarrow (approved):

swoboda1337:
Thanks!

PR #6747, opened 2026-06-07
[mipi_dsi] Document DMA2D async flush options

kyvaith:
## Description

Documents the `use_dma2d` and `async_lvgl_flush` options added by esphome/esphome#16853.

## Related pull request

- esphome/esphome#16853


kyvaith:
I opened esphome/esphome#16873 as the shared architecture/performance discussion for the ESP32-P4 LVGL/MIPI DSI PR stack. It includes how the MIPI DSI docs fit into the async flush, DMA2D, zero-copy/cache-sync, PPA, and LVGL 9.5 work, plus downstream benchmark numbers.

For cross-PR design questions it is probably easier to discuss there and keep this docs PR focused on documenting the scoped options.

Unfocused:
I think this would benefit from having some concise explanation of what possible downsides each option has.

ie, something to answer the obvious question: this seems like a good thing that makes things faster, why/when shouldn't I enable this? (Or: why isn't this enabled by default?)

PR #6749, opened 2026-06-07
[climate_ir] Document supports_heat_cool option

leodrivera:
## Description

Documents the new `supports_heat_cool` configuration option for the `climate_ir` base platform.

`climate_ir` previously advertised the heat/cool mode on every IR climate device, whatever `supports_heat` and `supports_cool` were set to. The matching esphome PR makes it a capability of its own: enabled by default only when the device supports both heating and cooling, and adjustable when the device does not follow that rule.

- **unset (default):** inferred from `supports_cool && supports_heat` — no change for existing configs
- **`true`:** advertise the mode regardless, e.g. cool-only units that still expose an auto mode
- **`false`:** hide the mode even on units that support both heating and cooling but have no usable auto mode

Most remotes label this button "Auto", while Home Assistant shows the mode as `heat_cool`. The option is named after the mode so it does not collide with `supports_auto` in `thermostat`, which controls a different mode (`CLIMATE_MODE_AUTO`, where the device picks the target temperature itself). The wording mentions both names so the option is easy to find either way.

Added as a base option in the climate_ir "Configuration Variables" section, alongside `supports_cool` / `supports_heat`, so it applies to every climate_ir platform.

**Related issue (if applicable):** fixes esphome/issues#3979

- esphome/esphome#19483

## Checklist


leodrivera:
Not Stale

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #6831, opened 2026-06-23
[bk72xx_ble_tracker] Add component documentation

Bl00d-B0b:
# Description

Documentation for the `bk72xx_ble_tracker` BLE scanner (esphome/esphome#17135): configuration variables and defaults, single-core WiFi/BLE coexistence guidance, and BLE sensor usage.

The controller page (`bk72xx_ble`) merged separately and is already on `next`, so this PR is now just the tracker page plus its `index.mdx` entry.

The page describes the shipped state of the component: it documents the scanner surface only. Automation triggers and the `start_scan` / `stop_scan` actions are not part of esphome/esphome#17135 and are noted as unavailable on this platform; they will be documented alongside the PR that adds them.

BLE sensor platforms attach through the shared listener interface (esphome/esphome#17150 + #17716). Connection-based components (`ble_client`, `bluetooth_proxy`) are not available on this platform.

## Checklist:

- [x] I am merging into `next` because this is new documentation with matching pull requests in [esphome](https://github.com/esphome/esphome) as linked above.

### DIFF
--- a/src/content/docs/components/bk72xx_ble_tracker.mdx
+++ b/src/content/docs/components/bk72xx_ble_tracker.mdx
@@ -1,0 +1,123 @@
+---
+description: "Instructions for setting up BK72xx (BLE 5.x) Bluetooth Low Energy device tracking using ESPHome."
+title: "BK72xx Bluetooth Low Energy Tracker Hub"
+---
+
+import APIRef from '@components/APIRef.astro';
+
+The `bk72xx_ble_tracker` component creates a global hub so that you can scan Bluetooth Low Energy
+advertisements on a **BLE 5.x** Beken node in the LibreTiny `beken-72xx` family.
+
+Chip support is inherited from the [BK72xx BLE](/components/bk72xx_ble/) controller component,
+which lists the supported and unsupported SoCs.
+
+This component provides the **BLE scanner** and the listener interface used by ESPHome's
+advertisement-based BLE sensor platforms —
+for example [BLE RSSI](/components/sensor/ble_rssi/),
+[BLE Presence](/components/binary_sensor/ble_presence/),
+[BLE Scanner](/components/text_sensor/ble_scanner/) and
+[Xiaomi BLE](/components/sensor/xiaomi_ble/) — which attach to this tracker automatically.
+
+> [!NOTE]
+> This tracker is **scan-only**: it receives advertisements but cannot open GATT connections,
+> so connection-based components such as [`ble_client`](/components/ble_client/) and
+> [`bluetooth_proxy`](/components/bluetooth_proxy/) are not available on this platform.
+
+> [!NOTE]
+> The tracker builds on the [BK72xx BLE](/components/bk72xx_ble/) controller component and
+> loads it automatically — the BLE stack (`CFG_SUPPORT_BLE`) is enabled in the build with no
+> manual `platformio_options` flag.
+
+> [!NOTE]
+> The BK72xx are **single-core SoCs** — WiFi and BLE share the same ARM CPU core.
+> See [Use on a single-core chip](#use-on-a-single-core-chip) below for how the scan `window` / `interval`
+> ratio balances BLE against WiFi.
+
+```yaml
+# Example configuration entry
+bk72xx_ble_tracker:
+```
+
+<span id="config-bk72xx_ble_tracker"></span>
+
+## Configuration variables
+
+- **scan_parameters** (*Optional*): Advanced parameters for configuring the scan behavior.
+  See also [this guide by Texas Instruments](https://dev.ti.com/tirex/explore/content/simplelink_academy_cc2640r2sdk_5_10_02_00/modules/blestack/ble_scan_adv_basic/ble_scan_adv_basic.html#scanning-basics)
+  for background reading.
+
+  > [!NOTE]
+  > The BK72xx controller scans **passive only** — there is no `active` option. Passive
+  > scanning is sufficient for BTHome/ATC thermometers and most beacons, which broadcast all
+  > their data in the advertisement (no scan response).
+
+  - **interval** (*Optional*, [Time](/guides/configuration-types#time)): The interval between
+    each consecutive scan window. The scanner cycles through the three BLE advertising channels
+    at this rate. Defaults to `100ms` — the BK reference scan rate, which together with the
+    default `window` gives a 30% duty cycle. A larger interval lowers the duty cycle and frees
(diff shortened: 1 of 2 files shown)


Bl00d-B0b:
@coderabbitai resolve

Bl00d-B0b:
Ready for review — rebased on `next`, single commit. The page matches the component at the current head of esphome/esphome#17135 (BLEHub conversion): advertisement-based sensor platforms documented as supported with a working example; `ble_client`/`bluetooth_proxy` marked not available on this platform.

Bl00d-B0b:
Updated to the final shape of esphome/esphome#17135 (scanner-only, per review): the automation triggers/actions sections are removed (they return with the automations follow-up PR and its docs), `continuous: false` now documents that the first start is external until those actions land, and the examples use the shipping defaults. The single-core window/interval guidance is retained.

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `bluetooth_proxy` documented as supported, contradicting the PR description and the `bluetooth_proxy` page

Bl00d-B0b:
All six points addressed — the important one was a real factual error on my part.

**🟡 ＃1 — `bluetooth_proxy` claim.** You were right, and right to distrust it: the note said `bluetooth_proxy` *is* supported in `active: false` mode, which is not true of the shipped state. Platform-neutral proxy support is a separate, unmerged change; on `beken-72xx` today a `bluetooth_proxy:` block does not work. Corrected to list it alongside `ble_client` as unavailable on this platform. When neutral proxy support lands, that PR updates this line and `bluetooth_proxy.mdx` together.

**🟢 ＃1 — references to unlanded work.** All three removed. The `continuous` description and the automations note now state the released fact ("not available on this platform yet") with no PR referent. The PR description was also stale as you noted — rewritten: it no longer claims automation sections are included, and it drops the stacking note since the controller page has merged.

**🟢 ＃2 — `continuous: false` + `api.on_client_connected` paragraph.** Dropped entirely, including "the pattern used by the hardware validation fleet", which was an internal reference that meant nothing to a reader.

**🟢 ＃3 — MAC placeholders.** Both examples now use `XX:XX:XX:XX:XX:XX`, matching `esp32_ble_tracker.mdx` and `sensor/xiaomi_ble.mdx`. I left the `bindkey` value concrete, since `xiaomi_ble.mdx:150` does the same — say the word if you want that placeholdered too.

**🟢 ＃4 — lead example.** Reduced to a bare `bk72xx_ble_tracker:`; the defaults are already documented per-key below.

**🟢 ＃5 — unused `APIClass` import.** Removed.

On the observation that the single-core guidance appears three times: the first is a signpost NOTE linking to the section, and the other two are the section itself and its `With BLE sensors` subsection — parent and child of the same heading rather than three independent repetitions, so I left that structure. Happy to condense if you disagree.

Verified with `script/lint.mjs` locally (passes) and rebased on latest `next`.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Thanks — five of the six nits are fixed; the last one I agree with but it belongs elsewhere.

**＃1 — duplicated chip matrix.** Fixed the right way round: rather than re-syncing two copies, the block is now one sentence pointing at the controller page, which keeps the authoritative list. I checked `bk72xx_ble.mdx` on `next` to confirm it really does carry the full list (including the T34 this page had dropped), so the link does not send readers somewhere less complete than what it replaced.

**＃2 — redundant NOTE.** Removed. Your distinction from the structure I defended last round is fair: that was a signpost plus its section, this was a third statement of two facts already made. The automation-availability note now stands on its own at that position, which is what a reader needs there.

**＃3 — `id` mentioning actions.** Fixed — trimmed to "in lambdas". It was a genuine self-contradiction two lines from the note saying actions are unavailable; inherited from `esp32_ble_tracker.mdx`, where they do exist.

**＃4 — `continuous: false` silent dead-end.** Took the stronger of your two options. The key now leads with "Leave this at `true`", and the consequence has moved into a `> [!WARNING]` next to the automation note, stating explicitly that there is no config error and no scan-start log line and that sensors simply stay unknown. That failure mode is invisible from the logs, so burying it in a defaults sentence was wrong.

**＃5 — `bindkey`.** Thanks for answering the question I asked — switched to `00112233445566778899aabbccddeeff`.

**＃6 — sensor pages documenting `esp32_ble_tracker` as the only hub.** Agreed, and agreed it is out of scope here: those four pages are being made hub-agnostic in esphome/esphome.io#7024, which is the right place for it rather than this page.

`script/lint.mjs` passes locally and the branch is rebased on `next`.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
All five fixed.

**＃1 — closing NOTE restating its own section.** You're right, and this one was mine to own: the ratio bullets were the only content in it. Dropped the NOTE wrapper and hung the bullets directly off the section's opening paragraph, which already sets up that exact framing. The single-core fact is now stated twice — the signpost and the section — instead of three times.

**＃2 — `continuous` overlapping the WARNING.** This was my error: the last round offered *either* a directive-first bullet *or* a WARNING, and I shipped both, so the failure mode was stated twice in ten lines. The bullet is now just the directive plus the mechanics, and the WARNING owns the consequence. Also dropped "the first start is external too" — you were right that it doesn't parse, since nothing before it introduces an external start and "external" is only defined later.

**＃3 — worked example buried under a tuning heading.** Promoted `With BLE sensors` to a top-level `##` section, placed before `Use on a single-core chip`. Sensor attachment is the reason most people reach this page, and it no longer sits in the table of contents underneath a chip-architecture constraint.

**＃4 — See Also.** Added `text_sensor/ble_scanner` and `sensor/xiaomi_ble`; the latter documents the platform and the `bindkey` the example uses, so its absence was the odd one.

**＃5 — repo style.** `behaviour` → `behavior` and `30 %` → `30%`. I checked both against `next` rather than take the counts on faith: `behavior` 70 files vs `behaviour` 9, and no-space `N%` 628 occurrences against zero for `N %`.

One thing to flag, since it will show up if you run it: `script/check_component_index.mjs` reports ordering problems in **Miscellaneous** and **Output Components**. Those are pre-existing on `next` — I verified by running it against an unmodified `next`, which reports the same two — and neither is near the Bluetooth block this PR touches. I've left them alone rather than pull unrelated sections into this diff.

`script/lint.mjs` passes and the branch is rebased on `next`.

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #6898, opened 2026-07-02
[modbus_controller] Document command queue, retry and dedup changes

exciton:
## Description

Documents the modbus / modbus_controller changes from [esphome/esphome#11781](https://github.com/esphome/esphome/pull/11781):

- Deprecate `command_throttle` (use `turnaround_time` on the `modbus` component) and `allow_duplicate_commands`. Polling commands are now deduplicated by the `modbus` hub while an identical frame is already pending (according to the hub rules); one-shot commands (writes and custom commands) are always transmitted and avoiding duplicates is the caller's responsibility. Both options are still accepted for backward compatibility and log a validation warning at compile time; they'll be removed in 2027.2.0. 
- Document that `max_cmd_retries` / `offline_skip_updates` are now counted per device rather than per command/register: after `max_cmd_retries` consecutive timeouts the whole device is marked offline (`on_offline` fires), its queued frames are dropped, and its updates are skipped. It is retried every `offline_skip_updates + 1` update intervals until it responds again (`on_online`).
- Document that multiple `modbus_controller` instances can share the same `address` (e.g. different `update_interval`s).
- Update the `data` lambda parameter type to `std::span<const uint8_t>` (was `const std::vector<uint8_t> &`) across the sensor, binary_sensor, text_sensor, number and select platforms.

- [[modbus_controller] Refactor to simplify message handling esphome#11781](https://github.com/esphome/esphome/pull/11781)

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `offline_skip_updates` description still states the old per-command offline rule, contradicting the new paragraph four lines below

swoboda1337:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Off-by-one: offline triggers after `max_cmd_retries` + 1 consecutive timeouts, and the new paragraph contradicts the bullet above it

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #6968, opened 2026-07-12
[it8951] Document it8951.load / it8951.refresh actions

FuNK3Y:
Documents the `it8951.load` and `it8951.refresh` actions (staged frame presentation / double buffering) added in esphome/esphome#17527, including a page-preloading example and the `is_idle()` re-render guard.

Code PR: esphome/esphome#17527

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FuNK3Y:
This is still relevant. I tried to further optimize the driver by attempting to store more than 1 frame in the panel's memory but there is not enough memory. This is so far my best shot at improving performance on e1003

PR #6976, opened 2026-07-13
[light] Document light state on its own schema

jesserockz:
## Description

Config schema headings in a component's docs are now matched to the schemas emitted from the code: a heading whose name resolves to a `*_SCHEMA` present in that component's generated JSON (for example "Light state" -> `LIGHT_STATE_SCHEMA`) documents that schema directly instead of the doc's main config schema. The generator change is generic and guarded so it does not hijack headings already handled as component/platform titles or action/condition/registry entries.

For the light component this fixes a leak: the light state color channels (`red`, `green`, `blue`, `white`, `brightness`, `color_temperature`, `cold_white`, `warm_white`) were being flattened onto the base `LIGHT_SCHEMA`, so they surfaced as top-level options on every light platform (including addressable ones such as `esp32_rmt_led_strip`). They are now documented on `LIGHT_STATE_SCHEMA` where they belong, and `color_mode`/`color_brightness` — which previously had no docs anywhere — are documented too. A full docs build shows only `light.json` changes as a result.

**Related issue (if applicable):** esphome/device-builder-frontend#1225

- N/A

## Checklist

  or

### DIFF
--- a/script/schema_doc.py
+++ b/script/schema_doc.py
@@ -655,6 +655,44 @@ def is_title(title):
     return title.startswith("#")
 
 
+# Schemas (e.g. "LIGHT_STATE_SCHEMA") of the doc section currently being parsed,
+# keyed by name. Set per-section in parse_file() so a heading can be matched to a
+# sibling schema emitted from the code.
+current_component_schemas = {}
+
+
+def title_schema_name(title):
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #7024, opened 2026-07-21
[sensors] Hub-agnostic wording on BLE advertisement sensor pages

Bl00d-B0b:
# Description

Rewords the advertisement-based BLE sensor pages to be tracker-agnostic: the platforms bind to any BLE tracker through the platform-neutral layer (esphome/esphome#17150, sensor migration esphome/esphome#17716), so prose that presented `esp32_ble_tracker` as the only option now reads "a BLE tracker (for example the ESP32 BLE Tracker)".

The three pages of the platforms migrated in batch 1 (`ble_presence`, `ble_rssi`, `ble_scanner`) additionally document the `ble_hub_id:` binding key and the `esp32_ble_id` deprecated alias (auto-migrates with a warning until ESPHome 2027.2.0).

Scope: 10 pages (`inkbird_ibsth1_mini`, `mopeka_pro_check`, `mopeka_std_check`, `ruuvitag`, `thermopro_ble`, `xiaomi_ble`, `xiaomi_miscale`, `ble_presence`, `ble_rssi`, `ble_scanner`). YAML examples, log excerpts and See Also links keep the esp32 tracker as the reference configuration. GATT-based pages (`ble_client` family) are untouched — they remain esp32-only.

Related: esphome/esphome#17716, tracker docs #6703 / #6831.

## Checklist:

- [x] I am merging into `next` because this is new documentation that has a matching pull-request in [esphome](https://github.com/esphome/esphome) as linked above.

PR #6859, opened 2026-06-26
[bthome_mithermometer] Add documentation

Bl00d-B0b:
## What does this implement?

Adds a documentation page for the `bthome_mithermometer` sensor platform, which is in-tree but currently undocumented:

- receiving measurements from [BTHome](https://bthome.io/)-format BLE devices (e.g. Xiaomi/Mijia thermometers running ATC/PVVX firmware in BTHome mode)
- `mac_address`, optional `bindkey` for encrypted advertisements, and the temperature / humidity / battery level / battery voltage / signal strength sensors
- `ble_hub_id:` binding key with the `esp32_ble_id` deprecated-alias note (matches the sensor migration series)
- component index entry

Scope note: earlier revisions of this branch also reworded the shared `ble_presence` / `ble_rssi` / `ble_scanner` / `xiaomi_ble` pages; that sweep now lives in #7024, and this PR is scoped to the new page only.

## Related

- esphome/esphome#17150 (`ble_device_base`)

PR #7025, opened 2026-07-21
[atc_mithermometer] Add component documentation

Bl00d-B0b:
# Description

Adds a dedicated option-reference page for the `atc_mithermometer` sensor platform.

As @nagyrobi correctly pointed out, the platform is already documented by example within the [LYWSD03MMC section of the Xiaomi BLE page](https://esphome.io/components/sensor/xiaomi_ble/#lywsd03mmc) — this PR does not duplicate that. The division of labor: the xiaomi page remains the **firmware-selection guide** for these thermometers (stock MiBeacon / ATC / PVVX / BTHome formats), while this page is the **per-platform option reference** (all five sensor options), consistent with the other standalone BLE platforms. The two are cross-linked in both directions, and the page notes the ATC custom format is deprecated in favor of BTHome v2. The page also documents the `ble_hub_id:` binding key with the `esp32_ble_id` deprecated-alias note, matching the sensor migration series.

Related: esphome/esphome#17716, #6859, #7024.

## Checklist:

- [x] I am merging into `next` because this is new documentation for an existing feature.

### DIFF
--- a/src/content/docs/components/index.mdx
+++ b/src/content/docs/components/index.mdx
@@ -299,6 +299,7 @@ Sensors are organized into categories; if a given sensor fits into more than one
 <ImgTable items={[
   ["Alpha3", "/components/sensor/alpha3/", "alpha3.jpg"],
   ["AM43", "/components/sensor/am43/", "am43.jpg", "Lux & Battery level"],
+  ["ATC MiThermometer", "/components/sensor/atc_mithermometer/", "bluetooth.svg", "dark-invert"],
   ["BLE Client Sensor", "/components/sensor/ble_client/", "bluetooth.svg", "dark-invert"],
   ["BLE RSSI", "/components/sensor/ble_rssi/", "bluetooth.svg", "dark-invert"],
   ["HHCCJCY10 (MiFlora Pink)", "/components/sensor/xiaomi_hhccjcy10/", "xiaomi_hhccjcy10.jpg", "Soil moisture & Temperature & Light"],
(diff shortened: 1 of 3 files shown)


nagyrobi:
> the component has existed since 2021 but never had a docs page (only a cookbook mention).

not quite true…

Bl00d-B0b:
You're right @nagyrobi — poorly worded on my part: the platform is documented by example in [xiaomi_ble#lywsd03mmc](https://esphome.io/components/sensor/xiaomi_ble/#lywsd03mmc); what it lacked was a per-platform option reference page. Reworked accordingly: this page is now purely the option reference (and notes the format is deprecated in favor of BTHome v2), it defers all firmware-selection/flashing guidance to the LYWSD03MMC section instead of duplicating it, and the two link to each other. PR body corrected.

swoboda1337:
Should there be a link in the component index? https://esphome.io/components/

swoboda1337:
@esphbot review

Bl00d-B0b:
Both items addressed in 2baf7d07:

1. @swoboda1337 — the index entry was missing; added under Bluetooth Low Energy (BLE) sensors, placed so `script/check_component_index.mjs` reports no new unsorted section (the five it still lists are pre-existing on this branch's base).
2. Copilot — correct, and verified against `xiaomi_ble.mdx:197`: PVVX firmware also emits `ATC1441` advertisements. The intro now scopes the platform to the ATC1441 advertisement format, covering both ATC firmware and PVVX set to "ATC1441", rather than to ATC firmware alone. The deprecation note now links the BTHome MiThermometer page.

Bl00d-B0b:
Lint fix: the BTHome MiThermometer link pointed at a page that only exists in esphome/esphome.io#7128, which is still open, so the internal-link check failed. Reverted to plain `bthome_mithermometer` — it can become a link once #7128 merges.

swoboda1337 (approved):

swoboda1337:
Thanks

PR #7036, opened 2026-07-23
[ln882h_ble] Add component documentation

Bl00d-B0b:
## Description

Documentation for the `ln882h_ble` component — BLE controller support for the LN882H LibreTiny chips: BLE stack enable/disable (`enable_on_boot`) and the controller's persistent BLE address, matching the scope of the code PR. Adds the component to the Bluetooth/BLE index table.

**Related issue (if applicable):** design discussion esphome/esphome#3758

- esphome/esphome#17777

## Checklist

  or

### DIFF
--- a/src/content/docs/components/ln882h_ble.mdx
+++ b/src/content/docs/components/ln882h_ble.mdx
@@ -1,0 +1,36 @@
+---
+description: "Instructions for setting up the LN882H BLE controller in ESPHome."
+title: "LN882H Bluetooth Low Energy"
+---
+
+import APIRef from '@components/APIRef.astro';
+
+The `ln882h_ble` component enables the Bluetooth Low Energy controller of the LN882H
+(LibreTiny `ln882x` platform) — the platform analog of
+[ESP32 BLE](/components/esp32_ble/). It owns the BLE stack bring-up and the controller's
+persistent BLE address; higher-level consumers (such as the LN882H BLE tracker) build on
+it and load it automatically.
+
+```yaml
+# Example configuration entry
+ln882h_ble:
+  enable_on_boot: false
+```
+
+## Configuration variables
+
+- **enable_on_boot** (*Optional*, boolean): Bring the BLE stack up during boot. Defaults
+  to `false`: on the single-core LN882H, stack init during boot competes with the WiFi
+  connection handshake, so consumers enable the stack lazily on first use instead.
+- **id** (*Optional*, [ID](/guides/configuration-types#id)): Manually specify the ID used
+  to reference this component.
+
+> [!NOTE]
+> **BLE adapter address.** The address is loaded from the device's persistent storage;
+> when none is stored it is derived once as **WiFi MAC + 1** (only the last byte
+> incremented, OUI unchanged — the standard factory pairing) and persisted.
+
+## See Also
+
+- [ESP32 Bluetooth Low Energy](/components/esp32_ble/)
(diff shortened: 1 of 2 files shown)


src/content/docs/components/ln882h_ble.mdx
```diff
@@ -0,0 +1,40 @@
...
+  to reference this component.
+
+> [!NOTE]
+> **BLE adapter address.** The address is loaded from the device's persistent storage;
+> when none is stored it is derived once as **WiFi MAC + 1** (only the last byte
+> incremented, OUI unchanged — the standard factory pairing) and persisted.
+
+> [!NOTE]
+> No framework patch or fork is needed since LibreTiny v1.13.0 (bundled with
+> ESPHome 2026.7.0 and newer).
```
bdraco: No need to have a comment that they don't need to do anything. This is a bit of slop. Please drop

esphbot: Agreed — drop lines 33-35. Note tells readers to do nothing. Second problem: it implies availability from 2026.7.0. LibreTiny 1.13.0 shipped there (changelog/2026.7.0.mdx:969), but `ln882h_ble` targets `next`, so 2026.7.0 has no such component. Version floor belongs in release notes for whichever release ships it.

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Maintainer-requested removal of the "no patch needed" note is still outstanding
> - `lightning-ln882h` is not a real platform/family identifier

src/content/docs/components/ln882h_ble.mdx
Bl00d-B0b: Dropped, and I applied the same rule to the component PR (esphome#17777) rather than treating it as a one-off — the equivalent "no framework patch is needed" paragraph is gone from the module docstring there too, along with ~23 lines of similar narration.

Also fixed the second item @esphbot raised: `lightning-ln882h` was not a real identifier. The page now says `ln882x`, which is the actual platform component (`libretiny/const.py` maps `FAMILY_LN882H` → `COMPONENT_LN882X`) and matches the `ln882x:` block users actually write.

Head is `99c98778`, rebased onto `next`.

One thing to flag rather than hide: while rebasing this branch I briefly force-pushed the wrong base onto it — I targeted `current` when this PR is based on `next`, the rebase stopped on a conflict, and I pushed the detached state. The branch was wrong for about a minute. It is fully restored: the head above contains exactly the two intended files (`ln882h_ble.mdx`, `index.mdx`) and nothing else. Apologies for any notification noise.

bdraco (approved):

src/content/docs/components/ln882h_ble.mdx
esphbot: Confirmed dropped at HEAD. Page now carries one NOTE, covering BLE adapter address. No version-availability claim remains.

esphbot: Both warnings verified fixed. Line 9 reads `ln882x`; matches `libretiny.mdx:28`. Force-push noted — diff contains exactly two files, no stray content. Three nits remain, none blocking: example sets `enable_on_boot: false`, its own default; "LN882H BLE tracker" unlinked, no page exists; See Also lacks `/components/libretiny/`. Description still cites #3758, wrong issue.

esphbot (approved):
> [!TIP]

PR #7061, opened 2026-07-28
[projects] Make the ready-made projects page data driven

jesserockz:
## Description

The ready-made projects page repeated a full block of markup, styles and copy for every device, so adding a device or a project group meant copy-pasting an entire section of MDX. Project types and devices now live in `src/data/readyMadeProjects.ts` and are rendered by a new `ReadyMadeProjects` component, so adding a device is an entry in an array and adding a project group is an entry in `PROJECT_TYPES`. `src/content/docs/projects/index.mdx` goes from 673 lines to 15.

Devices can name their own YAML repository via `githubUrl`, overriding their project group's. Home Assistant Voice PE now points at `esphome/home-assistant-voice-pe` in both the voice assistant and media player groups, and the XIAO IR Mate at `esphome/infrared-proxies` rather than `esphome/rf-proxies`. The "For advanced users" link follows the selected device when a group's devices disagree, and stays a single static line when they don't.

The rendered HTML was diffed against the previous build. Everything matches apart from four intentional differences, three of which are fixes to problems the old markup carried:

- Device descriptions were emitted as invalid nested `<p><p>...</p></p>` - MDX was re-wrapping the raw `<p>` blocks.
- The Bluetooth proxy and empty device sections were missing `projects-hidden`, so both were briefly visible before the script ran.
- The Bluetooth install button declared an `esphome-web` manifest that the script overwrites on load; it looks like a leftover from the empty device section.
- The three single-label `.types` containers for the ESP32-S3-BOX variants are now one container. No style targets `.types` itself.

Device-scoped show/hide moved from "device id as a CSS class" to `data-device="<id>"`, which covers info blocks, variant selectors and the per-device repository links uniformly and stops device ids from having to be valid class names.

Per CONTRIBUTING, flagging the new component for maintainers: this is the "reusable component for repeated patterns" case, and it also takes the page's `<style>` block out of the MDX.

## Checklist

  or


jesserockz:
@esphbot review

esphbot (approved):
> [!TIP]

src/components/ReadyMadeProjects.astro
```diff
@@ -0,0 +1,382 @@
...
+function githubUrlsFor(type: ProjectType): (string | undefined)[] {
+  return [...new Set(type.devices.map((device) => githubUrlFor(type, device)))];
+}
+---
+
+<script is:inline type="module" src={ESP_WEB_TOOLS_URL}></script>
+
+<blockquote class="unsupported projects-hidden">
+  <p>
+    This page requires a browser that supports WebSerial. Please open this website on your desktop using Google Chrome
```
balloob:
```suggestion
    This page requires a browser that supports WebSerial. Please open this website on your desktop using Firefox, Google Chrome
```

balloob (dismissed):

balloob:
The paragraphs inside the description don't have a margin, it should. Now things like Buy section js glued to description. (This was broken in migration to current theme)

balloob (approved):

PR #7113, opened 2026-08-05
[install] Restore the Intel Mac deprecation warning

jesserockz:
## Description

The Intel Mac deprecation warning added in esphome/esphome-desktop#366 is not present on https://esphome.io/install/. That PR added the notice to the desktop app's own download page; when the install flow moved here, `InstallSelector.astro` was written as a fresh implementation of that UI and carried over the arch segmented control, the per-arch download URLs and the UA-CH Intel auto-detection, but not the warning. It looks like an oversight in the port rather than a deliberate removal.

This restores it. The toggle is hooked into the existing `wireSegmented("mac-arch", ...)` callback rather than a separate refresh function, since that callback is the single path every architecture change flows through, so manual clicks and the UA-CH auto-detect both update the warning. The note carries `role="status"` so screen readers announce it when the architecture changes, matching the upstream behaviour. Styling uses Starlight's `--sl-color-orange` / `--sl-color-orange-low` tokens instead of upstream's hardcoded amber plus a manual dark-mode override, since those are already themed here.

Verified by driving the architecture buttons locally: hidden on Apple Silicon, visible on Intel, with the download link tracking correctly. `npm run build`, `format:check`, `lint:js` and `lint` all pass.

One thing worth a maintainer's eye: the wording is taken verbatim from the upstream PR, including "no longer publishes Intel builds as of version 49". If `cryptography` has since shipped that release, this may want rephrasing into the past tense.

## Checklist

  or

### DIFF
--- a/src/components/InstallSelector.astro
+++ b/src/components/InstallSelector.astro
@@ -205,6 +205,10 @@ if (!urls) {
           <button type="button" data-arch="x64" aria-pressed="false"> Intel </button>
         </div>
       </div>
+      <p class="install-selector__warning" data-mac-intel-warning role="status" hidden>
+        Support for Intel Macs will be removed soon. The Python ecosystem is dropping Intel macOS; the{" "}
+        <code>cryptography</code> package no longer publishes Intel builds as of version 49.
+      </p>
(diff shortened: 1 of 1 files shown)


bdraco:
Thanks for catching that

PR #7126, opened 2026-08-07
Rework modbus docs to incorporate new actions

exciton:
## Description

Reworks the modbus documentation so the pages reflect the component set as it now stands.
Hardware setup is consolidated onto the Modbus hub page, with Modbus Controller and Modbus Server pointing at it instead of repeating it.
The Modbus Client page gains the typed read/write actions and the lambda hook component, and the long EPEVER lambda example is replaced by a pointer to the actions that supersede it.
The `modbus_client` id in the Modbus Controller example is renamed to `modbus_client_hub` to avoid a name collision with the new component.

Review feedback has been addressed: the PZEMAC example now uses `logger.log` rather than the non-existent `log` action, and internal links, placeholder triggers and the minimal hub example follow the repository conventions.

Note on scope: the `register_count` option is documented on several platform pages that do not accept it.
Only `sensor`, `text_sensor` and `select` define it, so the bullet has been removed from `binary_sensor`, `switch` and `number`, and the `text_sensor` default corrected.
Those bullets predate this pull request, but it had already reworded them, so they are corrected here rather than left stating something the schemas reject.

**Related issue (if applicable):** N/A

**Pull request in [esphome](https://github.com/esphome/esphome) with YAML changes (if applicable):**

- esphome/esphome#18146

## Checklist

- [x] I am merging into `next` because this is new documentation that has a matching pull-request in [esphome](https://github.com/esphome/esphome) as linked above.
  or
- [ ] I am merging into `current` because this is a fix, change and/or adjustment in the current documentation and is not for a new component or feature.

- [ ] Link added in `/src/content/docs/components/index.mdx` when creating new documents for new components or cookbook.

<details>
<summary><strong>New Component Images</strong></summary>

If you are adding a new component to ESPHome, you can automatically generate a standardized black and white component name image for the documentation.

**To generate a component image:**

1. Comment on this pull request with the following command, replacing `component_name` with your component name in **lower_case** format with **underscores** (e.g., `bme280`, `sht3x`, `dallas_temp`):

   ```
   @esphomebot generate image component_name
   ```

2. The ESPHome bot will respond with a downloadable ZIP file containing the SVG image.

3. Extract the SVG file and place it in the `/public/images/` folder of this repository.

4. Use the image in your component's index table entry in `/src/content/docs/components/index.mdx`.

**Example:** For a component called "DHT22 Temperature Sensor", use:

```
@esphomebot generate image dht22
```

[...]

PR #7154, opened 2026-08-10
[modbus_controller] Fix broken examples and correct address/response_size docs

exciton:
## Description

Follow-up to #7126, addressing the non-blocking findings left over from its final review.

Two documentation corrections, both verified against the component source:

- `modbus_controller`'s `address` was described as "a hex byte, `0x00`-`0xFF`".
The validator is `cv.hex_uint8_t`, which accepts any integer literal, so decimal works too - and the page's own `custom_command` example uses `address: 2`.
The wording now matches the platform pages ("can be decimal or hexadecimal") and mentions the valid Modbus device range of `1`-`247`.
- The `binary_sensor` `response_size` bullet lost its default when the `register_count` reference was removed.
It defaults to `2`, and `SensorItem::get_register_size()` returns early for bit entities, so the option has no effect at all on `coil` and `discrete_input` - which is what a binary sensor most often is.

Three YAML examples on the Modbus Controller page did not parse, all pre-existing:

- `modbus_controller_id: epever id: charging_mode` - two keys on one line, which invalidates the rest of the block.
- `unit_of_measurement` and `accuracy_decimals` indented four spaces where their siblings sit at two, so the mapping never closes.
- `sensors:` as the top level key in two examples, which is not a valid ESPHome platform key; it should be `sensor:`.

Each of these blocks is a copy-paste target, and `script/lint.mjs` does not parse YAML, so CI does not catch them.

## Checklist

- [x] I am merging into `next` because two of these corrections apply to wording introduced by #7126, which only exists on `next`.
  or


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `response_size` scope statement omits `custom_command`, where it does apply

esphbot (approved):
> [!TIP]

bdraco (approved):

bdraco:
Thanks @exciton

PR #7161, opened 2026-08-11
[modbus] Correct modbus docs to match current behaviour

exciton:
## Description

This is an accuracy scrub of the modbus documentation against the current component code.
It corrects inaccuracies across the modbus, modbus_client, modbus_controller and modbus_server pages and their entity sub-pages.

I am targeting `next` because several corrections describe behaviour from the unreleased modbus overhaul.

Changes:

- modbus: fix the garbled register-table list and note that function code 23 (`0x17`) is available in both the client and server roles, not the server only.
- modbus_client: rewrite the broadcast section to the current fire-and-forget behaviour (`on_sent` only, does not block the bus, reads to address `0` are refused via `on_not_sent`) and add the broadcast exception to the reply-handler contract.
- modbus_controller: note that `on_command_sent` fires once per wire transmission, so a retried command triggers it again on each attempt.
- modbus_server: name the pre-flight `ILLEGAL_DATA_ADDRESS` and `ILLEGAL_DATA_VALUE` exception codes and the two-pass write for the register write path (validation is all-or-nothing; a `write_lambda` that fails partway leaves the registers written before it applied).
- sensor: correct the lambda `item` from a const pointer to a non-const pointer.
- switch: correct `skip_updates` of 5 to refresh on every 6th cycle.
- output: mark `value_type` as optional with a default of `U_WORD`.
- binary_sensor and text_sensor: state that the `response_size` default is derived from the register type rather than a fixed two bytes.
- text_sensor: use the non-deprecated `modbus::helpers::word_from_hex_str` in the example.

- Part of the modbus overhaul series (esphome/esphome#17467, esphome/esphome#17264, esphome/esphome#18215, esphome/esphome#17464), all merged.

## Checklist

  or


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `response_size` default claim is inverted — schema default is literally `2`

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

src/content/docs/components/modbus_server.mdx
```diff
@@ -85,6 +85,14 @@ See [Modbus](/components/modbus/#hardware-setup) for RS485 hardware setup instru
     as well as `address` containing the address of this register. You must return `true` if the operation was successful, `false` otherwise, in which case
     a `SERVICE_DEVICE_FAILURE` (code `4`) exception will be sent to the client.
 
+    > [!NOTE]
+    > Before any `write_lambda` runs, the whole request is validated: every targeted register must exist and be
+    > writable, or it is rejected with an `ILLEGAL_DATA_ADDRESS` (code `2`) exception and nothing is written. A
+    > register write that does not supply a complete value for a wide type (`U_DWORD`, `FP32`, …) — including a
+    > single-register write (`0x06`) to one — is rejected with `ILLEGAL_DATA_VALUE` (code `3`). Once these checks pass, the writes are applied in
```
bdraco:
```suggestion
    > single-register write (`0x06`) to one — is rejected with `ILLEGAL_DATA_VALUE` (code `3`). 
    > Once these checks pass, the writes are applied in
```

bdraco (approved):
Thanks @exciton

PR #7233, opened 2026-08-21
[modbus_client] Document the continuous option

exciton:
## Description

Documents the `continuous` option added to the Modbus Client actions in esphome/esphome#18542.

`continuous: true` turns a single request into a standing poll: the frame stays in the hub's queue and is sent again after each successful reply, with the action's reply handlers firing once per cycle.
The option is accepted by the four read actions and by `modbus_client.send`.

A new "Continuous Polling" section covers the parts that are not obvious from the option itself:

- There is no interval setting - the frame is re-queued as soon as its reply has been handled, paced by the hub's `send_wait_time` and `turnaround_time`.
- A poll ranks below every one-shot request, so it never delays an ordinary send.
- It stops on a Modbus exception, on a timeout with no retry granted, when the same action runs again with `continuous: false`, or when the queue is cleared for that device or address.
- Running an action with `continuous: true` while an identical one-shot from the same action is still in flight converts that request into a poll rather than queueing a second frame.
- Writes are guarded twice: a literal write PDU with `continuous: true` is rejected at config time, and anything that can only be decided at run time has the option stripped by the hub with a warning. `0x17` counts as mutating despite its read half, which is also why `read_write_multiple_registers` does not offer the option.

- esphome/esphome#18542

## Checklist

  or


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - "Clear the queue for the device" is paired with the address-scoped call
> - Run-time stripping rule contradicts the `continuous` config-var text

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #7275, opened 2026-08-25
[hoermann_hcp] Document fragility of HCP2 bus

burmistrzak:
## Description

As described in the linked issue, the HCP2 bus is somewhat fragile and doesn't tolerate accessories suddenly disappearing. This includes but is not limited to ESPHome OTA updates or manual restarts.

There's also a health & safety aspect to this. Again, see the original issue for context. 😳

**Related issue (if applicable):** https://github.com/esphome/esphome/issues/18739

- N/A

## Checklist

  or

### DIFF
--- a/src/content/docs/components/cover/hoermann_hcp.mdx
+++ b/src/content/docs/components/cover/hoermann_hcp.mdx
@@ -28,12 +28,12 @@ The following Series 4 motors communicate over the HCP bus (Modbus RTU) and shou
 >
 > Disrupting Modbus connectivity for any reason may render the motor **temporarily inoperable**.
 >
-> If this happens, all controls will be unresponsive in this error state, including BiSecur hand remotes.
+> If this happens, the motor will be unresponsive to all controls in this error state, including BiSecur hand remotes.
 > Ensure you have emergency access to your garage at all times,
 > as this type of error can only be cleared by power-cycling the motor.
 
 > [!IMPORTANT]
-> Updating ESPHome via OTA **will** trigger the aforementioned motor error.
+> Restarting the ESPHome device (including OTA updates) **will** trigger the motor error described above.
 
 Communication with the motor happens over [Modbus](/components/modbus/) with the `role` set to
 `server`, which in turn requires a [UART bus](/components/uart/). The UART **must** be configured with
--- a/src/content/docs/components/cover/hoermann_hcp.mdx
+++ b/src/content/docs/components/cover/hoermann_hcp.mdx
@@ -23,6 +23,18 @@ The following Series 4 motors communicate over the HCP bus (Modbus RTU) and shou
 > This component implements the same protocol as the [HCPBridge](https://github.com/hkiam/HCPBridge)
 > project.
 
+> [!CAUTION]
+> HCP2 accessories are not hot-pluggable!
+>
+> Disrupting Modbus connectivity for any reason will render the motor **temporarily inoperable**.
+>
+> All controls will be unresponsive in this error state, including BiSecur hand remotes.
+> Ensure you have emergency access to your garage at all times,
+> as this type of error can only be cleared by power-cycling the motor.
+
+> [!IMPORTANT]
+> Updating ESPHome via OTA **will** trigger the aforementioned motor error.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/components/cover/hoermann_hcp.mdx
```diff
@@ -23,6 +23,18 @@ The following Series 4 motors communicate over the HCP bus (Modbus RTU) and shou
...
 > project.
 
+> [!CAUTION]
+> HCP2 accessories are not hot-pluggable!
+>
+> Disrupting Modbus connectivity for any reason will render the motor **temporarily inoperable**.
+>
+> All controls will be unresponsive in this error state, including BiSecur hand remotes.
+> Ensure you have emergency access to your garage at all times,
+> as this type of error can only be cleared by power-cycling the motor.
```
bdraco:
```suggestion
> Disrupting Modbus connectivity for any reason may render the motor **temporarily inoperable**.
>
> If this happens, all controls will be unresponsive in this error state, including BiSecur hand remotes.
> Ensure you have emergency access to your garage at all times,
> as this type of error can only be cleared by power-cycling the motor.
```

Given we have reports both ways I think we shouldn't over state

burmistrzak: @bdraco I haven't seen any reports disagreeing with the original wording of the warning. Would you mind linking them?

On a technical level, it's extremely difficult to keep the HCP2 bus alive, particularly during ESP32 restarts. Due to the motor’s rapid polling of accessories, four consecutive missed check-ins are already sufficient to trigger the error state.
Timing is _extremely_ tight.
ESPHome would need to complete OTAU, initialize required components, start listening to Modbus, and respond with the correct HCP2 message.
All this would need to happen in less than ~230 ms.

esphbot: Hedging alone won't settle this. Better fix: state mechanism inline. Four consecutive missed polls trigger lockout, roughly 230 ms budget. No restart path meets that. Strong wording then reads as evidence, not overstatement — and survives future readers who'd soften it again.

esphbot: Timing detail belongs in page, not thread. Put four-missed-polls and ~230 ms budget into CAUTION block. Separate point: line 33 says only power-cycling clears error, while lines 125-126 tell users to repeat bus scan when something goes wrong. Two recovery procedures, one page. Worth reconciling.

esphbot (approved):
> [!TIP]

src/content/docs/components/cover/hoermann_hcp.mdx
zweckj: > On a technical level, it's extremely difficult to keep the HCP2 bus alive, particularly during ESP32 restarts. Due to the motor’s rapid polling of accessories, four consecutive missed check-ins are already sufficient to trigger the error state.
Timing is extremely tight.

Well this one is not generally true. I can freely restart, or even have a broken firmware on the ESP it won't block as long as the RS485 stays powered. I don't know your motor or the pre-built boards you are using. The general warning is fair though.

burmistrzak: > Well this one is not generally true. I can freely restart, or even have a broken firmware on the ESP it won't block as long as the RS485 stays powered. I don't know your motor or the pre-built boards you are using. The general warning is fair though.

@zweckj Again, HUGE _if_ true.
I'm personally unable to reproduce your results on my end. Been using [this board](https://shop.tynet.eu/rs485-bridge-diy-hoermann-mqtt-adapter-esp32-s3) from @Tysonpower.

Can you please share the exact details of **your** hardware setup (incl. motor model & date of manufacture)?

If we're still unable to reproduce with your _exact kit_, then motor revision might be to blame. If that's indeed the case, we'll have to properly document it.

bdraco: I'm not going to hold up a warning over this. I'm going to accept suggestions and merge. It can be adjusted later

bdraco (dismissed):

bdraco (approved):

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #7283, opened 2026-08-27
[hlk_fm22x] Document the new options, actions and sensors

jonny190:
Covers the additions in esphome/esphome#18817:

- Single frame enrollment, used when no direction is given, and the named directions middle, right, left, down and up alongside the numeric codes.
- The admin, timeout and allow_duplicate options on hlk_fm22x.enroll, and timeout on hlk_fm22x.scan.
- The hlk_fm22x.cancel and hlk_fm22x.get_face_details actions, and the on_face_details trigger that answers the latter.
- The admin variable on on_face_scan_matched.
- The scanning binary sensor and the serial_number and face_state text sensors, and the keyed form of the binary sensor platform.
- A table of the module's error codes, so the error variable on on_face_scan_invalid and on_enrollment_failed can be read without the datasheet.

Also notes two things found while testing against an HLK-FM225: a face state of 1 usually means the person is too close rather than absent, and the UART bus has to run at 115200.

## Description

- esphome/esphome#18817

## Checklist

  or

### DIFF
--- a/src/content/docs/components/hlk_fm22x.mdx
+++ b/src/content/docs/components/hlk_fm22x.mdx
@@ -303,11 +303,13 @@ on_enrollment_failed:
 
 Starts enrolling a face. There are two ways to enroll:
 
-- **Single frame**: leave out `direction`. The person looks straight at the camera and the face is stored as soon
-  as the module has one good frame. This is the quickest way and works well for most setups.
+- **Single frame**: leave out `direction`. The person looks straight at the camera and the face is stored from
+  one good frontal frame. It is the quickest way to enroll and is what Hi-Link's own tool uses by default.
 - **Five directions**: give a `direction`. The module captures one direction per call, so you call the action five
-  times (middle, right, left, down and up, in any order) while the person turns their head a little each time.
-  A failure in one direction requires enrolling the face again from the start. Use
+  times (middle, right, left, down and up, in any order) while the person turns their head a little each time,
+  roughly 5 to 55 degrees up and down and 8 to 60 degrees left and right. This gives the module reference views
+  for someone not facing the camera squarely, so prefer it when the camera sits above or below face height or
+  people approach from the side. A failure in one direction requires enrolling the face again from the start. Use
   [`hlk_fm22x.cancel`](#hlk_fm22x-cancel-action) to throw away a half finished enrollment.
 
 ```yaml
@@ -331,13 +333,17 @@ on_...:
 **Configuration options:**
 
 - **name** (**Required**, string, [templatable](/automations/templates)): The name associated with the face. Up to 31 ASCII characters.
+  Names are labels, not keys: two enrollments may share a name, and each keeps its own `face_id` and admin flag.
+  Use unique names if you want to tell people apart by name.
 - **direction** (*Optional*, [templatable](/automations/templates)): The direction to capture the face from: `middle`, `right`, `left`, `down` or `up`.
   The module's numeric codes are accepted as well. Leave this out, or pass `0`, for a single frame enrollment.
 - **admin** (*Optional*, boolean, [templatable](/automations/templates)): Mark the face as an administrator. The flag is reported back in
   [`on_face_scan_matched`](#on_face_scan_matched-trigger) so you can give some people extra rights. Defaults to `false`.
 - **timeout** (*Optional*, [Time](/guides/configuration-types#time)): How long the module waits for a usable face before it gives up, between 1 and 255 seconds. Defaults to `10s`.
-- **allow_duplicate** (*Optional*, boolean, [templatable](/automations/templates)): Whether the same person may be enrolled more than once.
-  With `false` the module refuses a face it already knows and [`on_enrollment_failed`](#on_enrollment_failed-trigger) fires with error `10`. Defaults to `true`.
+- **allow_duplicate** (*Optional*, boolean, [templatable](/automations/templates)): Whether the module should accept a face that matches one it already holds.
+  With `false` it compares each new capture against every stored template and, if there is a match, refuses and
+  [`on_enrollment_failed`](#on_enrollment_failed-trigger) fires with error `10`. The name plays no part in that check.
+  Defaults to `true`, which matches how the module enrolls without this option.
 
 ## `hlk_fm22x.scan` Action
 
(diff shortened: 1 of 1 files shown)


src/content/docs/components/hlk_fm22x.mdx
```diff
@@ -188,16 +301,27 @@ on_enrollment_failed:
...
 ## `hlk_fm22x.enroll` Action
 
-Starts the face enrollment process with a name and direction.
-To successfully enroll a face, you need to successfully and consecutively scan the face from all directions.
-A failure in one direction will require enrolling the face again from the start.
+Starts enrolling a face. There are two ways to enroll:
+
+- **Single frame**: leave out `direction`. The person looks straight at the camera and the face is stored as soon
+  as the module has one good frame. This is the quickest way and works well for most setups.
+- **Five directions**: give a `direction`. The module captures one direction per call, so you call the action five
```
OnFreund, line 308: What are the advantages of a five-direction enrollment over single-frame? You mention that single frame works for most setups, but is it less reliable? When does it not work?

src/content/docs/components/hlk_fm22x.mdx
```diff
@@ -206,17 +330,40 @@ on_...:
...
-- **direction** (**Required**, int, templatable): The direction to scan the face for. `1` for center, `2` for right, `4` for left, `8` for down, and `16` for up.
+- **name** (**Required**, string, [templatable](/automations/templates)): The name associated with the face. Up to 31 ASCII characters.
+- **direction** (*Optional*, [templatable](/automations/templates)): The direction to capture the face from: `middle`, `right`, `left`, `down` or `up`.
+  The module's numeric codes are accepted as well. Leave this out, or pass `0`, for a single frame enrollment.
+- **admin** (*Optional*, boolean, [templatable](/automations/templates)): Mark the face as an administrator. The flag is reported back in
+  [`on_face_scan_matched`](#on_face_scan_matched-trigger) so you can give some people extra rights. Defaults to `false`.
+- **timeout** (*Optional*, [Time](/guides/configuration-types#time)): How long the module waits for a usable face before it gives up, between 1 and 255 seconds. Defaults to `10s`.
+- **allow_duplicate** (*Optional*, boolean, [templatable](/automations/templates)): Whether the same person may be enrolled more than once.
```
OnFreund: Is that same person or same face? Why would you allow the same face to be enrolled twice? Should we default to false?

src/content/docs/components/hlk_fm22x.mdx
jonny190: Good question, and the wording should be tighter than it is.

The datasheet describes the difference purely as a matter of procedure. Five direction enrollment captures the face at five head poses, roughly 5 to 55 degrees up and down and 8 to 60 degrees left and right, and only stores a template once all five are in. Single frame stores one frontal capture. It does not quantify recognition reliability for either mode, and I have not measured that on hardware, so I should not have implied single frame is as good in every case.

What can be said: five directions gives the module reference views for someone who is off axis when scanned, which matters when the camera sits above or below head height or people approach from the side. Single frame is quicker for the person enrolling and is what Hi-Link's own debugging tool uses as its default enroll command.

I'll reword the paragraph to say exactly that, drop "works well for most setups", and recommend five directions when the camera is not at face height or people will not be looking straight at it.

src/content/docs/components/hlk_fm22x.mdx
jonny190: Face is the right word. The module compares face templates and has no notion of a person beyond that, so I'll fix the wording.

On the default: the plain enroll commands the component has always used, `0x13` and `0x1D`, do no duplicate check at all. The datasheet says the same face can be enrolled repeatedly and simply gets a new ID each time. Only the integrated enroll command, `0x26`, can refuse a duplicate, returning error `10`. So `allow_duplicate: true` is the existing behaviour, and `false` moves every enrollment onto `0x26`, which I have not yet exercised on hardware.

The case for allowing duplicates is enrolling one face more than once under different conditions, glasses on and off for example, to give the module more to match against. The case against is that most people want one entry per face and duplicates just clutter the ID space.

I'm happy to default to `false` if you prefer. It would be a behaviour change for existing users, so I'd call it out in the PR and want to test `0x26` on the module first. The alternative is to keep `true` for compatibility now and flip it in a later release with a deprecation note. Your call.

OnFreund: Makes sense RE the behavior but I still want to push back on the wording, though. It sounds like what's being duplicated is the name. The face is neither duplicated logically (new `face_id`) nor physically (this could be a different person altogether).
BTW what happens if the same name is enrolled twice, with one of them marked as an admin while the other isn't?

jonny190: Fair challenge, so I went back to the datasheet rather than trust my memory, and it turns out the name is the one thing this flag does not look at.

The Hi-Link manual describes the integrated enroll's `enable_duplicate` as:

> 0: The same person cannot be entered, but the user name can be repeated.
> 1: The same person can be entered repeatedly, and the user name can also be repeated.
> 2: The same person can be entered, but the username cannot be repeated (only valid when RGB is registered).

Mode 2 is the only one that keys on the name, and it only applies to photo enrollment, which this component does not expose. In modes 0 and 1 the name may repeat freely. The check in mode 0 is a match against the stored templates: the F900 manual's worked example enrolls a face as ID 4 with duplicates disabled, enrolls the same face again, and gets `MR_FAILED4_FACEENROLLED` (0x0A) back.

So you're right that "duplicate face" is the wrong phrase, just for a different reason: nothing is duplicated as a record, and "same" is the module's match decision, so a close lookalike would be refused too. I've reworded it to say what actually happens: with `false`, the module compares the new capture against every stored template and refuses, with error 10, if it matches one it already holds.

On the same name twice with different admin flags: names are labels, not keys. Every enrollment gets its own `face_id`, and the name and admin flag live on that record. Lookup and delete both take the `face_id`, and a scan reports the `face_id`, name and admin flag of whichever template matched. So the two records coexist, and which admin value you get back depends on which one the module matched. Neither the module nor the component enforces unique names. I've confirmed that on an FM225: enrolling the same name twice, once as admin and once not, gives two IDs, and a scan reports the flag of the one it matched. The docs now say so and recommend unique names.

Both wording changes are pushed.

src/content/docs/components/hlk_fm22x.mdx
jonny190: Reworded as discussed, in 5c5fab7. The paragraph now describes each mode by what the datasheet says it does, with the angle ranges for the five direction flow and when it is worth the extra effort, and no longer claims single frame works well for most setups.

PR #7296, opened 2026-08-30
[modbus] Document rx_timeout tuning for high baud rates

exciton:
Adds a subsection to *Performance vs Stability* on the `modbus` page covering `rx_timeout` as a throughput/CPU tuning knob.

Above 19200 baud the Modbus RTU interframe gap is fixed at 1.75 ms, far longer than the frames themselves, so the bus spends most of its time waiting and the component busy-waits the last part of that gap to meet it precisely.
Raising the UART's `rx_timeout` moves more of the wait into the UART hardware: the frame is handed over later, so less of the gap remains for the component to wait out.
Throughput is unaffected, because the gap is measured from the last byte on the wire rather than from when the frame was handed over.

Measured on a client polling three servers at 115200 baud: raising `rx_timeout` from 2 to 15 took the time the `modbus` component spent in each loop from 1.67 ms to 0.43 ms, with bus utilisation unchanged at ~96%.

The section also documents the ceiling (too large and the frame arrives too late to transmit within the gap, costing throughput) and warns that a long timeout means the UART waits longer before deciding a frame has ended, so a device that does not leave a proper gap can have two frames handed over as one and fail the checksum.

Companion to esphome/esphome#18898, though the two are independent: this documents an existing option and needs no code change.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Other

**Related issue or feature (if applicable):** N/A

**Pull request in [esphome](https://github.com/esphome/esphome) with functional changes (if applicable):** esphome/esphome#18898

### DIFF
--- a/src/content/docs/components/modbus.mdx
+++ b/src/content/docs/components/modbus.mdx
@@ -87,6 +87,42 @@ increase it.
 Reducing `send_wait_time` generally does not impact performance at all, unless you have devices that are physically
 offline. When all devices are responding, `send_wait_time` does nothing.
 
+### Reducing CPU Load at High Baud Rates
+
+Modbus RTU requires a silent gap between frames. Above 19200 baud the specification fixes that gap at 1.75 ms, which
+is far longer than the frames themselves, so at high baud rates the bus spends most of its time waiting. The component
+waits out the remainder of that gap before it transmits, and the last part of the wait is spent busy-waiting so the
+gap is met precisely.
+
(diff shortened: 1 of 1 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

jesserockz (approved):

PR #7324, opened 2026-09-04
Document hold_during_sleep option for ESP32

luar123:
Added documentation for hold_during_sleep option in ESP32 configuration.

## Description

- esphome/esphome#18964

## Checklist

  or

### DIFF
--- a/src/content/docs/guides/configuration-types.mdx
+++ b/src/content/docs/guides/configuration-types.mdx
@@ -107,6 +107,11 @@ Advanced options:
   i.e. the maximum amount of current can additionally be set. Defaults to `20mA`.
   Options are `5mA`, `10mA`, `20mA`, `40mA`.
 
+- **hold_during_sleep** (*Optional*, boolean): Hold pin during deepsleep on ESP32. Set this to `true` to hold
+  the output level or pullup/pulldown mode during deepsleep. Hold should be used only for GPIO pins and not
+  for pins that are used for peripherals like `i2c`, `uart` or `rmt`. If the pin is also used elsewhere and `allow_other_uses`
+  is set, make sure to use hold on all or none. Defaults to `false`.
+
 - **ignore_strapping_warning** (*Optional*, boolean): Certain pins on ESP32s are designated *strapping pins* and are read
   by the chip on reset to configure initial operation, e.g. to enable bootstrap mode.
   Using such pins for I/O should be avoided and ESPHome will warn if I/O is configured on a strapping pin.
--- a/src/content/docs/guides/configuration-types.mdx
+++ b/src/content/docs/guides/configuration-types.mdx
@@ -110,7 +110,8 @@ Advanced options:
 - **hold_during_sleep** (*Optional*, boolean): Hold pin during deepsleep on ESP32. Set this to `true` to hold
   the output level or pullup/pulldown mode during deepsleep. Hold should be used only for GPIO pins and not
   for pins that are used for peripherals like `i2c`, `uart` or `rmt`. If the pin is also used elsewhere and `allow_other_uses`
-  is set, make sure to use hold on all or none. Defaults to `false`.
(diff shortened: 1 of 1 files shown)


src/content/docs/components/output/gpio.mdx
```diff
@@ -19,7 +19,9 @@ output:
 
 ## Configuration variables
 
-- **pin** (**Required**, [Pin Schema](/guides/configuration-types#pin-schema)): The pin to turn on and off.
+- **pin** (**Required**, [Pin Schema](/guides/configuration-types#pin-schema)): The pin to turn on and off. When
+  enabling `hold_during_sleep` the driving component (e.g. light) or automation needs to restore the state after
+  wake-up. When using a power supply, make sure to enable `hold_during_sleep` there as well.
```
clydebarrow, line 24: This is potentially confusing (to me at least) - wouldn't a light do that by default? It _will_ come up in a defined state (which may or may not be the same as the power down state).

src/content/docs/guides/configuration-types.mdx
```diff
@@ -107,6 +107,12 @@ Advanced options:
   i.e. the maximum amount of current can additionally be set. Defaults to `20mA`.
   Options are `5mA`, `10mA`, `20mA`, `40mA`.
 
+- **hold_during_sleep** (*Optional*, boolean): Hold pin during deepsleep on ESP32. Set this to `true` to hold
+  the output level or pullup/pulldown mode during deepsleep. Hold should be used only for GPIO pins and not
+  for pins that are used for peripherals like `i2c`, `uart` or `rmt`. If the pin is also used elsewhere and `allow_other_uses`
```
clydebarrow, line 112: I can envisage cases where an i2c or uart pin might want to be held during sleep (in fact I have a specific device where it's vital to hold the UART output during sleep, currently using a lambda to do that.)

Maybe also mention it's not available on expander pins.

luar123: That sounds interesting. Could you give me an example? Would like to test if this is possible with the current implementation

clydebarrow: Well, the application is a garage door controller that has a half-duplex serial port, which also doubles as a button-press opener. So I have a GPIO as the uart TX that drives that line via a transistor. What was happening was that when I OTA upgraded the device, the software reset and startup sequence caused a glitch on the line that the opener interpreted as a button press (closure to ground.) So I had to add a lambda in the `on_shutdown` hook to disable the output.

However, I just reflashed it with the current dev (so not incorporating your changes) and took out the lambda, and now I can't reproduce the problem :-(

luar123: ok thanks. Will run some tests with uart. Compared to other peripherals uart can call `pin->setup()` and I think this would break uart completely when using hold.

luar123: Tested uart on current dev: during reboot or deep sleep it will always turn off, so the easiest solution is to use an external pull-up/down resistor.
With this PR (and a small modification I did not push yet) you could also use an internal pull-up together with `hold_during_sleep`.

Anyway, thanks for the input. Will include the uart fix and rework the logic so it also works over OTA/reboot and not only deep sleep.

clydebarrow: > you could also use an internal pull-up together with `hold_during_sleep`.

The hardware in question does have a pull-up (well, actually a pull-down, it's inverted) but the problem was that during the reboot process (I believe it was during the startup after boot) the output was driven to the active state for a brief period.

Like I said I can no longer reproduce the problem, so something else may have been fixed in the gpio component.

luar123: Oh wow there was a lot of back and forth during the last year: see history in esphome/esphome/pull/14978 and esphome/esphome/pull/18684
I think there is still a glitch when using non inverted tx and internal pullup/open drain, because pin->setup() is called without setting the level first.

But unrelated to the hold feature.

PR #7345, opened 2026-09-08
[hoermann_hcp] Document announce_pause

RaHehl:
## Description

Documents the `hoermann_hcp.announce_pause` action added in the linked pull request, and updates the
restart warning above it.

An ordinary restart is announced to the motor by the component itself. An update cannot be, because
the firmware is written long before the shutdown code runs, so it has to be announced from the OTA
`on_begin` trigger. The existing warning kept its certainty ("will trigger … unless the motor is told
first") and now also names the case that cannot be announced at all: a power cut or the reset button.

The action section says what the announcement costs, that this is longer than an `on_begin` action
should normally run and why it is nevertheless safe here, which log line tells the reader whether
their motor answered, and what happens if the update fails after the announcement.

**Related issue (if applicable):** https://github.com/esphome/esphome/issues/18739

- esphome/esphome#19043

## Checklist

  or

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/content/docs/components/cover/hoermann_hcp.mdx
+++ b/src/content/docs/components/cover/hoermann_hcp.mdx
@@ -32,7 +32,13 @@ The following Series 4 motors communicate over the HCP bus (Modbus RTU) and shou
 > as this type of error can only be cleared by power-cycling the motor.
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


RaHehl:
Closing together with esphome/esphome#19043.

PR #7352, opened 2026-09-09
-

JS3910:
Redacted as overzealous agent did not understand the difference between local development and "make public PR without asking for permission" 😒. This can and should be deleted by someone who actually can do that...


JS3910:
Code PR: esphome/esphome#19054

@esphomebot generate image OPT4048

PR #7367, opened 2026-09-11
Document the remote_receiver buffer_size unit and defaults

bdraco:
## Description

`buffer_size` is documented as an amount of data but its unit and defaults were out of date. On ESP32 variants with RMT the ring is now sized from `receive_symbols` unless set (esphome/esphome#19100), and on the other platforms the value now means bytes as the option always said, with the default moved to `4000b` so the number of buffered pulses stays at 1000 (esphome/esphome#19101). Anyone with an explicit `buffer_size` on ESP8266, RP2040, LibreTiny, ESP32-C2 or ESP32-C61 should multiply it by 4 to keep the same capacity. The RF guide's `buffer_size: 2kb` line for ESP8266 is removed rather than rescaled: with the unit fixed it would shrink the ring below the default, and the default already holds 1000 pulses; a sentence after the example points noisy receivers at the option.

- esphome/esphome#19100
- esphome/esphome#19101

## Checklist

  or

### DIFF
--- a/src/content/docs/components/remote_receiver.mdx
+++ b/src/content/docs/components/remote_receiver.mdx
@@ -80,9 +80,14 @@ Multiple remote receivers can be configured as a list of dict definitions within
     to `100%` or `0` to `4294967295us`.
 
 - **buffer_size** (*Optional*, int): The size in bytes of the internal buffer for storing the remote codes. On ESP32
-  variants with RMT the default is four receive slots derived from `receive_symbols` (3136 bytes at the defaults), and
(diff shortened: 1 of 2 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - RF setup guide still recommends `buffer_size: 2kb` for ESP8266 — now below the new default

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

swoboda1337 (approved):

PR #7418, opened 2026-09-21
[climate_ir] Mark gree as supporting the remote receiver

itayzimet:
## Description

Marks `gree` as supporting the remote receiver in the IR Remote Climate docs table.
The `on_receive()` support added in esphome/esphome#19451 makes the existing
`receiver_id` option functional, so the empty "Supports receiver" cell would no
longer be accurate.

**Related issue (if applicable):** Not applicable

- esphome/esphome#19451

## Checklist

  or


itayzimet:
The original PR was in conflict with another - so this PR is redundant and irrelevant.

